PR #1, opened 2020-02-05
first pass at: hardcoded number of electrons, added slider

cmaximilian:


yakutovicha:
btw, I was wondering - would you be able/like to make a more general widget to select pseudopotential family among the available ones? That would fix any incompatibility problems

There is a long-standing issue, which I never had time to look at: https://github.com/aiidalab/aiidalab-widgets-base/issues/27

csadorf:
@yakutovicha Let's see how far we get with this one first. @cmaximilian has only until Friday to work on the app.

PR #9, opened 2020-08-04
Provide url parameters to configure pw@localhost

yakutovicha:


ltalirz:
Just one comment: In the medium term, I think we may be moving to using the .yml files for configuring computers and codes as done in https://github.com/aiidateam/aiida-code-registry
It would therefore suggest to store these configurations in separate .yml files as well, rather than in python dictionaries inside the code.

This can be made into a separate issue (doesn't have to be solved now).

PR #11, opened 2020-08-13
Test with action supporting app names.

csadorf:
And custom docker image that does not fail with pre-mounted apps.


csadorf:
Blocked until https://github.com/aiidalab/aiidalab-test-app-action/pull/5 and https://github.com/aiidalab/aiidalab-docker-stack/pull/99 are merged.

PR #14, opened 2020-10-26
Add script for the automated compilation of the metadata.json file.

csadorf:
The metadata.json file contains metadata information about this app that
transcends versions so should ideally be published independently of a
specific repository commit.

This commit adds both a script for the compilation of the metadata.json
file from a template and a GitHub actions workflow for its publication.


csadorf:
Should be merged after merging https://github.com/aiidalab/aiidalab-registry/pull/52 .

PR #17, opened 2020-11-06
Only publish revised metadata via master branch.

csadorf:


csadorf:
And test against 'latest' and 'stable' of aiidalab-docker-stack (update commit message!).

PR #21, opened 2020-11-10
🐛 FIX: OptimadeQueryWidget moved to aiidalab_widgets_base

chrisjsewell:
According to https://github.com/aiidalab/aiidalab-optimade/blob/83fdcb60cfb1d87faba48038f7ba3cba771f2c83/aiidalab_optimade/widget.py#L12


csadorf (approved):
Thank you!

PR #39, opened 2021-03-08
Major revision: v2

csadorf:


yakutovicha (commented):
This is such a great tool, thanks a lot @csadorf.

I only noticed that sometimes the "+" and "-" symbols disappear, but I would actually not be worried about this for now.

Also, this widget (and many others) seem to be of general interest. Should we move them to aiidalab-widgets-base? Not sure when would be the best moment for this, but I would not wait too long.

yakutovicha:
@csadorf I believe the testing issue is not related to ipytree anymore. Can you please have a look at it?

csadorf:
> This is such a great tool, thanks a lot @csadorf.
> 

Thank you! :)

> I only noticed that sometimes the "+" and "-" symbols disappear, but I would actually not be worried about this for now.
> 

Me, too, but I think it may be a fault of the `ipytree` widget and I would not know how to fix it. I do not think it is a blocker to releasing this.

> Also, this widget (and many others) seem to be of general interest. Should we move them to aiidalab-widgets-base? Not sure when would be the best moment for this, but I would not wait too long.

Yes, agreed. I can prepare a PR for that.

csadorf:
> 
> Yes, agreed. I can prepare a PR for that.

@yakutovicha I've created https://github.com/aiidalab/aiidalab-widgets-base/pull/182 .

PR #40, opened 2021-03-08
Split configuration and submission, and results view into two steps.

csadorf:


yakutovicha (approved):
looks good, thanks @csadorf !

PR #49, opened 2021-03-30
First draft of the input report functions

mbercx:
Add two functions that extract some basic information regarding the input setup for the calculations in the work chain that can be reported in the results tab.


csadorf (approved):
Thank you very much!

Issue #56, opened 2021-06-05
👌 IMPROVE: Clean up top-level UI


mbercx:
I have some changes I would suggest to the top level UI:

* Remove the "Steps" buttons. You can just click on the tabs now, so I assume these are still here for some legacy reason?

PR #71, opened 2021-06-08
🐛 FIX: do not set params to default for structure change

### LINKED ISSUE #61: 🐛 FIX: Reset of work chain settings when choosing a new structure
After setting your work chain configuration in Step 2 and going back to select a different structure, the work chain settings are suddenly reset:

fix_61

Ideally this shouldn't happen, but at the very least this pathway shouldn't be possible.

mbercx:
Fixes #61 

When the user has already set some calculation settings and goes back to
choose a new structure, the `builder_parameters` are reset. I can't
think of a good reason to do this, so I removed the line that does this
from the `_observe_input_structure` method.


mbercx:
Hmm, seems this only _partially_ fixes the issue. When you haven't selected a structure yet it works fine, but _changing_ structures still resets the work chian settings.

csadorf:
@mbercx The reason for resetting the parameters is to ensure that we don't run with some unexpected parameters when starting a new workflow. Do you not think that is a valid reason?

mbercx:
Can we have a way to:

* not reset the parameters when setting a new structure.
* reset the parameters when a new workflow is selected.

mbercx:
Basically, invert my "solution" 😅

mbercx:
Moved the `CodesConfig` -> `CodeSettings` change to PR #72

mbercx:
Alright, I think this PR is not super critical atm, but I think it would more intuitive to have slightly different behaviour for the steps. As I've mentioned in #56, I'm not sure it's still necessary to have the steps buttons. I think the user should be able to move between Step 1 and Step 2 freely and make changes until she/he presses the "Submit" button. At this point all parts of the UI that change the calculation should be disabled. A couple of bullet points:

* For Step 1, the structure selection part should be disabled (including the tabs, It think). All of the structure viewer UI should still be enabled, of course.
* For step 2, this means most of the UI should be disabled, save for the `Tab`s and the "Expert mode" button. Currently you can submit, and then still change settings.
* I think the confirm button in step 1 can be removed in this case, along with the "Store in AiiDA" line.
* Once the user clicks the submit button, a check should triggered of the size of the structure and the resources selected. If we have a bit of a rule of thumb on what is reasonable here, we can add a warning that the structure might be prohibitively large for the number of resources selected.

csadorf:
> Can we have a way to:
> 
>     * not reset the parameters when setting a new structure.
> 
>     * reset the parameters when a new workflow is selected.

Of course! Previously made an attempt to restore the parameters from the selected work chain (not sure whether that was removed). So all it takes is to reset them to the defaults when selecting the (`None`) workchain, i.e., new workflow.

mbercx:
I'm closing this PR since we're probably going to work on a larger overhaul of how the steps and resetting should work, and this will supersede the minor changes made in b57235e.

PR #72, opened 2021-06-08
♻️ REFACTOR: defaults and widgets of the steps

### LINKED ISSUE #67: 🐛 FIX: Issue when disabling expert from code
When going into "Expert mode" -> "Select Codes" -> Disabling "Expert mode", the following error is raised:

```python

TraitError                                Traceback (most recent call last)
/opt/conda/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in _handle_msg(self, msg)
    674                 if 'buffer_paths' in data:
    675                     _put_buffers(state, data['buffer_paths'], msg['buffers'])
--> 676                 self.set_state(state)
    677 
    678         # Handle a state request.

/opt/conda/lib/python3.7/site-packages/ipywidgets/widgets/widget.py in set_state(self, sync_data)
    543                     from_json = self.trait_metadata(name, 'from_json',
    544                                                     self._trait_from_json)
--> 545                     self.set_trait(name, from_json(sync_data[name], self))
    546 
    547     def send(self, content, buffers=None):

/opt/conda/lib/python3.7/contextlib.py in __exit__(self, type, value, traceback)
    117         if type is None:
    118             try:
--> 119                 next(self.gen)
    120             except StopIteration:
    121                 return False

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in hold_trait_notifications(self)
   1120                                 self._trait_values.pop(name)
   1121                 cache = {}
-> 1122                 raise e
   1123             finally:
   1124                 self._cross_validation_lock = False

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in hold_trait_notifications(self)
   1106                 for name in list(cache.keys()):
   1107                     trait = getattr(self.__class__, name)
-> 1108                     value = trait._cross_validate(self, getattr(self, name))
   1109                     self.set_trait(name, value)
   1110             except TraitError as e:

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in _cross_validate(self, obj, value)
    597         if self.name in obj._trait_validators:
    598             proposal = Bunch({'trait': self, 'value': value, 'owner': obj})
--> 599             value = obj._trait_validators[self.name](obj, proposal)
    600         elif hasattr(obj, '_%s_validate' % self.name):
    601             meth_name = '_%s_validate' % self.name

/opt/conda/lib/python3.7/site-packages/traitlets/traitlets.py in __call__(self, *args, **kwargs)
    905         """Pass `*args` and `**kwargs` to the handler's function if it exists."""
    906         if hasattr(self, 'func'):
--> 907             return self.func(*args, **kwargs)
    908         else:
    909             return self._init_call(*args, **kwargs)

/opt/conda/lib/python3.7/site-packages/ipywidgets/widgets/widget_selectioncontainer.py in _validated_index(self, proposal)
     28             return proposal.value
     29         else:
---> 30             raise TraitError('Invalid selection: index out of bounds')
     31 
     32     # Public methods

TraitError: Invalid selection: index out of bounds
```

Steps to reproduce:

bug-codes

mbercx:
Fixes #67 

Several changes are made to clean up/refactor/modularize the widgets:

* Move defaults into `.yaml` that is read and used consistently. Use
`importlib_resources` for importing this into modules.
* Try to add consistency between the order of these default in the .yaml
file and how they are ordered in several commands of ths
`SubmitQeAppWorkChainStep`.
* Rename `configs` to `settings` (save for `code`, this was included in
PR #71).
* Fix issue where _both_ `OptionsConfig` (now `AdvancedSettings`) and
`SubmitQeAppWorkChainStep` would intialize a `PseudoFamilySelector`,
leading to issues in correctly assigning the pseudos.

Note that there is a small change in UI: The Materials Settings have
been moved outside of expert mode, since the layman usually can still
understand/decide whether their structure is magnetic/insulating or not.

This also seems to have fixed the issue we were having with the UI
breaking when disabling expert mode on the "Select Codes" tab.

### DIFF
--- a/aiidalab_qe/parameters/__init__.py
+++ b/aiidalab_qe/parameters/__init__.py
@@ -1,0 +1,6 @@
+import yaml
+
+from importlib import resources
+from aiidalab_qe import parameters
+
+DEFAULT_PARAMETERS = yaml.safe_load(resources.read_text(parameters, "qeapp.yaml"))
--- a/aiidalab_qe/pseudos.py
+++ b/aiidalab_qe/pseudos.py
@@ -5,15 +5,11 @@ from threading import Thread
 
 import ipywidgets as ipw
 import traitlets
-import yaml
 from aiida import orm, plugins
-from importlib_resources import files
 
-from aiidalab_qe import parameters
+from aiidalab_qe.parameters import DEFAULT_PARAMETERS
 
 SsspFamily = plugins.GroupFactory("pseudo.family.sssp")
-
-DEFAULT_PARAMETERS = yaml.safe_load(files(parameters).joinpath("qeapp.yaml").open())
 
 
 class Spinner(ipw.HTML):
@@ -155,13 +151,11 @@ class PseudoFamilySelector(ipw.VBox):
         PBE and PBEsol.</div>"""
     )
     installed = traitlets.Bool()
-    disable = traitlets.Bool()
+    disabled = traitlets.Bool()
 
     value = traitlets.Unicode(
         default_value=DEFAULT_PARAMETERS["pseudo_family"],
     )
-    FUNCTIONAL_DEFAULT = DEFAULT_PARAMETERS["pseudo_family"].split("/")[2]
-    PROTOCOL_DEFAULT = DEFAULT_PARAMETERS["pseudo_family"].split("/")[3]
 
     def __init__(self, **kwargs):
 
@@ -180,18 +174,19 @@ class PseudoFamilySelector(ipw.VBox):
         self.show_ui = ipw.Valid(value=True)
         self.set_pseudo_family.observe(self.set_show_ui, "value")
         self.set_pseudo_family.observe(self.set_text_color, "value")
+        self.set_pseudo_family.observe(self.set_value_trait, "value")
 
         # Choose the DFT functional
         self.dft_functional = ipw.Dropdown(
             options=["PBE", "PBEsol"],
-            value=self.FUNCTIONAL_DEFAULT,
+            value=DEFAULT_PARAMETERS["pseudo_family"].split("/")[2],
             style={"description_width": "initial"},
         )
         self.dft_functional.observe(self.set_value_trait, "value")
 
         self.protocol_selection = ipw.ToggleButtons(
             options=["efficiency", "precision"],
-            value=self.PROTOCOL_DEFAULT,
+            value=DEFAULT_PARAMETERS["pseudo_family"].split("/")[3],
             layout=ipw.Layout(max_width="80%"),
         )
         self.protocol_selection.observe(self.set_value_trait, "value")
@@ -234,7 +229,9 @@ class PseudoFamilySelector(ipw.VBox):
 
     def set_value_trait(self, _=None):
         self.value = (
-            f"SSSP/1.1/{self.dft_functional.value}/{self.protocol_selection.value}"
+            DEFAULT_PARAMETERS["pseudo_family"]
+            if not self.set_pseudo_family.value
+            else f"SSSP/1.1/{self.dft_functional.value}/{self.protocol_selection.value}"
         )
 
     def set_show_ui(self, change):
--- a/aiidalab_qe/pseudos.py
+++ b/aiidalab_qe/pseudos.py
@@ -5,9 +5,15 @@ from threading import Thread
 
 import ipywidgets as ipw
 import traitlets
+import yaml
 from aiida import orm, plugins
+from importlib_resources import files
+
+from aiidalab_qe import parameters
 
 SsspFamily = plugins.GroupFactory("pseudo.family.sssp")
+
+DEFAULT_PARAMETERS = yaml.safe_load(files(parameters).joinpath("qeapp.yaml").open())
 
 
 class Spinner(ipw.HTML):
@@ -127,7 +133,6 @@ class PseudoFamilySelector(ipw.VBox):
         <b><a href="https://www.materialscloud.org/discover/sssp/table/precision"
         target="_blank">Standard Solid-state pseudopotential</a> (SSSP) protocol</b></div>"""
     )
-
     pseudo_family_help = ipw.HTML(
         """<div style="line-height: 140%; padding-top: 10px; padding-bottom: 0px; opacity:0.5;">
         If you are unsure what to choose, select 'SSSP efficiency', which for
@@ -142,7 +147,6 @@ class PseudoFamilySelector(ipw.VBox):
         <div style="line-height: 140%; padding-top: 0px; padding-bottom: 10px; opacity:0.5;"><b>
         Exchange-correlation  functional</b></div>"""
     )
-
     dft_functional_help = ipw.HTML(
         """<div style="line-height: 140%; padding-top: 10px; padding-bottom: 10px; opacity:0.5;">
         The exchange-correlation energy is calculated based on the charge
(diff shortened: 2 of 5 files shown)


aiidalab_qe/pseudos.py
```diff
@@ -5,10 +5,16 @@
...
 import traitlets
+import yaml
 from aiida import orm, plugins
+from importlib_resources import files
+
+from aiidalab_qe import parameters
 
 SsspFamily = plugins.GroupFactory("pseudo.family.sssp")
 
+DEFAULT_PARAMETERS = yaml.safe_load(files(parameters).joinpath("qeapp.yaml").open())
```
csadorf, line 16: Why not load this directly within the currently `parameters/__init__.py` module and then just import it here?

aiidalab_qe/pseudos.py
```diff
@@ -142,18 +147,21 @@ class PseudoFamilySelector(ipw.VBox):
...
     dft_functional_help = ipw.HTML(
         """<div style="line-height: 140%; padding-top: 10px; padding-bottom: 10px; opacity:0.5;">
         The exchange-correlation energy is calculated based on the charge
         density using this functional. We currently provide support for two
         well-established generalised gradient approximation (GGA) functionals:
         PBE and PBEsol.</div>"""
     )
-
     installed = traitlets.Bool()
+    disable = traitlets.Bool()
```
csadorf, line 158: This should be:
```suggestion
    disabled = traitlets.Bool()
```
And are you actually using that trait?

aiidalab_qe/pseudos.py
```diff
@@ -142,18 +147,21 @@ class PseudoFamilySelector(ipw.VBox):
...
 
-    value = traitlets.Unicode(allow_none=True)
+    value = traitlets.Unicode(
+        default_value=DEFAULT_PARAMETERS["pseudo_family"],
+    )
+    FUNCTIONAL_DEFAULT = DEFAULT_PARAMETERS["pseudo_family"].split("/")[2]
+    PROTOCOL_DEFAULT = DEFAULT_PARAMETERS["pseudo_family"].split("/")[3]
```
csadorf, line 164: Why are these class attributes?

aiidalab_qe/pseudos.py
```diff
@@ -208,7 +218,7 @@ def __init__(self, **kwargs):
         )
         ipw.dlink((self.show_ui, "value"), (self.protocol_selection, "disabled"))
         ipw.dlink((self.show_ui, "value"), (self.dft_functional, "disabled"))
-        self.set_pseudo_family.observe(self.set_value_trait, "value")
```
csadorf, line 211: Why is this removed? Was the function removed?

aiidalab_qe/steps.py
```diff
@@ -35,7 +39,10 @@ def update_resources(builder, resources):
                 update_resources(v, resources)
 
 
-class WorkChainConfig(ipw.VBox):
+DEFAULT_PARAMETERS = yaml.safe_load(files(parameters).joinpath("qeapp.yaml").open())
```
csadorf, line 42: Same comment as before.

aiidalab_qe/steps.py
```diff
@@ -130,34 +137,21 @@ def set_bands_button_style(self, change):
...
-    _DEFAULT_ELECTRONIC_TYPE = "METAL"
+    spin_type = traitlets.Instance(SpinType, allow_none=True)
+    electronic_type = traitlets.Instance(ElectronicType, allow_none=True)
 
     def __init__(self, **kwargs):
 
         self._spin_type = ipw.Dropdown(
             options=["NONE", "COLLINEAR"],
-            value="NONE",
+            value=DEFAULT_PARAMETERS["spin_type"].upper(),
```
csadorf, line 154: I believe it should be possible to use 2-tuples as values for these widgets. The first one corresponds to the display value, and the second one to the actual value.

We could use this to make those dropdowns look a bit nicer, .e.g.: `("Metal", "METAL")` instead of `"METAL"` would appear as "Metal" within the menu, but set the internal value to `METAL`.

Does not need to be applied as part of this PR, just a general suggestion.

setup.cfg
```diff
@@ -23,6 +23,7 @@ install_requires =
     aiidalab-qe-workchain @ https://github.com/aiidalab/aiidalab-qe/raw/develop/src/dist/aiidalab_qe_workchain-1.0-py3-none-any.whl
     aiidalab-widgets-base~=1.0b
     widget_bandsplot~=0.2.1
+    importlib_resources~=5.1.4
```
csadorf, line 26: Please make sure that the list of install requirements remains in alphabetical order.

However, what is the reason to use this library? Why not use the one in the standard library https://docs.python.org/3.8/library/importlib.html?highlight=importlib#module-importlib.resources ?

csadorf (changes requested):

aiidalab_qe/pseudos.py
mbercx: That is cleaner, indeed! 👍

aiidalab_qe/pseudos.py
mbercx: No. But didn't you say we needed the `disabled` trait? 😭 

At some point I thought I knew why, so I reinstated it 😎

aiidalab_qe/pseudos.py
mbercx: Fair, I only use them once. Might just as well directly load them from `DEFAULT_PARAMETERS` when setting the widget `value`.

aiidalab_qe/steps.py
mbercx: Might as well, seems easy enough to fix! Let me see if this works ;)

setup.cfg
mbercx: Indeed, it seems to work fine for Python 3.8 at least:

```python
import yaml

from importlib import resources
from aiidalab_qe import parameters 

DEFAULT_PARAMETERS = yaml.safe_load(resources.read_text(parameters, "qeapp.yaml"))
```

even looks cleaner to me! Not sure why the backport only recommends "users of Python 3.9 and beyond to use the standard module".

mbercx:
@csadorf thanks for the review! I think I've covered all your suggestions, but have left some of the comments as unresolved so you can still see my answers to the comments without having to open them.

aiidalab_qe/pseudos.py
csadorf: > No. But didn't you say we needed the `disabled` trait? 😭
> 
> At some point I thought I knew why, so I reinstated it 😎

The `disabled` trait should toggle the widget into a "disabled" state when set to True. What that exactly means depends a little bit on the widget so that's why there is no default implementation for this. Ideally, subsequent steps cannot be modified (i.e. are disabled) until the previous step has been confirmed. Hence why most widgets within the app should have this trait and also implement the corresponding behavior.

aiidalab_qe/parameters/__init__.py
```diff
@@ -0,0 +1,6 @@
+import yaml
+
+from importlib import resources
+from aiidalab_qe import parameters
```
csadorf, line 4: This line doesn't make sense as it implies a self-import, doesn't it?

aiidalab_qe/pseudos.py
mbercx: Thanks for the explanation :). Makes sense!

aiidalab_qe/parameters/__init__.py
mbercx: Maybe, but `importlib.resources` needs that import in the way I'm using it.

Typically, you would just use this in a _different_ file though, as I was using it before (i.e. in the `steps.py` and `pseudo.py` files).

I mean, it works like this. Not sure how to make it work otherwise. Probably could, but I don't know how of the top of my head.

mbercx: Maybe not use `importlib.resources`? But it does seem like a clean way of importing data files.

csadorf: If it works, it works.

csadorf (approved):
🚢

mbercx:
Screenshot 2021-06-11 at 11 48 16
😭

PR #81, opened 2021-06-11
Remove obsolete SSSPInstallWidget.

csadorf:


mbercx:
Thanks @csadorf! Note that I've left the widget in there because I wasn't sure if we would _only_ have a version deployed where all the pseudos are installed automatically. 

Another thing to consider is if we want to perhaps still make installing other pseudos possible (e.g. the [PseudoDojo](http://www.pseudo-dojo.org/) ones also supported by `aiida-pseudo`).

csadorf:
> Thanks @csadorf! Note that I've left the widget in there because I wasn't sure if we would _only_ have a version deployed where all the pseudos are installed automatically.
> 
> Another thing to consider is if we want to perhaps still make installing other pseudos possible (e.g. the [PseudoDojo](http://www.pseudo-dojo.org/) ones also supported by `aiida-pseudo`).

True, but if they are not installable, then it's not helpful, is it? Let's remove it for now and then bring it back if needed.

mbercx (approved):
Let's make our workflows submittable again! 🚀

PR #91, opened 2021-06-24
Use automated bug report on app crash.

### LINKED ISSUE #74: 👌 IMPROVE: Find way to give clean error message when things break
Currently, when things break because an error is raised, the whole Trace is printed out, and things get funky. It would be better if we could somehow catch this and give a clean error messages that creates a report (strack trace + state + inputs?) that is stored on the cluster.

csadorf:
Requires https://github.com/aiidalab/aiidalab-widgets-base/pull/209

https://user-images.githubusercontent.com/1441208/123243613-79240300-d4e3-11eb-92d4-3d091f87d027.mp4

Closes #74.

### DIFF
--- a/qe.ipynb
+++ b/qe.ipynb
@@ -17,7 +17,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "scrolled": false
+   },
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


csadorf:
https://github.com/aiidalab/aiidalab-widgets-base/pull/209 has been merged, so I will merge this as well.

PR #92, opened 2021-06-29
Fix the structure appearance in the final geometry tab.

### LINKED ISSUE #69: 🐛 FIX: Structure not showing in the "Final Geometry" tab of the outputs.
In Step three, the `StructureDataViewer` in the "Final Geometry" tab is not showing the structure properly. This is most likely related to this issue:

https://github.com/nglviewer/nglview/issues/871

There is a suggested fix there, but I didn't manage to get this to work yet. This may be due to the convoluted logic I implemented to have these `Tab`s only show up once the results are completed.

yakutovicha:
fixes #69

### DIFF
--- a/aiidalab_qe/node_view.py
+++ b/aiidalab_qe/node_view.py
@@ -222,14 +222,14 @@ class WorkChainViewer(ipw.VBox):
         # An ugly fix to the structure appearance problem
         # https://github.com/aiidalab/aiidalab-qe/issues/69
         def on_change(change):
-            index = change['new']
+            index = change["new"]
             # Accessing the viewer only if the corresponding tab is present.
-            if self.result_tabs._titles[str(self.result_tabs.selected_index)] == 'Final Geometry':
+            if self.result_tabs._titles[str(index)] == "Final Geometry":
                 self.struct_view._viewer.handle_resize()
-                self.struct_view._viewer.camera = 'perspective'
-                self.struct_view._viewer.camera = 'orthographic'
+                self.struct_view._viewer.camera = "perspective"
+                self.struct_view._viewer.camera = "orthographic"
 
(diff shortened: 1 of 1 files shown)


aiidalab_qe/node_view.py
```diff
@@ -218,6 +218,18 @@ def __init__(self, node, **kwargs):
...
+
+        # An ugly fix to the structure appearance problem
+        # https://github.com/aiidalab/aiidalab-qe/issues/69
+        def on_change(change):
+            index = change["new"]
+            # Accessing the viewer only if the corresponding tab is present.
+            if self.result_tabs._titles[str(index)] == "Final Geometry":
+                self.struct_view._viewer.handle_resize()
+                self.struct_view._viewer.camera = "perspective"
+                self.struct_view._viewer.camera = "orthographic"
```
csadorf, line 230: - Do you need all three of these changes to trigger the rendering or would either the first or the last two suffice?
- What happens if the camera was previously set to "perspective"?

yakutovicha: > Do you need all three of these changes to trigger the rendering or would either the first or the last two suffice?

I couldn't entirely figure out the origin of the problem. The solution suggested [here](https://github.com/nglviewer/nglview/issues/871#issuecomment-594038985) didn't work as-is. By pure luck, I discovered that changing the camera mode would let the structure appear. Therefore, I always switch camera to the non-default "perspective" mode and back to the default "orthographic".

> What happens if the camera was previously set to "perspective"?

It is changed back to "orthographic".

csadorf: I've tested this, all commands are necessary for the fix to work. I'm going to submit a minor patch that will keep the user set camera when switching between tabs, otherwise this is good.

csadorf (approved):

PR #93, opened 2021-07-01
👌 IMPROVE: Step 2 Workflow tab

### LINKED ISSUE #76: 👌 IMPROVE: Add more explanation to material settings
Currently there is very little explanation to the materials settings (none, in fact). Would be good to explain the differences in layman's terms, since these settings are not part of the "Expert mode" panels. Also make the values more user friendly than `METAL`, `COLLINEAR`, etc...

### LINKED ISSUE #90: 👌 IMPROVE: Step 2 Workflow tab
Reorganise the "Workflow" tab in Step 2 as follows:

#### Structure

First have three options:

- structure as is
- optimize atoms
- optimize atoms and unit cell

Add "Material settings" here as well.

#### Properties

Here the user can simply select the properties to calculate. Band structure for now, DOS/PDOS later.

#### Protocol

Same as before.

mbercx:
Fixes #90 
Fixes #76 

Rearrange the content of the "Workflow" tab in step 2 of the QEapp:

Make a separate header for "Structure" where the selection of the
geometry optimization is done via three options:

* Structure as is (`RelaxType.NONE`)
* Atomic positions (`RelaxType.POSITIONS`)
* Full geometry (`RelaxType.POSITIONS_CELL`)

The "material settings" are also moved into the `WorkChainSettings`, and
placed right below the geometry optimization choice with some more
explanation.

The next header is still "Properties", but (for) now only the band
structure can be chosen here.

Finally, the "Protocol" header and content has remained the same as
before, but is not the final thing users can choose in this tab.

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -39,15 +39,29 @@ def update_resources(builder, resources):
 
 class WorkChainSettings(ipw.VBox):
 
+    structure_title = ipw.HTML(
+        """<div style="padding-top: 0px; padding-bottom: 0px">
+        <h4>Structure</h4></div>"""
+    )
+    structure_help = ipw.HTML(
+        """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 5px">
+        By default, the workflow will optimize the provided geometry. Select "Structure
+        as is" if this is not desired. You can either optimize the atomic positions ("Full geometry")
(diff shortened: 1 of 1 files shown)


csadorf (approved):
LGTM!

csadorf:
@mbercx Please merge at your leisure.

PR #137, opened 2021-10-21
👌 IMPROVE: Block non-3D structures

### LINKED ISSUE #84: 👌 IMPROVE: Basic approach for dealing with non-3D structures
All of the setup and protocols currently only really consider 3D structures, and running anything else quickly results in issues:

* 2D: seekpath fails since there is a check that the system is periodic in all three directions.
* Molecule: A simple molecule took very long to run. This might be related to the fact that the protocol doesn't consider the periodic boundary conditions at the moment, and so even with the large unit cell it could use more than just one k-point.

We should have some basic way of dealing with these problems for the release. Either we block the selection of such structures for now, or fix e.g. the k-points issue at the app level (although over time we may want to fix this for the plugin).

mbercx:
Fixes #84 

Add a warning above the structure selection that only 3D structures are
supported in addition to blocking structures that do not have 3D
periodic boundary conditions.


mbercx:
@csadorf this is a very basic approach to dealing with non-3D structures. We simply:

* add a message at the top of the structure selection step that only 3D structures are supported.
* block structures that do not have three-dimensional boundary conditions.

One issue might be that e.g. a molecule is defined in a unit cell with 3D periodic boundary conditions. Dealing with these cases is a bit trickier, since we'd have to detect the vacuum in the unit cell somehow. Maybe we can set a threshold (e.g. 5-10 angstrom) and then check a layer of this thickness on each face of the unit cell. In case any of these layers doesn't contain any sites, I think it's fair to assume the structure is not 3D.

PR #157, opened 2021-11-18
Add sentence on definition of k-points distance value

### LINKED ISSUE #78: 👌IMPROVE: Add units to every selection/output that needs them
Some of the selection widgets (e.g. the kpoints distance) do not yet have units. Make sure every inputs and output has clearly defined units.

mbercx:
Fixes #78


mbercx:
@csadorf I looked through the various steps of the app, but couldn't see any widgets/outputs that were still missing units. So I simply added a sentence explaining the k-points distance as requested in #78.

csadorf (approved):

PR #158, opened 2021-11-19
Add smearing width and type set

### LINKED ISSUE #64: ✨ NEW: Add option to set smearing settings for `ElectronicType.METAL`
Currently the expert user cannot set the smearing settings (`smearing` and `degauss`) when running a metal (`ElectronicType.METAL`). This should be added to the "Advanced Settings" page.

unkcpz:
Fixes #64 

The WorkChain has also changed so the workchain need to be released and change the dependency info. 
@mbercx What is the correct workflow to make the PR when the work chain is changed? Should those change in another PR and release first?

### DIFF
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -72,10 +72,17 @@ def _generate_report_dict(qeapp_wc):
     pw_parameters = None
     energy_cutoff_wfc = None
     energy_cutoff_rho = None
+    degauss = None
+    smearing_type = None
     scf_kpoints_distance = None
     bands_kpoints_distance = None
+    try: 
+        degauss = qeapp_wc.inputs.degauss_override.value
+    except AttributeError:
+        degauss = None
+
     try:
-        scf_kpoints_distance = qeapp_wc.inputs.kpoints_distance_override.value
+        scf_kpoints_distance = qeapp_wc.inputs.degauss_override.value
     except AttributeError:
         scf_kpoints_distance = None
     nscf_kpoints_distance = None
@@ -95,9 +102,13 @@ def _generate_report_dict(qeapp_wc):
     if pw_parameters:
         energy_cutoff_wfc = round(pw_parameters["SYSTEM"]["ecutwfc"])
         energy_cutoff_rho = round(pw_parameters["SYSTEM"]["ecutrho"])
+        degauss = pw_parameters["SYSTEM"]["degauss"]
+        smearing_type = pw_parameters["SYSTEM"]["smearing"]
 
     yield "energy_cutoff_wfc", energy_cutoff_wfc
     yield "energy_cutoff_rho", energy_cutoff_rho
+    yield "degauss", degauss
+    yield "smearing_type", smearing_type
     yield "scf_kpoints_distance", scf_kpoints_distance
     yield "bands_kpoints_distance", bands_kpoints_distance
     yield "nscf_kpoints_distance", nscf_kpoints_distance
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -73,19 +73,27 @@ def _generate_report_dict(qeapp_wc):
     energy_cutoff_wfc = None
     energy_cutoff_rho = None
     degauss = None
-    smearing_type = None
+    smearing = None
     scf_kpoints_distance = None
     bands_kpoints_distance = None
-    try: 
-        degauss = qeapp_wc.inputs.degauss_override.value
-    except AttributeError:
-        degauss = None
 
     try:
-        scf_kpoints_distance = qeapp_wc.inputs.degauss_override.value
+        scf_kpoints_distance = qeapp_wc.inputs.kpoints_distance_override.value
     except AttributeError:
         scf_kpoints_distance = None
     nscf_kpoints_distance = None
+
+    try:
+        degauss = qeapp_wc.inputs.degauss_override.value
+    except AttributeError:
+        # read default from protocol
+        degauss = 0.01
+
+    try:
+        smearing = qeapp_wc.inputs.smearing_override.value
+    except AttributeError:
+        # read default from protocol
+        smearing = "cold"
 
     if run_relax:
         pw_parameters = qeapp_wc.inputs.relax.base.pw.parameters.get_dict()
@@ -102,13 +110,11 @@ def _generate_report_dict(qeapp_wc):
     if pw_parameters:
         energy_cutoff_wfc = round(pw_parameters["SYSTEM"]["ecutwfc"])
         energy_cutoff_rho = round(pw_parameters["SYSTEM"]["ecutrho"])
-        degauss = pw_parameters["SYSTEM"]["degauss"]
-        smearing_type = pw_parameters["SYSTEM"]["smearing"]
 
     yield "energy_cutoff_wfc", energy_cutoff_wfc
     yield "energy_cutoff_rho", energy_cutoff_rho
     yield "degauss", degauss
-    yield "smearing_type", smearing_type
+    yield "smearing", smearing
     yield "scf_kpoints_distance", scf_kpoints_distance
     yield "bands_kpoints_distance", bands_kpoints_distance
     yield "nscf_kpoints_distance", nscf_kpoints_distance
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -83,17 +83,20 @@ def _generate_report_dict(qeapp_wc):
         scf_kpoints_distance = None
     nscf_kpoints_distance = None
 
+    default_params = PwBaseWorkChain.get_protocol_inputs(
+        builder_parameters["protocol"]
+    )["pw"]["parameters"]["SYSTEM"]
     try:
         degauss = qeapp_wc.inputs.degauss_override.value
     except AttributeError:
         # read default from protocol
-        degauss = 0.01
+        degauss = default_params["degauss"]
 
     try:
         smearing = qeapp_wc.inputs.smearing_override.value
     except AttributeError:
         # read default from protocol
-        smearing = "cold"
+        smearing = default_params["smearing"]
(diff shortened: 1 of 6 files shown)


mbercx:
> What is the correct workflow to make the PR when the work chain is changed? Should those change in another PR and release first?

Not sure that's necessary? I haven't made any changes to the work chain since it was placed in the `src` directory, but I'd say it's fine to have a single PR for changing both the app and work chain, unless @csadorf objects?

aiidalab_qe/steps.py
```diff
@@ -147,11 +148,84 @@ def __init__(self, **kwargs):
...
+            else self.smearing_default
+        )
+
+
 class KpointSettings(ipw.VBox):
 
     kpoints_distance_description = ipw.HTML(
         """<p>
-        The k-points mesh density is set by the chosen <b>protocol</b>.
+        The k-points mesh density of scf calculation is set by the chosen <b>protocol</b>.
```
mbercx, line 228:
```suggestion
        The k-points mesh density of the SCF calculation is set by the chosen <b>protocol</b>.
```

aiidalab_qe/steps.py
```diff
@@ -147,11 +148,84 @@ def __init__(self, **kwargs):
...
+
+    # The default of `smearing` and `degauss` the type and width
+    # must be linked to the `protocol`
+    degauss_default = traitlets.Float(default_value=0.01)
+    smearing_default = traitlets.Unicode(default_value="cold")
+
+    def __init__(self, **kwargs):
+
+        self.override_protocol_smearing = ipw.Checkbox(
+            description="Override default smearing setting.",
```
mbercx, line 168: Maybe we can just shorten the description of both this checkbox and the k-points one to "Override" in boldface. Should be clear and more succinct.

aiidalab_qe/steps.py
```diff
@@ -147,11 +148,84 @@ def __init__(self, **kwargs):
...
+            indent=False,
+            value=False,
+        )
+        self.degauss = ipw.FloatText(
+            value=self.degauss_default,
+            step=0.005,
+            description="Degauss (Ry):",
```
mbercx, line 175: Although these are definitely "expert" settings, maybe we can use the more generic "Smearing width" instead of the name of the QE input?

aiidalab_qe/report.py
```diff
@@ -72,14 +72,31 @@ def _generate_report_dict(qeapp_wc):
...
     try:
         scf_kpoints_distance = qeapp_wc.inputs.kpoints_distance_override.value
     except AttributeError:
         scf_kpoints_distance = None
     nscf_kpoints_distance = None
 
+    try:
+        degauss = qeapp_wc.inputs.degauss_override.value
+    except AttributeError:
+        degauss = None
```
mbercx, line 93: Currently, the report shows `None` for the smearing type and width when they are not overridden. Best to grab these from the work chain or protocol in case they are not specifically set by the user.

aiidalab_qe/steps.py
```diff
@@ -147,11 +148,84 @@ def __init__(self, **kwargs):
...
+            disabled=False,
+            style={"description_width": "initial"},
+        )
+        self.smearing = ipw.Dropdown(
```
mbercx, line 179: Maybe it's more natural to put the smearing type first, then the smearing width?

mbercx (changes requested):
Thanks @unkcpz! I've given the implementation a test run and left some comment regarding the frontend.

I'm also considering we should maybe add two headers in the "Advanced settings". Maybe

* Functional and pseudopotential
* Brillouin zone sampling

What do you think?

aiidalab_qe/report.py
unkcpz: Yes, I think it is hard to get it from the workchain because the default value of these variables are set in the base calculation. If we run band structure and pdos at the same time in the future, they both set the degauss and smearing in the innermost base calculation by default then we don't know which one to read is more proper. 

I now hard code the value from the protocol, but maybe there is a util function (not the one in `ProtocolMixin` have to call with `cls`) that can extract the value from the protocol?

unkcpz (commented):
Thanks @mbercx !
I update the commits and accept most of your advice. Could you have another around look at this?

aiidalab_qe/steps.py
mbercx: No worries, boldface is not really critical. I forgot how tricky it can be to get the widget format exactly as you want it sometimes. ;)

aiidalab_qe/steps.py
mbercx: Agreed! 👍

aiidalab_qe/report.py
mbercx: Sorry for the slow response to this one. Wanted to look into this more closely but got distracted by other work.

1. Regarding getting the values from the work chain: I preferred this method for the report, since then we'd actually see the values that were used, which was useful as an extra check to find bugs. I don't immediately see the issue with running both the bands and pdos. We'd use the same smearing and `degauss` for all SCF calculations, no?
2. I now use the `PwBaseWorkChain.get_protocol_inputs()` method for obtaining the defaults for the k-points density:

https://github.com/aiidalab/aiidalab-qe/blob/b5c6bcbf0392a108178c8ab6ac0fe87365dd0615/aiidalab_qe/steps.py#L276-L282

This is the easiest way to obtain the defaults for each protocol, I think. For example, to get the smearing:

```python
PwBaseWorkChain.get_protocol_inputs('moderate')['pw']['parameters']['SYSTEM']['smearing']
```

Is this the one you wanted to avoid, or?

csadorf:
> > What is the correct workflow to make the PR when the work chain is changed? Should those change in another PR and release first?
> 
> Not sure that's necessary? I haven't made any changes to the work chain since it was placed in the `src` directory, but I'd say it's fine to have a single PR for changing both the app and work chain, unless @csadorf objects?

Indeed, unless I am missing something, adding a new work chain is a two-step process, because one has to update the dependency URL to something that has already been pushed:
https://github.com/aiidalab/aiidalab-qe/blob/b5c6bcbf0392a108178c8ab6ac0fe87365dd0615/setup.cfg#L31

In practice, I would recommend to

1. Manually install the work chain during development, then
2. Push both the new work chain and related changes with dependency set to the next (pre-)release.

The tests will almost certainly have to fail during the transition phase.

aiidalab_qe/report.py
unkcpz: Almost forget about this PR 🥲. I update it as advised, try to finish this.

unkcpz:
@mbercx Have a final look when you back from holiday next year 😄

mbercx:
@unkcpz sorry for leaving this PR in the cold for so long! If you rebase and fix the conflicts I'll give it a final review so we can wrap this up.

unkcpz:
Hi @mbercx, I rebase the PR and think it is good to have another round.

unkcpz:
When looking at the original feature request https://github.com/aiidalab/aiidalab-qe/issues/64
In this PR I did nothing to distinguish the METAL and insulator.

aiidalab_qe/steps.py
```diff
@@ -159,11 +160,84 @@ def __init__(self, **kwargs):
         )
 
 
+class SmearingSettings(ipw.VBox):
+
+    smearing_description = ipw.HTML(
+        """<p>
+        The smearing type and width is set by the chosen <b>protocol</b>.
+        Untick the box to override the default.
```
mbercx, line 168: With the current interfact the user has to tick the box to override.

mbercx: This should also be adapted for the k-points override. :)

aiidalab_qe/steps.py
```diff
@@ -159,11 +160,84 @@ def __init__(self, **kwargs):
...
+    </p>"""
+    )
+
+    # The default of `smearing` and `degauss` the type and width
+    # must be linked to the `protocol`
+    degauss_default = traitlets.Float(default_value=0.01)
```
mbercx, line 174: Here the defaults are still hard coded instead of retrieved from the protocol. Currently the defaults don't really have to be linked to the protocol, since the smearing is the same for all protocols. This might change in the future, however.

mbercx (approved):
Thanks @unkcpz! Just a few final comments. I'll already approve and let you decide whether it makes sense to already link the protocol to the default smearing values. Now the smearing settings are the same for all protocols, so perhaps adding extra code isn't sensible at this point.

aiidalab_qe/steps.py
unkcpz: In the `Configuration` it is linked to the protocol already, it will change when you choose a different protocol. Here, I think it is fine to have a default value for the class.

aiidalab_qe/steps.py
unkcpz: Good catch! I will change this.

aiidalab_qe/steps.py
unkcpz: I see your point. Yes, it is somehow not linked to the protocol. I have a fake illusion I have done that :1

unkcpz: Wait, I think there are duplicate codes in both `KpointSetting` and `ConfigureQeAppStep` that link to the protocol settings.

mbercx: You did do it for the report ^^

It's not necessary to implement the same logic as was implemented for the `kpoints-distance` though, since the smearing settings are the same for all protocols and not likely to change soon.

unkcpz: > You did do it for the report ^^

True. You have a good memory. 
Anyway, I copy and paste the logic from kpoint-distance default, it should works well. Could you have a final look at the last commit and we merge this?

mbercx: I didn't test the latest changes, but it all looks good! Go ahead and merge. 🚀

PR #159, opened 2021-11-19
Running PDOS workchain and plot bands + pdos

### LINKED ISSUE #62: Support the aiida-quantumespresso `PdosWorkChain`
Currently the `PdosWorkChain` is not supported by the QEapp, limiting the properties to geometry and band structure. Since the work chain is there and very useful, we should add support for it ASAP.

There are still some issues on `aiida-quantumespresso` that need to be fixed, though, see for example:

https://github.com/aiidateam/aiida-quantumespresso/issues/685

unkcpz:
Fixes #62

- [x] The PR is blocked by https://github.com/aiidalab/aiidalab-widgets-base/issues/262 since pdos workchain have outputs in namespace.
- [x] Clean `qe.ipynb`
- [x] add checkbox and able to run PDOS workchain when the box checked
- [x] newly added `pdos` and `projwfc` code setting will break backward compatibility. 
- [x] newly added codes is not default set for the code dropdown widget.
- [x] Dependency: new `aiida-quantumespresso` plugin otherwise will raise `tetrahydro` setting error.
- [x] bands+pdos plot 
- [x] `show_band_structure` widget tab name change. When only one bands or pdos calculation run.
- [x]  aiida-quantumespresso 3.5.0 for the tetrahedra override

### DIFF
--- a/aiidalab_qe/node_view.py
+++ b/aiidalab_qe/node_view.py
@@ -428,7 +428,7 @@ class WorkChainViewer(ipw.VBox):
             layout=ipw.Layout(min_height="380px"),
         )
         self.bands_tab = ipw.VBox(
-            [ipw.Label("Band structure not available.")],
+            [ipw.Label("Electronic Structure not available.")],
             layout=ipw.Layout(min_height="380px"),
         )
         self.result_tabs = ipw.Tab(
@@ -437,7 +437,7 @@ class WorkChainViewer(ipw.VBox):
 
         self.result_tabs.set_title(0, "Workflow Summary")
         self.result_tabs.set_title(1, "Final Geometry (n/a)")
-        self.result_tabs.set_title(2, "Band Structure (n/a)")
+        self.result_tabs.set_title(2, "Electronic Structure (n/a)")
 
         # An ugly fix to the structure appearance problem
         # https://github.com/aiidalab/aiidalab-qe/issues/69
@@ -498,11 +498,14 @@ class WorkChainViewer(ipw.VBox):
         self.result_tabs.set_title(1, "Final Geometry")
 
     def _show_band_structure(self):
+        data = export_data(self.node)
+        bands_data = data.get("bands", None)
+        dos_data = data.get("dos", None)
         self._bands_plot_view = BandsPlotWidget(
-            bands=[export_bands_data(self.node)], plot_fermilevel=True
+            bands=[bands_data], dos=dos_data, plot_fermilevel=True
         )
         self.result_tabs.children[2].children = [self._bands_plot_view]
-        self.result_tabs.set_title(2, "Band Structure")
+        self.result_tabs.set_title(2, "Electronic Structure")
 
     def _show_workflow_output(self):
         self.workflows_output = WorkChainOutputs(self.node)
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -121,6 +121,15 @@ class WorkChainSettings(ipw.VBox):
             layout=ipw.Layout(max_width="10%"),
         )
 
+        # Checkbox to see if the PDOS should be calculated
+        self.pdos_run = ipw.Checkbox(
+            description="",
+            tooltip="Calculate the electronic PDOS.",
+            indent=False,
+            value=True,
+            layout=ipw.Layout(max_width="10%"),
+        )
+
         # Work chain protocol
         self.workchain_protocol = ipw.ToggleButtons(
             options=["fast", "moderate", "precise"],
@@ -137,6 +146,12 @@ class WorkChainSettings(ipw.VBox):
                 self.properties_title,
                 ipw.HTML("Select which properties to calculate:"),
                 ipw.HBox(children=[ipw.HTML("<b>Band structure</b>"), self.bands_run]),
+                ipw.HBox(
+                    children=[
+                        ipw.HTML("<b>Projected density of states</b>"),
+                        self.pdos_run,
+                    ]
+                ),
                 self.properties_help,
                 self.protocol_title,
                 ipw.HTML("Select the protocol:", layout=ipw.Layout(flex="1 1 auto")),
@@ -227,11 +242,35 @@ class CodeSettings(ipw.VBox):
                 "remote_abs_path": "/usr/bin/pw.x",
             },
         )
+        self.dos = CodeDropdown(
+            input_plugin="quantumespresso.dos",
+            description="dos.x:",
+            setup_code_params={
+                "computer": "localhost",
+                "description": "dos.x in AiiDAlab container.",
+                "label": "dos-6.7",
+                "input_plugin": "quantumespresso.dos",
+                "remote_abs_path": "/usr/bin/dos.x",
+            },
+        )
+        self.projwfc = CodeDropdown(
+            input_plugin="quantumespresso.projwfc",
+            description="projwfc.x:",
+            setup_code_params={
+                "computer": "localhost",
+                "description": "projwfc.x in AiiDAlab container.",
+                "label": "projwfc-6.7",
+                "input_plugin": "quantumespresso.projwfc",
+                "remote_abs_path": "/usr/bin/projwfc.x",
+            },
+        )
         super().__init__(
             children=[
                 self.codes_title,
                 self.codes_help,
                 self.pw,
+                self.dos,
+                self.projwfc,
             ],
             **kwargs,
         )
@@ -261,6 +300,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
         self.workchain_settings = WorkChainSettings()
         self.workchain_settings.relax_type.observe(self._update_state, "value")
         self.workchain_settings.bands_run.observe(self._update_state, "value")
+        self.workchain_settings.pdos_run.observe(self._update_state, "value")
 
         self.kpoints_settings = KpointSettings()
         self.pseudo_family_selector = PseudoFamilySelector()
@@ -273,6 +313,8 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
         self.codes_selector.pw.observe(
             self._set_num_mpi_tasks_to_default, "selected_code"
         )
+        self.codes_selector.dos.observe(self._update_state, "selected_code")
+        self.codes_selector.projwfc.observe(self._update_state, "selected_code")
         ipw.dlink(
             (self.workchain_settings.workchain_protocol, "value"),
             (self.kpoints_settings, "kpoints_distance_default"),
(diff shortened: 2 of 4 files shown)


mbercx:
@unkcpz `aiida-quantumespresso` 3.5.0 is now released with the fix you need for the `tetrahedra` override, so you can update the dependency.

unkcpz:
For the band structure output view widget tab name, it is only a good to have feature. So I think this PR is ready to review.

aiidalab_qe/steps.py
```diff
@@ -137,6 +146,7 @@ def __init__(self, **kwargs):
                 self.properties_title,
                 ipw.HTML("Select which properties to calculate:"),
                 ipw.HBox(children=[ipw.HTML("<b>Band structure</b>"), self.bands_run]),
+                ipw.HBox(children=[ipw.HTML("<b>PDOS</b>"), self.pdos_run]),
```
mbercx, line 149:
```suggestion
                ipw.HBox(children=[ipw.HTML("<b>Projected density of states</b>"), self.pdos_run]),
```

aiidalab_qe/node_view.py
```diff
@@ -498,11 +498,14 @@ def _show_structure(self):
...
+        data = export_data(self.node)
+        bands_data = data.get("bands", None)
+        dos_data = data.get("dos", None)
         self._bands_plot_view = BandsPlotWidget(
-            bands=[export_bands_data(self.node)], plot_fermilevel=True
+            bands=[bands_data], dos=dos_data, plot_fermilevel=True
         )
         self.result_tabs.children[2].children = [self._bands_plot_view]
-        self.result_tabs.set_title(2, "Band Structure")
+        self.result_tabs.set_title(2, "Band Structure and DOS")
```
mbercx, line 508: This doesn't seem to fit in the tab header anymore. 😅 Maybe use "Electronic Structure"? And/or we have to tweak the tab header widths.

Also note that the original tab title still is "Band structure (n/a)".

mbercx (changes requested):
Thanks @unkcpz! Just left some minor comments. Two other things:

1. The dependency version of `aiida-quantumespresso` doesn't seem to be upgraded to `3.5.0`?
2. When I refresh the app, the `dos.x` and `projwfc.x` codes weren't automatically set to the defaults. But this is most likely because I didn't install the app with this branch, since if I understand correctly the default are set upon installation.

aiidalab_qe/node_view.py
unkcpz: Yes, agree that the name 'Electronic Structure' is a good idea.

unkcpz:
> The dependency version of aiida-quantumespresso doesn't seem to be upgraded to 3.5.0?

Ah, true. I forget to push the latest commits with some other changes.

> I refresh the app, the dos.x and projwfc.x codes weren't automatically set to the defaults. But this is most likely because I didn't install the app with this branch, since if I understand correctly the default are set upon installation

This is an issue, I think the dos.x and projwfc.x were installed with pw.x. Just not sure which line of code influence this default fresh and setting behavior.  I need to check this more.

aiidalab_qe/steps.py
```diff
@@ -393,10 +444,11 @@ def _identify_submission_blockers(self):
         if not (
             self.workchain_settings.relax_type.value != "none"
             or self.workchain_settings.bands_run.value
+            or self.workchain_settings.pdos_run.value
         ):
             yield (
                 "The configured work chain would not actually compute anything. "
-                "Select either a structure relaxation method or the bands "
+                "Select either a structure relaxation method or the bands or the PDOS"
                 "calculation or both."
```
csadorf, line 452:
```suggestion
            "Select either a structure relaxation method or at least one of the "
            "the bands or the PDOS calculations or both."
```

aiidalab_qe/steps.py
```diff
@@ -433,10 +485,24 @@ def _auto_select_code(self, change):
...
+                    parameters["dos_code"]
+                )
+            except NotExistent:
+                pass
+            try:
+                self.codes_selector.projwfc.selected_code = load_code(
+                    parameters["projwfc_code"]
+                )
+            except NotExistent:
+                pass
```
csadorf, line 505: I would suggest that we move the utility `_load_code()` function from line 582/561 into the module namespace and then use it here as well for improved readability.

```suggestion
        self.codes_selector.pw.selected_code = _load_code(parameters["pw_code"])
        self.codes_selector.dos.selected_code = _load_code(parameters["dos_code"])
        self.codes_selector.projwfc.selected_code = _load_code(
            parameters["projwfc_code"]
        )
```

The function itself could be expressed a bit more succinctly like this:

```python
from contextlib import suppress

def _load_code(code):
    with suppress(NotExistent, ValueError):
        return load_code(None)
```

aiidalab_qe/steps.py
```diff
@@ -631,6 +711,9 @@ def submit(self, _=None):
         if not run_bands:
             builder.pop("bands")
 
+        if not run_pdos:
+            builder.pop("pdos")
```
csadorf, line 715: Do we actually need the `run_pdos` (and the `run_bands`) variable? And do we need to remove them from the `parameters` dict? This would also work, right?
```suggestion
        if not parameters["run_pdos"]:
            builder.pop("pdos")
```

csadorf (changes requested):

unkcpz:
I make all the changes you mentioned in the review, but I need to give it a final run and test. Will request another review after it.

aiidalab_qe/steps.py
unkcpz: Looks very cool! changed.

aiidalab_qe/steps.py
unkcpz: Will try whether not pop in from parameters works.

unkcpz:
> When I refresh the app, the dos.x and projwfc.x codes weren't automatically set to the defaults. But this is most likely because I didn't install the app with this branch, since if I understand correctly the default are set upon installation.

I am not sure since I create a new container with install the app with this branch but got the QE setup error `'NoneType' object has no attribute 'uuid'`.  @csadorf I am a bit lot in the `QESetupWidget` code. Could you tell me where this error might come from?

Moreover, I can not figure out how to retrigger the codes install and code setup or maybe just manually set the related parameters for code `dos` and `projwfc`? I don't think create a new container to get the code set is a convenient way to do this.

unkcpz:
The only block of this is the code selector not setting the default code as expected.

csadorf:
> The only block of this is the code selector not setting the default code as expected.

I will try to reproduce the issue.

csadorf:
I think we should have a bot that automatically generates the following command for us 😁 :
```console
aiidalab install quantum-espresso@git+https://github.com/unkcpz/aiidalab-qe@pdos --yes
```

csadorf:
I am still investigating, but I will start to try to answer some of your questions.

> > When I refresh the app, the dos.x and projwfc.x codes weren't automatically set to the defaults. But this is most likely because I didn't install the app with this branch, since if I understand correctly the default are set upon installation.
> 

I just installed this app version in a new environment and did not experience any issues with the qe installation. However, I also found that the codes were not selected by default.

> I am not sure since I create a new container with install the app with this branch but got the QE setup error `'NoneType' object has no attribute 'uuid'`. @csadorf I am a bit lot in the `QESetupWidget` code. Could you tell me where this error might come from? 

I am not able to reproduce this problem.

> The thing confuse me is in the method `SubmitQeAppWorkChainStep:_auto_select_code` it calls `get_input_parameters` which need the corresponding code selector being set?

Indeed, I am not exactly sure how that currently works either. I will keep investigating.

> Moreover, I can not figure out how to retrigger the codes install and code setup or maybe just manually set the related parameters for code `dos` and `projwfc`? I don't think create a new container to get the code set is a convenient way to do this.

To re-trigger the setup process just remove the AiiDA codes with the corresponding labels (pw-6.7@localhost etc.) and optionally also remove the `quantum-espresso-6.7` conda environment.

csadorf:
I can confirm that the pw.x code is currently not automatically selected, not even on the "master" branch. Trying to figure out when and how this regression was introduced.

csadorf:
Please rebase this branch after merging #167 .

csadorf (changes requested):
Please apply my suggestion on how to fix the remaining code selection issue and then rebase on master after #167 has been merged.

unkcpz:
Hi @csadorf, I have run `bumpver update` to create a new `tag: v22.01.0`. I need then push this tag? Or do this after pr https://github.com/aiidalab/aiidalab-qe/pull/169 merged?

csadorf:
> Hi @csadorf, I have run `bumpver update` to create a new `tag: v22.01.0`. I need then push this tag? Or do this after pr #169 merged?

As discussed offline, we should not bump the version on the feature branch. That said, `bumpver` should have pushed the tag, not sure why that did not happen.

csadorf (changes requested):
I tests this on the latest aiidalab version and everything appeared to work fine, however the DOS and the Projwfc calculations never appeared to fully finish:
Screen Shot 2022-01-10 at 12 42 01

That's despite the fact the logs appear completed:
Screen Shot 2022-01-10 at 12 42 08


I have one last question regarding the widgets base requirement, but other than that this is good to go.

setup.cfg
```diff
@@ -26,12 +26,12 @@ packages = find:
 install_requires =
     Jinja2~=2.11.3
     aiida-core~=1.0
-    aiida-quantumespresso~=3.4
+    aiida-quantumespresso~=3.5
     aiidalab-qe-workchain@https://github.com/aiidalab/aiidalab-qe/raw/v21.07.0b3/src/dist/aiidalab_qe_workchain-1.0-py3-none-any.whl
     aiidalab-widgets-base~=1.1
```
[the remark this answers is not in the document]
csadorf: This is almost certainly the case, see also issue: https://github.com/aiidalab/aiidalab-qe/issues/172

Because of this issue, I was not able to retrieve the previously completed workchain.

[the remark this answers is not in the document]
unkcpz: Thanks for testing @csadorf . I think this is the reason for the problem.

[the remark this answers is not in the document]
csadorf: @unkcpz Version 1.2.0 was just released: https://github.com/aiidalab/aiidalab-widgets-base/releases/tag/v1.2.0

```suggestion
    aiidalab-widgets-base~=1.2,>=1.2.0
```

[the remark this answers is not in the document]
unkcpz: Thanks! Changed, good to go.

csadorf (approved):
I reviewed the code and just ran a pdos work chain for testing on Si.
Screen Shot 2022-01-19 at 18 54 57

Everything appears to be working. 👍

PR #165, opened 2022-01-07
Update pre-commit config.

csadorf:

PR #166, opened 2022-01-07
Ci/automate release

csadorf:


csadorf:
Merge after #165.

.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+        steps:
+            - uses: actions/download-artifact@v2
+              name: Download distribution artifact
+              with:
+                  name: release
+                  path: dist/
+
+            - uses: softprops/action-gh-release@v0.1.14
+              name: Create release
+              if: startsWith(github.ref, 'refs/tags/v')
```
unkcpz, line 58: When this release action is triggered the workchain module is built, but where is it stored?

unkcpz: I see~ Is that mean every time the QEAppWorkchain changed, we need to increase to a new version? Can you add a note to README for maintainers about this.

unkcpz (changes requested):

.github/workflows/release.yml
csadorf: No, we only have to change this dependency line unless the work chain is actually changed. I am not sure how to address that in an automated manner.

I think the optimal solution would be to release the wheel onto the [github.com/aiidalab/packages](https://github.com/orgs/aiidalab/packages) once [Python packages are supported](https://github.com/github/roadmap/issues/94).

I will try to clarify as part of the README.

unkcpz (approved):
Thanks @csadorf Looks all good.

PR #167, opened 2022-01-07
Automatically select code after installation.

### LINKED ISSUE #162: Make choice of number of CPUs more obvious (and allow for optional advanced parallelisation and resources settings)
I'm always frustrated when I start (with AiiDAlab on my fast workstation, or connected to a supercomputer) a QE calculation with only 1 MPI process, that takes ages to finish any structure different than silicon :-)

- A request for the number of CPUs that is more conspicuous and I cannot miss (now it's hidden in Expert mode -> last tab -> bottom right corner). This should be always explicitly asked, until when we have good automatic selection of resources. I think even as an additional step between input selection (step 2) and the workflow results (step 3).
- Partially related to this: most QE calculations will be much slower if I don't use k-pools (`-nk` command like flag). Ideally this should be decided automatically by the workflow. But there should be a way for now for the user to decide the value. Maybe as an "expert flag" in the new step (asking the amount of resources).
- Once we are discussing advanced parallelisation features: pools is the most important one (when there are multiple kpoints); but we should probably then support also a few more of the parallelisation options of QE (number of tasks, ...), as well as a couple of resources more that it would be important to optionally be able to set (max wall time, max memory; and for running on CSCS, also some options for the scheduler (partition, account, ...)).

**Describe alternatives you've considered**
- For the choice of the # of CPUs I need to know where to look and do it actively
- For the number of pools (or other options) I think now I cannot set them via the GUI even if I wanted to?

### LINKED ISSUE #164: The pw.x code is not automatically selected
**Describe the bug**

After starting the app the first time, the QE installation is triggered, however the pw.x code is not automatically selected.

**To Reproduce**
1. Install the QE app in its latest stable release (`aiidalab install quantum-espresso --yes`).
2. Open the QE app.
3. Go to app step 2.
4. Wait for the setup processes to complete.
5. Inspect the code selection in the "Codes" tab of the expert mode.

**Expected behavior**

The code should be automatically selected after the setup process is completed.

**Screenshots**

Screen Shot 2022-01-07 at 14 44 53

**Version (if known)**
 - App version: v21.11.0

**Additional context**

This issue represents a regression.

csadorf:
Fixes #164.

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -430,10 +430,11 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
 
     def _auto_select_code(self, change):
         if change["new"] and not change["old"]:
-            parameters = self.get_input_parameters()
             self.codes_selector.pw.refresh()
             try:
-                self.codes_selector.pw.selected_code = load_code(parameters["pw_code"])
+                self.codes_selector.pw.selected_code = load_code(
+                    DEFAULT_PARAMETERS["pw_code"]
+                )
             except NotExistent:
                 pass
 


unkcpz (approved):
Ha, the code label is in this yaml file.

PR #176, opened 2022-01-17
Detail of wf info in workchain selector

### LINKED ISSUE #160: WorkChainSelector: show the properties the workchain calculate (relax, bands, pdos)
For `WorkChainSelector` it should show for each work chain which properties users are set to calculate. For example, if user choose to run band structure calculation plus pdos calculation on the structure without structure relaxation, the indicator show in selector panel will be like: `2D ago Finished[0] Si2 -- structure unrelax, properties on [band structure, pdos]`

unkcpz:
Fix #160


csadorf:
@unkcpz Please request my review when you think this is ready. Thanks.

csadorf (approved):

PR #182, opened 2022-01-24
IMPROVE: Rework resource specification

### LINKED ISSUE #79: 👌 IMPROVE: Set max resources based on computer
Currently the user can specify whatever resources he/she likes in the `ResourceSelectionWidget`. This can lead to calculations failing, but also to significant slowdowns when the code is run on the `localhost`. I see three things that we should improve here:

- [x] Allow the user to specify resources separately for each code.
- [x] Determine the max number of CPU's available. For the `localhost`, this can be achieved with `os.cpu_count()`. For the remote computers, this is usually specified as `mpiprocs_per_machine`.
- [x] Limit the amount of resources that can be specified:
  * Nodes: 1 for the `localhost`. For the remote computer, we should perhaps let this be configurable and stored as an extra?
  * CPUs: For the remotes, this should be `mpiprocs_per_machine` by default and also enforced as a maximum. The localhost should probably leave at least 1 CPU, so maybe `max(1, os.cpu_count() - 1)` should be the maximum?

Since we'll be touching this part of the code, I think it makes sense that we also:

- [x] Disable the `dos.x` and `projwfc.x` codes when the PDOS is not being calculated.

### LINKED ISSUE #83: ✨ NEW: Add parallelization options to "Resources" tab
Currently the parallelization options are not set automatically, and the user doesn't have a way to specify them. It would be good as a first rudimentary approach to set the `npools` to the number of nodes used for the calculation by default, as well as allow the user to specify the _basic_ parallelization options in the "Resources" tab.

### LINKED ISSUE #163: Number of MPI tasks ignored?
**Describe the bug**
I selected to run with 14 MPI processors, as shown here:

Screenshot 2021-12-21 at 16 15 28

but my calculation is still running "serially" (`mpirun -np 1`)... E.g. in the attributes of the CalcJob, I get:
```
    "resources": {
        "num_machines": 1
    },
````

**To Reproduce**
1. Run a new calculation, change the number of MPI tasks in the advanced tab
2. check the number of CPUs (either directly how many `pw.x` are running, or with `verdi node attributes PK` with the PK of the CalcJobNode).

**Expected behavior**
The app should use the number I specify, or stop with a clear message if it thinks the number is not appropriate.
(I'm not sure if #96 is related)?

**Version (if known)**
 - App version: v21.11.0, commit 0fd135b96277c6f99b49c750cc0bbeed6d31ff0d

mbercx:
Fixes #83 
Fixes #163
Fixes #79

Slight overhaul of the resource specification for the work chain:

* Disable the PDOS codes if the property is not calculated.
* Add a check to make sure that all codes are run on the same computer
if the PDOS is calculated.
* Allow the user to specify both #nodes and #CPUs for remote codes.
For the localhost codes the "nodes" widget is disabled. Defaults and
maxima of the number of CPUs is set based on the number of CPUs and the
default number of MPI processes.
* Allow to specify the parallelization. A default value is set to try
and avoid parallelization issues with too many MPI processes.
* Add `export OMP_NUM_THREADS=1` to `localhost` code setup.

Beside these changes, the builder is now properly updated to use the
specified resources and parallelization settings.

### DIFF
--- a/aiidalab_qe/setup_codes.py
+++ b/aiidalab_qe/setup_codes.py
@@ -82,7 +82,7 @@ def _setup_code(code_name, computer_name="localhost"):
                 "--computer",
                 computer_name,
                 "--prepend-text",
-                f"conda activate {CONDA_ENV_PREFIX}",
+                f"conda activate {CONDA_ENV_PREFIX}\nexport OMP_NUM_THREADS=1",
                 "--remote-abs-path",
                 CONDA_ENV_PREFIX.joinpath("bin", f"{code_name}.x"),
             ],
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -5,12 +5,12 @@ Authors:
 
     * Carl Simon Adorf <simon.adorf@epfl.ch>
 """
-from math import ceil
+import os
 
 import ipywidgets as ipw
 import traitlets
 from aiida.common import NotExistent
-from aiida.engine import ProcessState, submit
+from aiida.engine import ProcessBuilderNamespace, ProcessState, submit
 from aiida.orm import ProcessNode, WorkChainNode, load_code
 from aiida.plugins import DataFactory
 from aiida_quantumespresso.common.types import ElectronicType, RelaxType, SpinType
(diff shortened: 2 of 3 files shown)


csadorf (approved):
Code looks good to me and I did not encounter any issues during testing.

PR #190, opened 2022-01-25
FIX: Replace `ComputationalResourcesWidget.selected_code` by `value`

mbercx:


yakutovicha (approved):
all good, thanks @mbercx!

PR #191, opened 2022-01-25
Fix report of nscf k-distance

unkcpz:


src/aiidalab_qe_workchain/__init__.py
```diff
@@ -288,6 +288,7 @@ def run_pdos(self):
             inputs.nscf.pw.parent_folder = self.ctx.scf_parent_folder
         else:
             if "kpoints_distance_override" in self.inputs:
+                # scf? nscf?
                 inputs.scf.kpoints_distance = self.inputs.kpoints_distance_override
```
unkcpz, line 292: Hi @mbercx, I am not quite sure the code here is working as expected. It says in the spec that the `kpoints_distance_override` will replace the k-distance of nscf. But here it set the value of scf.

mbercx: The help says that:

https://github.com/aiidalab/aiidalab-qe/blob/e21836931a81a7d2bfd726a6a1e3c9f7d779d044/src/aiidalab_qe_workchain/__init__.py#L50

The "save for" is perhaps not the clearest language. Maybe "except" would have been simpler. ^^

mbercx (approved):
Thanks @unkcpz! Just remove the comment and this one is good to go.

src/aiidalab_qe_workchain/__init__.py
unkcpz: I see. So these two parameters of nscf are only able to be set from protocol.

> The "save for" is perhaps not the clearest language. Maybe "except" would have been simpler. ^^

Haha, a little big confusing. I change it to "except".

PR #193, opened 2022-01-25
Limit MPI ranks for `projwfc.x` and `dos.x`

mbercx:
Currently the `PdosWorkChain` runs the `projwfc.x` and `dos.x`
calculations with the full number of MPI ranks specified by the user,
without any kind of parallelization. This can cause failures when the
number of MPI ranks exceeds the number of FFT slabs in the
decomposition.

Here we use the same parallelization settings for the `projwfc.x`
calculation. For the `dos.x`, there is no parallellization implemented,
but this post-processing step is typically very fast, so we simply limit
the number of MPI ranks to the value that we used to set the default
number of `npools`.

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -702,11 +702,25 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
             self.projwfc_code.code_select_dropdown.disabled = True
 
     def submit(self, _=None):
-        def update_builder(buildy, resources, npools=None):
+        def update_builder(buildy, resources, npools):
+            """Update the resources and parallelization of the ``QeAppWorkChain`` builder."""
             for k, v in buildy.items():
                 if isinstance(v, (dict, ProcessBuilderNamespace)):
                     if k == "pw" and v["pseudos"]:
                         v["parallelization"] = Dict(dict={"npool": npools})
+                    if k == "projwfc":
+                        v["settings"] = Dict(dict={"cmdline": ["-nk", str(npools)]})
+                    if k == "dos":
+                        v["metadata"]["options"]["resources"] = {
+                            "num_machines": 1,
+                            "num_mpiprocs_per_machine": min(
+                                self.MAX_MPI_PER_POOL,
+                                self.resources_config.num_cpus.value,
+                            ),
+                        }
+                        # Continue to the next item to avoid overriding the resources in the
+                        # recursive `update_builder` call.
+                        continue
                     if k == "resources":
                         buildy["resources"] = resources
                     else:


aiidalab_qe/steps.py
```diff
@@ -703,10 +703,22 @@ def set_pdos_status(self):
 
     def submit(self, _=None):
         def update_builder(buildy, resources, npools=None):
```
unkcpz, line 705: What is the 'buildy'? I don't think this will affect by the scope of variable `builder` outside this local function. I guess you mean to say 'beauty' or 'buddy'? 🤣

aiidalab_qe/steps.py
```diff
@@ -703,10 +703,22 @@ def set_pdos_status(self):
...
+            """Update the resources and parallelization of the ``QeAppWorkChain`` builder."""
             for k, v in buildy.items():
                 if isinstance(v, (dict, ProcessBuilderNamespace)):
                     if k == "pw" and v["pseudos"]:
                         v["parallelization"] = Dict(dict={"npool": npools})
+                    if k == "projwfc":
+                        v["settings"] = Dict(dict={"cmdline": ["-nk", str(npools)]})
```
unkcpz, line 712: In this case if `npools` is `None`. Might be problematic.

aiidalab_qe/steps.py
```diff
@@ -703,10 +703,22 @@ def set_pdos_status(self):
...
+                    if k == "dos":
+                        v["metadata"]["options"]["resources"] = {
+                            "num_machines": 1,
+                            "num_mpiprocs_per_machine": min(
+                                self.MAX_MPI_PER_POOL,
+                                self.resources_config.num_cpus.value,
+                            ),
+                        }
+                        continue
```
unkcpz, line 721: Why `continue` the loop?

unkcpz (changes requested):
Thanks @mbercx

I give it a quick look. Going to have a run test after.

aiidalab_qe/steps.py
mbercx: It doesn't affect the scope, but I was still apprehensive towards using builder to distinguish it. Happy to change it to `builder` though. ^^

aiidalab_qe/steps.py
mbercx: Fair point! I had set `None` as the default for `npools` when I was testing this function, but forgot to remove it. Will do so now.

aiidalab_qe/steps.py
mbercx: To avoid that the `resources` of `dos` still get updated by the recursive `update_builder` call.

The logic is a bit convoluted now, I know. It can probably be improved. I just wanted to fix this ASAP since anyone that tests the QEapp on Eiger now will probably run into parallelization issues when running the PDOS.

aiidalab_qe/steps.py
unkcpz: What about moving the function out and using `builder`?

aiidalab_qe/steps.py
unkcpz: Thanks for the explanation.

> I just wanted to fix this ASAP since anyone that tests the QEapp on Eiger now will probably run into parallelization issues when running the PDOS.

I see, then we leave it as it is and add a detailed comment for this.

aiidalab_qe/steps.py
mbercx: The function was a standalone one at the top of the module before, but since it's only used here for now, I thought it better to move it into `submit` so it's easier to spot. Let's just leave `buildy` be, so the scope is clearly different.

Besides, buildy sounds cute, no? 😬

unkcpz: > Besides, buildy sounds cute, no? 

Indeed~

unkcpz:
Thanks @mbercx 
I give it a run on eiger and didn't find any issue. Merged.

PR #196, opened 2022-01-31
🐛 FIX: use primitive structure for PDOS `nscf` restart

### LINKED ISSUE #195: `QeAppWorkChain`: restart for PDOS `nscf` fails for non-primitive starting structure
**Describe the bug**

Currently, the `QeAppWorkChain` tries to reuse the charge density calculated for the band structure `bands` calculation for the `nscf` calculation of the `PdosWorkChain`. This fails in case the original structure is not already the primitive one, however, since SeeK-path will primitivize the structure before running the `scf` that calculates the charge density. Since the `RemoteData` folder of this calculation is passed as a `parent_folder` to the `nscf` calculation of the `PdosWorkChain`, but the structure is still passed from the `PwRelaxWorkChain`, there is a mismatch in the number of atoms specified in the `aiida.in` and the structure retrieved from the `data-file-schema.xml` and the calculation will fail with:

```
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine read_conf_from_file (1):
     bad number of atoms
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

**Expected behavior**

The work chain should either:

1. Calculate the charge density again for the relaxed structure
2. Use the primitive structure to calculate the PDOS

I'd use the second approach, since it's faster and I think it's also important to have the PDOS match the band structure. It's also easier to fix. 😁 

**Version (if known)**
 - App version: `master` branch

mbercx:
Fixes #195 

When both the band structure and PDOS are calculated, the
`QeAppWorkChain` will restart the `nscf` for the `PdosWorkchain` from
the `scf` calculation in the `PwBandsWorkChain`. However, the
`PwBandsWorkChain` will also primitivize the structure via Seek-path.
This meant that the input structure for the `nscf` calculation did not
match the one that Quantum ESPRESSO finds in the `data-file-schema.xml`
for the restart, causing the calculation to crash.

Here we set the `current_structure` in the work chain context to the
primitive one obtained from SeeK-path in the `inspect_bands` step of the
outline.

### DIFF
--- a/src/aiidalab_qe_workchain/__init__.py
+++ b/src/aiidalab_qe_workchain/__init__.py
@@ -262,10 +262,9 @@ class QeAppWorkChain(WorkChain):
             )
             return self.exit_codes.ERROR_SUB_PROCESS_FAILED_BANDS
 
-        scf = workchain.get_outgoing(
-            WorkChainNode, link_label_filter="scf"
-        ).all_nodes()[0]
+        scf = workchain.get_outgoing(WorkChainNode, link_label_filter="scf").one().node
         self.ctx.scf_parent_folder = scf.outputs.remote_folder
+        self.ctx.current_structure = workchain.outputs.primitive_structure
 
     def should_run_pdos(self):
         """Check if the projected density of states should be calculated."""


mbercx:
@csadorf I've tested this on the Al structure that @yakutovicha originally highlighted the issue for, and it seems to work fine. Will open a separate issue where we can discuss if we want to _always_ run the primitive structure etc.

csadorf (approved):
Looks fine to me. We will need a separate commit/release to install the revised work chain.

PR #199, opened 2022-02-01
Automatically update the workchain dependency.

csadorf:


src/README.md
```diff
@@ -10,24 +10,21 @@ In this way one can install the wheel directly from the GitHub release via a PEP
 
 ## How to update the `aiidalab_qe_workchain` package
 
-If changes to the workchain package cannot be avoided, this is how you can update it:
```
unkcpz, line 13: I suggest still keep the line `If changes to the workchain package cannot be avoided`. It is still mandatory to increase the version if there are changes to the workchain package.

unkcpz (commented):
Thanks @csadorf. I am going to finish the PR https://github.com/aiidalab/aiidalab-qe/pull/158. So to get the https://github.com/aiidalab/aiidalab-qe/pull/158 merged, I need to increase the version number in that PR, correct? Rather than separate the workchain modifications in another PR as it originally instructed?

src/README.md
csadorf: The previous sentence implies that changes to the work chain should be avoided. Since the procedure is now simplified, this is no longer the case.

I am not exactly sure what you mean by: 

> It is still mandatory to increase the version if there are changes to the workchain package.

If you want to release an update to the work chain you must make an app release, however that is not different to any other kind of change. The only difference is that you will have to manually install the work chain as described until said release was made.

unkcpz: I see. Yes, users who install the package from released PyPI will never have the problem of installing an incompatible QeWorkChain version.

unkcpz (approved):
I am good with the changes.

PR #207, opened 2022-02-22
FIX: Reporting when only PDOS is calculated

### LINKED ISSUE #204: Only calculating the PDOS shows incorrect settings
**Describe the bug**

When you only calculate the PDOS for a structure, the report shows `None` for some of the chosen parameters.

**To Reproduce**

Try to run a work chain that only calculates the PDOS

**Expected behavior**

The parameters should be properly extracted from the SCF of the `PdosWorkChain`, or we should rethink where these parameters are extracted from.

**Screenshots**

Screenshot 2022-02-21 at 20 08 44

**Version (if known)**

Latest `master`.

mbercx:
Fixes #204 

Fix two issues when calculating only the PDOS:

* Show the correct energy cutoffs and k-points distance for the SCF.
* Properly show the "Electronic Structure" output, which currently isn't
shown because the `_update_view` method only checks for the
`band_structure` output.

### DIFF
--- a/aiidalab_qe/node_view.py
+++ b/aiidalab_qe/node_view.py
@@ -116,12 +116,14 @@ def export_bands_data(work_chain_node):
             )[0]
         )
         data["fermi_level"] = work_chain_node.outputs.band_parameters["fermi_energy"]
-        return jsanitize(data)
+        return [
+            jsanitize(data),
+        ]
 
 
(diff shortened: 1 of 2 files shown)


unkcpz (approved):
@mbercx Thanks! Look all good. 

Also test with running some test with only pdos, no issues found.

PR #212, opened 2022-03-03
Fix: Leftover reset of input_structure in ConfigureStep

unkcpz:
There is no such attribute `input_structure` of class `ConfigureQeAppWorkChainStep`. Although this cause nothing when reset in the configure step (I am surprised this does not raise any exception🧐), anyway should be deleted.


csadorf (approved):
Why are you surprised that this didn't cause an exception?

unkcpz:
> Why are you surprised that this didn't cause an exception?

Well, not a surprise anymore. Just intuitively confused seems found no this attribute in the class. This is python 😆, an assignment expression without declaration totally what it is.

PR #213, opened 2022-03-07
Deps: update widget-bandsplot to 0.2.8

unkcpz:
Thanks for @dou-du 's work, we can now distinguish up and down spins in red and black lines when showing the band structure.


mbercx (approved):
Let's plot some spin-polarised band structures, baby! 🚀

PR #214, opened 2022-03-10
Configure confirm button disabled when clicked

unkcpz:
In configure step, after the `confirm` button is clicked and goes to the next step, the button should be disabled. 
This is a very trivial issue that literally impacts nothing if the user does not go back to modify the workchain setting once confirmed. 

I think here we need to decide whether we allowed users to modify the workchain setting once confirmed. If we allow them to do so, once they change settings the button should be activated again. It is currently update for `relax_type`, `bands_run`, `dos_run`, but not for the widgets that not callback `_update_state`.

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -427,7 +427,7 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
             self.set_input_parameters(DEFAULT_PARAMETERS)
 
     def confirm(self, _=None):
-        self.confirm_button.disabled = False
+        self.confirm_button.disabled = True
         self.state = self.State.SUCCESS
 
     @traitlets.default("state")


csadorf (changes requested):
Have a look at the `._update_state()` function. The _confirm_ button _disabled_ state is updated there. I assume something is going wrong?

Also, it is not a good idea to manually change the step state as is done in line 431. The state should be a function of the general widget state since it might get updated again by the `._update_state()` function.

unkcpz:
I think I misunderstood the expected behavior of how every step connects and how they transfer to other steps. 
If I understand correctly, the design was: click confirm -> next step. If we want to reset previous step, user should: click `previous step` -> reset -> click confirm again.
The issue now is when `Reset` is clicked, it pop to the structure setting step and asks the user to do it from the beginning. Which I think the expected behavior should be reset all the settings of current step. Do you agree? @csadorf 

> Also, it is not a good idea to manually change the step state as is done in line 431. The state should be a function of the general widget state since it might get updated again by the ._update_state() function.

I see no way to improve this. Since the `confirm` method has its only function is to actively set state to `SUCCESS` while the `_update_state` is more detecting and updating state passively.

csadorf:
> I think I misunderstood the expected behavior of how every step connects and how they transfer to other steps. If I understand correctly, the design was: click confirm -> next step. If we want to reset previous step, user should: click `previous step` -> reset -> click confirm again. The issue now is when `Reset` is clicked, it pop to the structure setting step and asks the user to do it from the beginning. Which I think the expected behavior should be reset all the settings of current step. Do you agree? @csadorf

Yes, the reset function should reset the app up until the currently selected step.

> > Also, it is not a good idea to manually change the step state as is done in line 431. The state should be a function of the general widget state since it might get updated again by the ._update_state() function.
> 
> I see no way to improve this. Since the `confirm` method has its only function is to actively set state to `SUCCESS` while the `_update_state` is more detecting and updating state passively.

Since the state should generally be a function of the step trait state, clicking on confirm should modify some trait, e.g., copy the work chain settings from `workchaing_settings` to `confirmed_workchain_settings`. [This is how it is done in the first step](https://github.com/aiidalab/aiidalab-qe/blob/0f92d845b84c50804a65e12b98163a705a198cc5/aiidalab_qe/structures.py#L45-L46), where we have a `.structure` and a `confirmed_structure `trait. Or simply have another trait called `settings_confirmed` or so that is toggled.

PR #215, opened 2022-03-14
Add select from Example structures widget back

unkcpz:
Add the `StructureExamplesWidget` back to structure manager widget
with more examples.


unkcpz:
Happy to add more example structures if you can provide the files.

csadorf (commented):
Are all of these structures well-tested, i.e., do we know that they run robustly and yield the correct results?

unkcpz:
Should be all good after the spin-polarization PR merged, which failed the Cobalt calculation. I will give all structures final tests late today.

unkcpz:
New examples are all working fine, should be safe to merged. Test on Au.cif which is not a primitive cell and Co.cif which has spin-polarization turned on and others are old examples the regular insulators.

csadorf (approved):

PR #216, opened 2022-03-21
Add basic structure editor to the app.

yakutovicha:


unkcpz:
@yakutovicha thanks for the effort of the widget. But I think this one is more compatible with the molecules rather than for the periodic systems that qe-app deals with more frequently. I will take it as a base class and see how I can extend or modified it to fit the qe-app.

yakutovicha:
> @yakutovicha thanks for the effort on the widget. But I think this one is more compatible with the molecules rather than for the periodic systems that qe-app deals with more frequently. I will take it as a base class and see how I can extend or modified it to fit the qe-app.

Are you sure? The idea behind the widget is to do standard editing operations that do not depend on periodicity.

unkcpz:
I am not sure. But some concepts such as functional groups and mirror reflect by axis are not so common in condensed matter materials (not precisely the definition of 'condensed matter'), since now the QE only supports structures with periodic [True, True, True]. 

I think what I expected from this widget in QE context are for example 1. primitive cell <-> conventional cell conversion,  2. unit cell transformation with the transformation matrix. 3. lattice parameters modifier.

yakutovicha:
> I am not sure. But some concepts such as functional groups and mirror reflect by axis are not so common in condensed matter materials (not precisely the definition of 'condensed matter'), since now the QE only supports structures with periodic setting equal to `[True, True, True]`.

Well, this is used quite a lot in Empa. The operations of moving, rotating and copying atoms are not specific to molecules. Not forgetting, that the periodic requirement you mentioned there is a temporary one.


> I think what I expected from this widget in QE context are for example 1. primitive cell <-> conventional cell conversion, 2. unit cell transformation with the transformation matrix. 3. lattice parameters modifier.

Those should be different editors, I believe. For example, I plan to add a spin assignment as a separate editor.

unkcpz:
@yakutovicha thanks! Do you have an app that already works quite well with this widget? I test it with the QE and don't know the issue I encountered is from the widget itself or the qe-app or the structure I uploaded is not compatible with the widget yet. I'd better have a proper taste of how this widget is supposed to be😁.

yakutovicha:
> @yakutovicha thanks! Do you have an app that already works quite well with this widget?

I think many of the other apps use it: `aiidalab-lsmo`, `nanoribbons`, `surfaces`, `empa-molecules`.

>  I test it with the QE and don't know the issue I encountered is from the widget itself or the qe-app or the structure I uploaded is not compatible with the widget yet. I'd better have a proper taste of how this widget is supposed to be😁.

Hmm, strange. It worked fine for me. But let me know what the issue is.

unkcpz:
> Hmm, strange. It worked fine for me. But let me know what the issue is.

Here is one https://github.com/aiidalab/aiidalab-qe/issues/217. I upload the structure from cif file. And I am not able to see any changes from the structure viewer widget when I click any button in the structure editor widget. Then this error pops up.

yakutovicha:
I cannot reproduce it. I tried with the attached structure, and it all went fine.
I used file [MON.txt](https://github.com/aiidalab/aiidalab-qe/files/8332973/MON.txt) (rename `MON.txt` with`MON.cif`). Can you share your CIIF file maybe?

unkcpz:
[Ag.txt](https://github.com/aiidalab/aiidalab-qe/files/8340869/Ag.txt)
Hi @yakutovicha, this is the structure I use.

yakutovicha:
> [Ag.txt](https://github.com/aiidalab/aiidalab-qe/files/8340869/Ag.txt) Hi @yakutovicha, this is the structure I use.

I had 0 issues editing it:
Screenshot 2022-03-24 at 12 01 05

yakutovicha:
@unkcpz is there anything I could do to advance this?

unkcpz:
@yakutovicha Sorry, I miss your message yesterday. 
I think the issue is that I did not click to choose any atoms before I click 'move to xyz'. We just a warning for this instead of raising an exception.

unkcpz:
I play around with it a bit more, and all works well for me.

unkcpz:
Hi @yakutovicha, I think the issues about this widget should all be fixed on aiidalab-base-widget side. I personally would like to merge this after we release a version with the spin-polarized issue fixed. If you don't mind, I'll keep this PR open for a bit longer, but all is good from my side with adding this widget.

unkcpz (approved):
I rebased the branch. Looks all good to me. I'll open an issue in ABW for the issue that if no atoms choose it will fail once using the editor.

yakutovicha:
For the information, the issue is https://github.com/aiidalab/aiidalab-widgets-base/issues/309

PR #223, opened 2022-04-11
Fix downstream CI and add manually triggered downstream CI test

unkcpz:


unkcpz:
I accidentally push a fix of downstream ci fix (untested) to the master, since as admin we are not limited. The owner permission setting is reset to as owner of the repo we also need to push to unprotected branch first instead of push to master directly.

csadorf (changes requested):
I don't think this test should run on every push, should it?

unkcpz:
> I don't think this test should run on every push, should it?

No, It shouldn't, I added it to trigger it manually. I change it back and after this merged, we can go to CI action and trigger the test manually, hopefully?

tests/test_qe_app.py
```diff
@@ -16,8 +16,8 @@ def test_qe_app_select_silicon(selenium, url):
     selenium.set_window_size(1920, 985)
     time.sleep(10)
     selenium.find_element(
-        By.XPATH, '//li[@id="tab-key-17" and @class="lm-TabBar-tab p-TabBar-tab"]'
-    ).click()  # click `From Examples` for input structure
+        By.XPATH, '//li[@id="tab-key-17"]'
+    ).click()  # click `From Examples` tab for input structure
```
csadorf, line 20: This looks like an unrelated change. Do you want to merge it with this PR?

unkcpz: It is the fix of a previously wrongly pushed commit https://github.com/aiidalab/aiidalab-qe/commit/f2d840de219a8d2ab70b9811652ff0c936338323. I test locally this change will pass the downstream CI.

I think we can merge it with this PR and I then specify the details about this commit.

csadorf: For next time I would recommend to explicitly revert f2d840d and then apply the additional change. You could still do this here and then either rebase and merge or squash merge.

csadorf (approved):

PR #232, opened 2022-05-13
aiidalab-registry expect list for categories in setup.cfg

unkcpz:
Metadata read by aiidalab-registry expect a list for `categories` in `setup.cfg`. This is subtle mistake and hard to find when writing metadata.
We'd better then add an example in the documentation for this field. Or you think it is possible to add a schema check somewhere for it? @csadorf 

EDIT: Add my name to the author list.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ description = Package for the AiiDAlab QE app
 long_description = file: README.md
 long_description_content_type = text/markdown
 url = https://github.com/aiidalab/aiidalab-qe
-author = Carl Simon Adorf, Aliaksandr Yakutovich, Marnik Bercx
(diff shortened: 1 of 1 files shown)


csadorf:
> We'd better then add an example in the documentation for this field. Or you think it is possible to add a schema check somewhere for it? @csadorf

Both. We could offer a pre-commit hook that automatically checks the schema.

csadorf (approved):

unkcpz:
@csadorf thanks. As for the documentation for this field, it must be on the aiidalab documentation. But for the schema check pre-commit, we can add it here for aiidalab-qe repository, but it is still not check for all the apps that registered. `aiidalab-registry` will only fail on deploy CI.

PR #251, opened 2022-07-05
Refactor installation routines and allow to run them from the CLI.

csadorf:


csadorf:
Depends on https://github.com/aiidalab/aiidalab/pull/295 .

aiidalab_qe/setup_codes.py
```diff
@@ -281,6 +277,10 @@ def _update(self, change):
 
             if self.error or self.installed:
                 self._progress_bar.value = 1.0
+            elif self.busy:
+                self._progress_bar.value = ProgressBar.AnimationRate(1.0)
```
unkcpz, line 281: What is the rate equal to `1.0` means? Can you add a comment in the code?

unkcpz (commented):
@csadorf thanks! The codes all look good. 
Since this change depend on the `aiidalab`, I think we need to also add `aiidalab` dependency in the `setup.cfg` file? 

I still need to test this PR thoroughly both with CLI and install qe from setup widget.

aiidalab_qe/setup_codes.py
```diff
@@ -98,6 +98,64 @@ def setup_codes():
...
+                    install_qe()
+                except CalledProcessError as error:
+                    raise RuntimeError(f"Failed to create conda environment: {error}")
+
+            # After installing QE, we install the corresponding
+            # AiiDA codes:
+            for code_name in CODE_NAMES:
+                if not _code_is_setup(code_name):
+                    yield f"Setting up AiiDA code ({code_name})..."
+                    _setup_code(code_name)
```
unkcpz, line 147: After installing and setup of the code finished successfully, the `FN_DO_NOT_SETUP` need to be removed? Or it is already somewhere else?

csadorf: The `FN_DO_NOT_SETUP` file prevents the installation. It is only created if the setup failed. That is to prevent the setup to make an attempt to perform the installation every time the app is opened.

aiidalab_qe/setup_codes.py
csadorf: Sure, would that be sufficient?
```suggestion
                # Start progress bar animation at constant rate (1.0).
                self._progress_bar.value = ProgressBar.AnimationRate(1.0)
```

csadorf:
> Since this change depend on the aiidalab, I think we need to also add aiidalab dependency in the setup.cfg file?

It does not, what made you think that?

unkcpz:
> > Since this change depend on the aiidalab, I think we need to also add aiidalab dependency in the setup.cfg file?
> 
> It does not, what made you think that?

Because the `post_install` is added here which requires https://github.com/aiidalab/aiidalab/pull/295 to be merged first.

aiidalab_qe/setup_codes.py
unkcpz: Hmm, I still don't get the intuitive idea of what the `1.0` mean and in what unit?

unkcpz (approved):
@csadorf Thanks! It is all good from my side. Go ahead if you like.

aiidalab_qe/setup_codes.py
```diff
@@ -98,6 +98,64 @@ def setup_codes():
...
+                )
+
+            if not qe_installed():
+
+                # First, install Quantum ESPRESSO.
+                yield "Installing QE..."
+                try:
+                    install_qe()
+                except CalledProcessError as error:
+                    raise RuntimeError(f"Failed to create conda environment: {error}")
```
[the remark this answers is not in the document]
csadorf: Yes, that's correct. I left it in there for backwards compatibility. But we could also remove it and require a higher aiidalab version. What would you prefer?

We would still need some mechanism that blocks the app until the installation is completed either way.

csadorf:
> > > Since this change depend on the aiidalab, I think we need to also add aiidalab dependency in the setup.cfg file?
> > 
> > 
> > It does not, what made you think that?
> 
> Because the `post_install` is added here which requires [aiidalab/aiidalab#295](https://github.com/aiidalab/aiidalab/pull/295) to be merged first.

Yes, but the installation is still triggered like before if post_install is ignored. But we could also require a higher aiidalab version if you prefer that.

aiidalab_qe/setup_codes.py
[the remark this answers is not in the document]
unkcpz: > But we could also remove it and require a higher aiidalab version. What would you prefer?

I see problem now. The `aiidalab` is installed along with container in `/opt/conda`, and we don't want different apps set this dependency which will potentially cause incompatible problems. Then I agree with your implementation that will also works before https://github.com/aiidalab/aiidalab/pull/295 merged.

unkcpz (approved):

aiidalab_qe/setup_codes.py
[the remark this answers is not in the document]
csadorf: Great, then let's leave it as is.

Issue #259, opened 2022-08-08
Bug report: Application crashed with KeyError


csadorf:
## Automated report

_This issue was created with the app's automated bug reporting feature.
Attached to this issue is the full traceback as well as an environment
fingerprint that contains information about the operating system as well as all
installed libraries._

## Additional comments (optional):

I submitted a pure-relaxation calculation for Silica.

## Attachments

<details>
<summary>Traceback</summary>

```python-traceback
~/apps/quantum-espresso/aiidalab_qe/widgets.py in _observe_node(self, change)
    376                 clear_output()
    377                 if change["new"]:
--> 378                     display(viewer(change["new"]))
    379 
    380 

~/.local/lib/python3.8/site-packages/aiidalab_widgets_base/viewers.py in viewer(obj, downloadable, **kwargs)
     77             )
     78             return obj
---> 79         raise exc
     80 
     81 

~/.local/lib/python3.8/site-packages/aiidalab_widgets_base/viewers.py in viewer(obj, downloadable, **kwargs)
     69     try:
     70         _viewer = AIIDA_VIEWER_MAPPING[obj.node_type]
---> 71         return _viewer(obj, downloadable=downloadable, **kwargs)
     72     except (KeyError) as exc:
     73         if obj.node_type in str(exc):

~/apps/quantum-espresso/aiidalab_qe/node_view.py in __init__(self, node, **kwargs)
    427             """
    428         )
--> 429         self.workflows_summary = SummaryView(self.node)
    430 
    431         self.summary_tab = ipw.VBox(children=[self.workflows_summary])

~/apps/quantum-espresso/aiidalab_qe/node_view.py in __init__(self, wc_node, **kwargs)
    193             return "Yes" if truthy else "No"
    194 
--> 195         report = generate_report_dict(self.wc_node)
    196 
    197         env = Environment()

~/apps/quantum-espresso/aiidalab_qe/report.py in generate_report_dict(qeapp_wc)
    132 def generate_report_dict(qeapp_wc):
    133     """Generate a dictionary for reporting the inputs for the `QeAppWorkChain`"""
--> 134     return dict(_generate_report_dict(qeapp_wc))
    135 
    136 

~/apps/quantum-espresso/aiidalab_qe/report.py in _generate_report_dict(qeapp_wc)
     31     # Properties
     32     run_relax = builder_parameters["relax_type"] != "none"
---> 33     run_bands = builder_parameters["run_bands"]
[...]

Issue #205, opened 2022-02-22
CI: Downstream-integration tests are failing


csadorf:
The [_scheduled_ downstream-integration workflow](https://github.com/aiidalab/aiidalab-qe/actions/workflows/di.yml) is currently failing. 

@unkcpz I would recommend to fix it and disable the workflow until then.

csadorf:
Yes, we probably need to revise the action. Without any additional test scripts, it primarily acts as a smoke test, where it tries to open the app in the browser and checks that it loaded without any issues. This will catch 80% of critical failure cases related to dependency incompatibilities and so on.

unkcpz:
Mmore subtle issue about downstream CI test. The qe-app can't be loaded by using the `[aiidalab-test-app-action](https://github.com/aiidalab/aiidalab-test-app-action)`.

csadorf:
The goal is to be able to run `pytest` inside the repository root, and it will automatically run all tests.

Proposed approach for implementing tests:
1. Copy the approach for setting up a test environment from the [aiidalab-docker-stack test config](https://github.com/aiidalab/aiidalab-docker-stack/blob/main/tests/conftest.py) with [pytest-docker](https://pypi.org/project/pytest-docker/) (test against `docker.io/aiidalab/full-stack:edge`).
2. Install the tested app by mounting the repository (/the GitHub workspace directory) and adding it to the AIIDALAB_DEFAULT_APPS environment variable (see example below).
3. Implement basic smoke tests (is app directory present, is app reported installed by aiidalab etc.)
4. Implement interface tests with selenium, you can get the url from the [`notebook_service` fixture](https://github.com/aiidalab/aiidalab-docker-stack/blob/7f89fde47696ce2d80e1c959da5a1181fe2b9a27/tests/conftest.py#L30).

```yml
---
# tests/docker-compose.yml
version: '3.4'
services:
    aiidalab:
        image: aiidalab/full-stack:edge
        ports:
            - 8888
        volumes:
            - ${PWD}:/home/jovyan/workspace
        env:
            AIIDALAB_DEFAULT_APPS: quantum-espresso@file:///home/jovyan/workspace
```

unkcpz:
This will be fixed by #271 where I implement the basic test along the way suggested.

PR #260, opened 2022-08-08
Fix: Pure relaxation fails

unkcpz:
This is a quick fix for #259, but when I look at the code, I find there could be some problem with the parameters not properly set from `qeapp.yaml`. 
So if it is not extremely emergence, I'll dig into it more and update this PR. What you thought? @csadorf

### DIFF
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -30,7 +30,7 @@ def _generate_report_dict(qeapp_wc):
 
     # Properties
     run_relax = builder_parameters["relax_type"] != "none"
-    run_bands = builder_parameters["run_bands"]
+    run_bands = builder_parameters.get("run_bands", False)
     run_pdos = builder_parameters.get("run_pdos", False)
 
     yield "relaxed", run_relax


csadorf:
> This is a quick fix for #259, but when I look at the code, I find there could be some problem with the parameters not properly set from `qeapp.yaml`. So if it is not extremely emergence, I'll dig into it more and update this PR. What you thought? @csadorf

I would recommend to release the fix immediately since this bug brakes a core function of the app. As a next step I would recommend to

1. Identify what caused this bug.
2. Implement tests that would catch the issue in the future.

Were you able to reproduce the problem and confirm that this fix solves it or should I test it myself?

unkcpz:
I mess up my local qeapp aiidalab environment, need some time to fix and test it. I reproduce issue will online aiidalab but can not check if this PR fixes that.

unkcpz:
> Implement tests that would catch the issue in the future.

How are we going to add a test for this? We literally have zero tests for anything of this app.

csadorf:
> > Implement tests that would catch the issue in the future.
> 
> How are we going to add a test for this? We literally have zero tests for anything of this app.

[That is not true](https://github.com/aiidalab/aiidalab-qe/blob/master/tests/test_qe_app.py), they are just not automatically executed at the moment. It is not too hard to implement additional tests with selenium.

unkcpz:
> [That is not true](https://github.com/aiidalab/aiidalab-qe/blob/master/tests/test_qe_app.py), they are just not automatically executed at the moment. It is not too hard to implement additional tests with selenium.

I mean the smoke test is now turned off which will actually trigger the test. Could you have a look at it if you get the time?

csadorf:
> I mean the smoke test is now turned off which will actually trigger the test. Could you have a look at it if you get the time?

Sure, can you create an issue, please?

unkcpz:
There is an issue for that https://github.com/aiidalab/aiidalab-qe/issues/205. I add it to the roadmap.

PR #262, opened 2022-08-08
Set installed traitlets of pseudo/qe install widget when it's done

### LINKED ISSUE #261: Critical: the SSSP library is not installed
After the PR https://github.com/aiidalab/aiidalab-qe/pull/251 there is a issue that `SsspInstallWidget` can not update the state to `installed` even the pseudo libraries are properly installed. 
The reason is that `SSSPInstallWidget` once updates the `installed` traitlets in `_refresh_installed` but not in the new implementation. 
I think the way to fix this is to return the installed state from function `install()` and update the traitlets.

unkcpz:
fixes #261

### DIFF
--- a/aiidalab_qe/setup_codes.py
+++ b/aiidalab_qe/setup_codes.py
@@ -212,8 +212,6 @@ class QESetupWidget(ipw.VBox):
             for msg in install():
                 self.set_message(msg)
 
-            self.installed = True
-
         except Exception as error:
             self.set_message("Failed to setup QE on localhost.")
             self.set_trait("error", str(error))
--- a/aiidalab_qe/setup_codes.py
+++ b/aiidalab_qe/setup_codes.py
@@ -219,6 +219,7 @@ class QESetupWidget(ipw.VBox):
             self.set_trait("error", str(error))
             FN_DO_NOT_SETUP.touch()
         else:
+            self.set_trait("installed", True)
             self.set_message("OK")
         finally:
             self.set_trait("busy", False)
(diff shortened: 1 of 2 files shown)


aiidalab_qe/setup_codes.py
```diff
@@ -219,6 +219,7 @@ def _refresh_installed(self):
             self.set_trait("error", str(error))
             FN_DO_NOT_SETUP.touch()
         else:
+            self.set_trait("installed", True)
```
csadorf, line 222: This isn't necessary since the trait is already set on line 215.

aiidalab_qe/sssp.py
```diff
@@ -102,6 +102,7 @@ def _refresh_installed(self):
             self.set_trait("error", str(error))
             self.set_message(str(error))
         else:
+            self.set_trait("installed", True)
```
csadorf, line 105: To be safe I'd recommend:
```suggestion
            self.set_trait("installed", not bool(pseudos_to_install()))
```

csadorf (changes requested):

aiidalab_qe/setup_codes.py
unkcpz: That's true, didn't notice that. I remove line 215 to align with the `install()` of `sssp.py`

aiidalab_qe/sssp.py
unkcpz: Good advice! If all the libraries are install by hands `bool(pseudos_to_install())` will be `False` empty list.

csadorf (approved):

PR #263, opened 2022-08-09
properties selecting using toggle buttons

### LINKED ISSUE #293: All selection elements should ideally be buttons to make the interface a bit more homogeneous, that means all exclusive choices are button groups and the "toggle" widgets that are currently check boxes become toggle buttons.

unkcpz:
fixes #293


unkcpz:
Close this one since I think it is not necessary to use toggle buttons for properties, which may cause more confusion.

PR #269, opened 2022-08-16
CalcJobOutputFollower: Fetch retrieved output file if available

danielhollas:
When the CalcJob is finished, we can fetch the output file from the `retrieved` folder instead of always going through the `remote_folder`. Besides performance reasons, this works better when the remote folder is cleared.

Note that I have tested this only in the context of my own AiiDAlab app where I am reusing this code, but I see no reason why it shouldn't work here as well (please test).

### DIFF
--- a/aiidalab_qe/widgets.py
+++ b/aiidalab_qe/widgets.py
@@ -296,7 +296,15 @@ class CalcJobOutputFollower(traitlets.HasTraits):
 
     def _fetch_output(self, calcjob):
         assert isinstance(calcjob, CalcJobNode)
-        if "remote_folder" in calcjob.outputs:
+        if "retrieved" in calcjob.outputs:
+            try:
+                self.filename = calcjob.attributes["output_filename"]
+                with calcjob.outputs.retrieved.open(self.filename) as f:
+                    return f.read().splitlines()
+            except OSError:
+                return list()
+
+        elif "remote_folder" in calcjob.outputs:
             try:
                 fn_out = calcjob.attributes["output_filename"]
                 self.filename = fn_out


danielhollas:
@csadorf I believe you did this part of the code in case you want to review this.

csadorf (approved):
LGTM, but I would appreciate if @unkcpz could also have a quick look at this. Thanks!

aiidalab_qe/widgets.py
```diff
@@ -296,7 +296,15 @@ def _follow_output(self, calcjob):
 
     def _fetch_output(self, calcjob):
         assert isinstance(calcjob, CalcJobNode)
-        if "remote_folder" in calcjob.outputs:
+        if "retrieved" in calcjob.outputs:
+            try:
+                self.filename = calcjob.attributes["output_filename"]
+                with calcjob.outputs.retrieved.open(self.filename) as f:
```
unkcpz, line 302: I think this is based on the `output_filename` in the retrieved list, in this case, the `aiida.out` (or maybe it is the default file to retrieve, can you double check? @danielhollas ). It is okay for the pw.x calculation. But if we want to make this widget more general for example move into widgets-base in the future, it is better to check that the `output_filename` is in the retrieve list.

unkcpz (commented):
Thanks, @danielhollas. Only a minor request if we want this widget more generic for other calcjob.

aiidalab_qe/widgets.py
unkcpz: This should be fine for the QE app, we can consider this after we want to make this widget more general by moving it to AWB in the future.

PR #270, opened 2022-08-16
Fail if process exits with non-zero status

danielhollas:
Follow up on https://github.com/aiidalab/aiidalab-widgets-base/pull/336, see discussion there.

(I tested this in my own app, but suppose it works here too)

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -927,9 +927,12 @@ class ViewQeAppWorkChainStatusAndResultsStep(ipw.VBox, WizardAppWidgetStep):
                 ProcessState.WAITING,
             ):
                 self.state = self.State.ACTIVE
-            elif process_state in (ProcessState.EXCEPTED, ProcessState.KILLED):
+            elif (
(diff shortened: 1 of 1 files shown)


unkcpz (approved):
@danielhollas thanks! Looks all good to me.

aiidalab_qe/steps.py
```diff
@@ -927,9 +927,12 @@ def _update_state(self):
                 ProcessState.WAITING,
             ):
                 self.state = self.State.ACTIVE
-            elif process_state in (ProcessState.EXCEPTED, ProcessState.KILLED):
+            elif (
+                process_state in (ProcessState.EXCEPTED, ProcessState.KILLED)
+                or self.process.is_failed
+            ):
```
unkcpz, line 933: Are these all cases of `not self.process.is_finished_ok`? But this is more explicit and all good from my side.

danielhollas: Yes, as far as I know. Thanks!

unkcpz:
Cheers!

PR #271, opened 2022-08-29
Mig/2.1

### LINKED ISSUE #272: QE app: Support AiiDA core 2.0
fixed by #271

unkcpz:
- [x] Update dependency (block by AWB dep update https://github.com/aiidalab/aiidalab-widgets-base/pull/344)
- [x] Update QE conda version (better to 7.1 but no conda forge supported yet)
- [x] bump aiida-quantumespreeso version to 4.0.0
- [x] Test

fixes #272


unkcpz:
Since it is able to install the qe code and sssp library by [`post-install`](https://github.com/aiidalab/aiidalab/issues/290), and the aiidalab full-stack have that feature build-in. 
We do not need to install the qe code the first time App start. 
Therefore, I simply using `auto_start=False` to disable the auto_start of the code setup which will get rid of the thread issue. 
But the thread issue should anyway be solved and the code setup widget need to be decoupled and refactored to the AWB. I open a dedicate issue for it https://github.com/aiidalab/aiidalab-qe/issues/299.

tests/helper_dep_requirements.py
```diff
@@ -0,0 +1,11 @@
...
+
+from configparser import ConfigParser
+
+cf = ConfigParser()
+cf.read("setup.cfg")
+lst = cf["options"]["install_requires"].split("\n")
+requirements_lst = [i for i in lst if not i.startswith("aiidalab-qe-workchain")]
+
+with open("/tmp/requirements.txt", "w") as fh:
+    fh.write("\n".join(requirements_lst[1:]))
```
unkcpz, line 11: This script is run inside the test container in order to properly install the `aiidalab-qe-workchain` from the current folder. 
Since the `setup.cfg` pinned its version, the `aiidalab-qe-workchain` can not installed by `pip install -U .` directly for test. Otherwise it is not the version for test.

src/setup.cfg
```diff
@@ -18,8 +18,8 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    aiida-core~=1.0
-    aiida-quantumespresso~=3.2
+    aiida-core~=2.0
+    aiida-quantumespresso~=4.0
```
unkcpz, line 22: Need to be further update for the `aiida-pseudo` CLI issue.

danielhollas:
@unkcpz FYI by turning on the deprecation warnings by setting `AIIDA_WARN_v3=1` env variable, I identified a few more things in my app that should be upgraded, such as deprecated `set_extra` and `get_extra` methods and some others. They are also relevant here, you might want to take a look at this commit https://github.com/ispg-group/aiidalab-ispg/pull/80/commits/26157c17f6e2f351a453df06de0ad84b1185b81e

btw: AiiDA v2.1.2 has been released yesterday.

unkcpz:
@danielhollas thanks for the advice. But how to turn on the config exactly, are the warning messages will show on the notebook? But the shell environment variables only have an effect on the shell that trigger it, correct?

danielhollas:
Yes, it is a bit tricky. Since I am building my own Docker image, I have simply set the environment variable in the Dockerfile.

unkcpz:
Seems it can be set for the notebook by ipython magic https://stackoverflow.com/questions/37890898/how-to-set-env-variable-in-jupyter-notebook. I'll give it a try tomorrow.

sphuber:
> I wonder if maybe there is AiiDA API to trigger this (AIIDA_WARN_v3=1)? @sphuber

It is just an environment variable, so I guess you could do
```python
import os
os.env['AIIDA_WARN_v3'] = True
```

unkcpz:
Hi @superstar54, after the release of aiida-quantumespresso, I think there is no block for testing with this branch, can you test it again? Thanks!

src/setup.cfg
unkcpz: These deps need to be updated aiida-core~=2.1 and qe~=4.1

qe.ipynb
```diff
@@ -29,11 +45,74 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
    "metadata": {
     "scrolled": false
    },
-   "outputs": [],
+   "outputs": [
```
unkcpz, line 62: Clean up this nb file.

tests/helper_dep_requirements.py
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python3
+
+from configparser import ConfigParser
+
```
unkcpz, line 4: Add a description DOC for this file.

unkcpz:
> the pdos above the fermi level shift a little bit away from the axis, as shown below:

We have updated the widget-bandsplot, but in another PR. @dou-du and I make a lot changes of to the widget, so maybe the problem solved naturally. We can test it there.

superstar54:
@unkcpz I think we can merge this PR after you clean up the nb file. 

The test only covers step 1 (structure selection). However, since I did all tests manually, I think it should be fine. In the future, tests on steps 2~4 should also be added.

qe.ipynb
danielhollas: @unkcpz FYI I am running `nbstripout` to get rid of these outputs as a pre-commit hook, see

https://github.com/kynan/nbstripout](https://github.com/ispg-group/aiidalab-ispg/blob/00f054b675902f2f5100d36e8858dae76d465257/.pre-commit-config.yaml#L14

unkcpz: Interesting, thanks I'll give it a try.

dou-du:
> I tested this PR by running a few examples with Si (diamond) structure.
> 
>     * scf only
> 
>     * scf + bands + pdos
> 
>     * relax
> 
>     * relax + bands + pdos
> 
> 
> All the tests passed. Great!
> 
> There are two points which could be improved in the feature.
> 
>     * After the calculation, I need to refresh the page to load the results (structure and bands).
> 
>     * the pdos above the fermi level shift a little bit away from the axis, as shown below:
> 
> 
> 

This bug has been fixed in our latest released version.

unkcpz:
> The test only covers step 1 (structure selection). However, since I did all tests manually, I think it should be fine. In the future, tests on steps 2~4 should also be added.

I open the issue in https://github.com/aiidalab/aiidalab-qe/issues/316

The failed test is fixed. Somehow the firefox test failed because firefox is not installed properly. Since @superstar54 already mentioned the PR is ready. I merge it without approval.

.github/workflows/di.yml
```diff
@@ -44,5 +44,16 @@ jobs:
               run: |
                   pip install -U -r requirements_test.txt
 
+            - name: Set jupyter token env
+              run: echo "JUPYTER_TOKEN=$(openssl rand -hex 32)" >> $GITHUB_ENV
+
+            - uses: browser-actions/setup-firefox@latest
+              with:
+                  firefox-version: ${{ matrix.firefox }}
+              run: which firefox
```
danielhollas, line 53: @unkcpz I think this line broke this whole file and tests are not running.

https://github.com/aiidalab/aiidalab-qe/actions/runs/3527890535

obrazek

unkcpz: You are right, I simply see All test pass. My fault, will open a PR to fix it.

PR #274, opened 2022-09-01
Set relax type none and no properites to run SCF only

### LINKED ISSUE #264: Allowed to run SCF calculation only
Redirected from #245. For the properties to calculate if no properties are chosen a warning message is raised and blocks to go next step. It should be that even not selecting relaxation or bands or pdos should still run: then the user can still download the QE output file.

Besides changing the widget to suppress the submission blocker, we need change the logic of the QeAppWorkchain to allowed only the run SCF calculation. Now when `relax`, `bands`, and `pdos` all not set for input ports the workchain will skip to the final step.  @mbercx any idea how we organise the `QeAppWorkChain`? I think we can either add one more `if_` for SCF only or run SCF from relax or bands sub-workchain.

unkcpz:
fixes #264 

Using `RelaxType.NONE` for relax workchain to run the SCF only. 
The builder is constructed with all three workchain input parameters port, `relax`, `bands`, `pdos`.
Whether to run  sub-workchain is controlled by popping the input key out from the builder, right after the builder is created. 
Also, the relax/properties infos of workchain selector is changed to show this newly add calculation type as 'SCF only'.

### DIFF
--- a/aiidalab_qe/process.py
+++ b/aiidalab_qe/process.py
@@ -82,22 +82,25 @@ class WorkChainSelector(ipw.HBox):
         )
 
         for process in projected[1:]:
-            pk = process[0]
-            formula = load_node(pk).inputs.structure.get_formula()
+            workchain = load_node(process[0])
+            formula = workchain.inputs.structure.get_formula()
 
             properties = []
-            if "pdos" in load_node(pk).inputs:
+            if "pdos" in workchain.inputs:
                 properties.append("pdos")
-            if "bands" in load_node(pk).inputs:
+            if "bands" in workchain.inputs:
                 properties.append("bands")
 
-            if "relax" not in load_node(pk).inputs:
+            if "relax" in workchain.inputs:
+                builder_parameters = workchain.get_extra("builder_parameters", {})
+                relax_type = builder_parameters.get("relax_type")
+
+                if relax_type != "none":
+                    relax_info = "structure is not relaxed"
+                else:
+                    relax_info = "static SCF calculation on structure"
+            else:
                 relax_info = "structure is not relaxed"
-            if "relax" in load_node(pk).inputs and properties:
-                relax_info = "structure is relaxed"
-            if "relax" in load_node(pk).inputs and not properties:
-                relax_info = "structure is not relaxed"
-                properties = ["SCF only"]
 
             if not properties:
                 properties_info = ""
--- a/aiidalab_qe/process.py
+++ b/aiidalab_qe/process.py
@@ -96,7 +96,7 @@ class WorkChainSelector(ipw.HBox):
                 relax_type = builder_parameters.get("relax_type")
 
                 if relax_type != "none":
-                    relax_info = "structure is not relaxed"
+                    relax_info = "structure is relaxed"
                 else:
                     relax_info = "static SCF calculation on structure"
             else:
--- a/aiidalab_qe/process.py
+++ b/aiidalab_qe/process.py
@@ -90,7 +90,7 @@ class WorkChainSelector(ipw.HBox):
                 properties.append("pdos")
             if "bands" in load_node(pk).inputs:
                 properties.append("bands")
-                
+
             if "relax" not in load_node(pk).inputs:
                 relax_info = "structure is not relaxed"
             if "relax" in load_node(pk).inputs and properties:
@@ -103,7 +103,7 @@ class WorkChainSelector(ipw.HBox):
                 properties_info = ""
             else:
                 properties_info = f"properties on {', '.join(properties)}"
-                
+
             yield cls.WorkChainData(
                 formula=formula,
                 relax_info=relax_info,
--- a/aiidalab_qe/process.py
+++ b/aiidalab_qe/process.py
@@ -84,21 +84,26 @@ class WorkChainSelector(ipw.HBox):
         for process in projected[1:]:
             pk = process[0]
             formula = load_node(pk).inputs.structure.get_formula()
-            if "relax" in load_node(pk).inputs:
-                relax_info = "structure is relaxed"
-            else:
-                relax_info = "structure is not relaxed"
 
             properties = []
             if "pdos" in load_node(pk).inputs:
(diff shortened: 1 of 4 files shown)


superstar54:
One comment, after the SCF calculation, what results could the user get? Currently, the `energy` and `force` are not printed out in the result section. Better to implement this also. Otherwise, the calculation is meaningless.

superstar54:
It is also confusing that one runs a `SCF` calculation using the `relax` port. Better to add a `SCF` workchain, to make the code clean and readable.

unkcpz:
Thanks! @superstar54 

> after the SCF calculation, what results could the user get? Currently, the energy and force are not printed out in the result section. Better to implement this also. Otherwise, the calculation is meaningless.

This is a very good point. But this issue came from the in-person testing with Nicola that when selecting not relax and no properties are chosen the qeApp will block the calculation from running by throwing a warning. 
Is that possible to leave it in the future since we do not decide yet what to show as the output of the SCF only calculation? I'll open an issue for it. 

> It is also confusing that one runs a SCF calculation using the relax port. Better to add a SCF workchain, to make the code clean and readable.

I also had same feeling when start working on this. But qeApp is simply the UI wrapper of aiida-quantumespresso, I think the change should be make there. @mbercx can you comment on if you think this is worth implementing? 
I think we can basically add an alias workchain called `PwScfOnlyWorkChain` sort of, and have it point to Relax with the parameters fixed. The question is where to put this workchain? If it is not a case to add in aiida-quantumespresso, I think it is also make sense to have the workchain here in QeApp/src. I'll add it in this PR.

unkcpz:
Hi @superstar54, I change the message show in workflow bar. But I still not add the energies showed in the output tabs. Since I don't decide yet where to add it, it can be on in the `Workflow summary` or as a new tab with all related and necessary output parameters. I open an issue at https://github.com/aiidalab/aiidalab-qe/issues/311. 
Let's focus on it with another PR.

mbercx:
Re the SCF only output: I think one of the possible outputs Nicola had in mind would be the charge density. However, I don't think it's possible to extract the charge density file atm. I'll add that originally we blocked this type of calculation for this very reason: I saw little point in running only an SCF for a layman user. The total energy of a single calculation is also not super meaningful, I'd say? The forces _could_ be useful in some cases, I suppose.

Re the relax work chain: An "SCF" work chain would simply be the `PwBaseWorkChain`, no? The `RelaxType.NONE` implementation comes from the common workflows project, there it was decided to add this relax type for technical reasons, and here we simply repurpose this to make the implementation somewhat easier. I'd have to take a closer look at the code (which I'll try to make some time for ASAP, many apologies @unkcpz!), but I think you'd also have to make quite some significant changes to the `QeAppWorkChain` to add this to a proper (optional) port. Not sure if the resulting code would be easier to maintain.

unkcpz:
@mbercx Thanks! I update the #311 with your comment. I think for the current one, it is okay to use the PwRelaxWorkChain with RelaxType.None. Since in my opinion, we should not add complexity to the workchain of QeApp but just light wrapper of what already implement in the `aiida-quantumespresso`. 

@superstar54 could you have another look and let's try get this PR done?

aiidalab_qe/process.py
```diff
@@ -82,23 +82,31 @@ def find_work_chains(cls):
...
-            if "bands" in load_node(pk).inputs:
+            if "bands" in workchain.inputs:
                 properties.append("bands")
 
+            if "relax" in workchain.inputs:
+                builder_parameters = workchain.get_extra("builder_parameters", {})
+                relax_type = builder_parameters.get("relax_type")
+
+                if relax_type != "none":
+                    relax_info = "structure is not relaxed"
```
superstar54, line 99:
```suggestion
                    relax_info = "structure is relaxed"
```

aiidalab_qe/process.py
```diff
@@ -82,23 +82,31 @@ def find_work_chains(cls):
...
                 properties.append("pdos")
-            if "bands" in load_node(pk).inputs:
+            if "bands" in workchain.inputs:
                 properties.append("bands")
 
+            if "relax" in workchain.inputs:
+                builder_parameters = workchain.get_extra("builder_parameters", {})
+                relax_type = builder_parameters.get("relax_type")
+
+                if relax_type != "none":
```
superstar54, line 98: Why is "none", not `None` or "None". Please make sure "none" is correct.

If it is correct, we need to change it to a more standard way, e.g. `None`

aiidalab_qe/report.py
```diff
@@ -98,27 +99,27 @@ def _generate_report_dict(qeapp_wc):
         # read default from protocol
         smearing = default_params["smearing"]
 
-    if run_relax:
+    try:
```
superstar54, line 102: I think this `try ... except` is not used, because it will always run a relax (scf or real relax) calculation.

aiidalab_qe/steps.py
```diff
@@ -878,6 +865,12 @@ def update_builder(buildy, resources, npools):
         if "smearing_override" in parameters:
             builder.smearing_override = Str(parameters["smearing_override"])
 
+        # skip relax sub-worflow only when RelaxType is NONE and has property calculated.
```
superstar54, line 868: Could you explain the difference between relax (only scf) and relax sub-workflow? I am not very clear about this, thanks.

superstar54, line 868:
```suggestion
        # skip relax sub-workflow only when RelaxType is NONE and has property calculated.
```

superstar54 (changes requested):

aiidalab_qe/steps.py
unkcpz: You can find how this tag change the parameters set in `pw.x` input file at https://github.com/aiidateam/aiida-quantumespresso/blob/d15f8cd1032e37ae59a19d733ee5ae1ed59e1f10/src/aiida_quantumespresso/workflows/pw/relax.py#L123-L146

superstar54 (approved):
LGTM

PR #276, opened 2022-09-01
Not pop run_bands/pdos from builder parameters

### LINKED ISSUE #275: The ecutrho is not shown in the workchain report when run with bands or dos, but without relax
**Describe the bug**

Energy cutoff not show if running bands/pdos calculation without relax

**Expected behavior**

**Screenshots**
image

**Version (if known)**
 - App version: [e.g. v22.08.0]

unkcpz:
fixes #275 

The `run_bands` and `run_dos` are poped out from builder_parameters which is not needed and will then not be set as extra attributes of workchain.
It cause the issue that the report of the calculation is not correctly prepared.

### DIFF
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -29,9 +29,9 @@ def _generate_report_dict(qeapp_wc):
     builder_parameters = qeapp_wc.get_extra("builder_parameters", {})
 
     # Properties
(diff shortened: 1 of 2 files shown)


mbercx (approved):
LGTM! 🚀

unkcpz:
Thanks @mbercx, cheers!

PR #297, opened 2022-09-21
Change spin/electronic type setting to toggle

### LINKED ISSUE #293: All selection elements should ideally be buttons to make the interface a bit more homogeneous, that means all exclusive choices are button groups and the "toggle" widgets that are currently check boxes become toggle buttons.

unkcpz:
fixes #293

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -96,17 +96,17 @@ class WorkChainSettings(ipw.VBox):
             ],
             value="positions_cell",
         )
-
+        
         # SpinType: magnetic properties of material
-        self.spin_type = ipw.Dropdown(
-            options=[("Non-magnetic", "none"), ("Ferromagnetic", "collinear")],
+        self.spin_type = ipw.ToggleButtons(
+            options=[("Off", "none"), ("On", "collinear")],
             value=DEFAULT_PARAMETERS["spin_type"],
             description="Magnetism:",
             style={"description_width": "initial"},
         )
-
+        
         # ElectronicType: electronic properties of material
-        self.electronic_type = ipw.Dropdown(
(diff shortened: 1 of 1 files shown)


superstar54:
Could you please add a screenshot that quickly shows the UI before and after this PR? 
I don't really agree with the idea that changing all choices to the toggle button. What if there are more choices, eg. ten options, will there be ten buttons?

unkcpz:
Hi @superstar54, thanks for the quick reply. Here is the screenshot

Before:
image

After:
image

> I don't really agree with the idea that changing all choices to the toggle button. What if there are more choices, eg. ten options, will there be ten buttons?

This is the feedback from Nicola actually. The widget for parameters setting need to be uniformed otherwise not very convenient for users. For this case explicitly, I think we don't have too many choices, the electronic type will influence how the [`occupation` for pwscf](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm348), `smearing` for metal and `fixed` for insulator. The magnism will be none or collinear, all options are defined and can be find from the protocol https://github.com/aiidateam/aiida-quantumespresso/blob/main/src/aiida_quantumespresso/common/types.py

superstar54:
Thanks for the details. In this case, with only two options, I agree that the toggle button is better.

One more question. Is it easy to align left for the buttons: `Metal` and `Off`? It would look nicer if they were aligned.

unkcpz:
@superstar54 that's a good point. It can but more or less increase the complexity of the code. Here it initially uses the `{"description_width": "initial"}` for the description of the widget as we usually set it for a simple widget. If we want more control, have to separate the description out as the `ipw.Label` as I did in the new commit. 
But I think this is a fair change. Please check the screenshot below for what it looks like.
image

superstar54 (approved):
Great!

PR #306, opened 2022-10-21
band structure use dos fermi energy if exist

### LINKED ISSUE #268: Fermi energy of DOS and bands are not compatible and alignment
**Describe the bug**
The electronic structure band structure and pdos the fermi energy are incompatible, the band structure use the fermi energy at VBM while pdos at CBM for some cases for example SiO2 from MC3D (id=7564).

**To Reproduce**
1. Submit calculation for band structure and pdos on SiO2 (MC3D id=7564)
2. Check electronic structure

**Screenshots**
image

**Version (if known)**
 - App version: [e.g. v22.08.0]

**Additional context**
Band structure fermi energy read from `work_chain_node.outputs.bands_parameters["fermi_energy"]` while dos fermi energy read from `work_chain_node.outputs.nscf_parameters["fermi_energy"]` they are not always the same.

unkcpz:
fixes #268 

If both bands and dos calculations are run, using the fermi energy from nscf results of dos calculation. Otherwise using the fermi energy from bands calculation is not scientifically robust.

### DIFF
--- a/aiidalab_qe/node_view.py
+++ b/aiidalab_qe/node_view.py
@@ -109,79 +109,82 @@ class MinimalStructureViewer(ipw.VBox):
 
 
 def export_bands_data(work_chain_node, fermi_energy=None):
-    data = json.loads(
-        work_chain_node.outputs.band_structure._exportcontent("json", comments=False)[0]
-    )
-    # The fermi energy from band calculation is not robust.
-    data["fermi_level"] = (
-        fermi_energy or work_chain_node.outputs.band_parameters["fermi_energy"]
-    )
-    return [
-        jsanitize(data),
-    ]
-
-
-def export_pdos_data(work_chain_node, fermi_energy):
-    _, energy_dos, energy_units = work_chain_node.outputs.dos.get_x()
-    tdos_values = {f"{n} | {u}": v for n, v, u in work_chain_node.outputs.dos.get_y()}
-
-    pdos_orbitals = []
-
-    if "projections" in work_chain_node.outputs:
-        projection_list = [
-            (work_chain_node.outputs.projections, None),
+    if "band_structure" in work_chain_node.outputs:
+        data = json.loads(
+            work_chain_node.outputs.band_structure._exportcontent(
+                "json", comments=False
+            )[0]
+        )
+        # The fermi energy from band calculation is not robust.
+        data["fermi_level"] = (
+            fermi_energy or work_chain_node.outputs.band_parameters["fermi_energy"]
+        )
+        return [
+            jsanitize(data),
         ]
     else:
-        projection_list = [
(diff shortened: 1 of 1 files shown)


aiidalab_qe/node_view.py
```diff
@@ -108,74 +108,84 @@ def _update_structure_viewer(self, change):
...
+        dos = export_pdos_data(work_chain_node, fermi_energy)
+    else:
+        fermi_energy = None
+        dos = None
+
+    if "band_structure" in work_chain_node.outputs:
+        # using the fermi energy from dos nscf if calculated.
+        bands = export_bands_data(work_chain_node, fermi_energy)
+    else:
+        bands = []
```
superstar54, line 184: If there is no band data, set it to `None`.
```suggestion
        bands = None
```

aiidalab_qe/node_view.py
```diff
@@ -108,74 +108,84 @@ def _update_structure_viewer(self, change):
...
 
-def export_pdos_data(work_chain_node):
-    if "dos" in work_chain_node.outputs:
-        fermi_energy = work_chain_node.outputs.nscf_parameters["fermi_energy"]
-        _, energy_dos, energy_units = work_chain_node.outputs.dos.get_x()
-        tdos_values = {
-            f"{n} | {u}": v for n, v, u in work_chain_node.outputs.dos.get_y()
-        }
+def export_pdos_data(work_chain_node, fermi_energy):
+    _, energy_dos, energy_units = work_chain_node.outputs.dos.get_x()
```
superstar54, line 125: I don't understand why you remove this:
```
if "dos" in work_chain_node.outputs:
```
We should make sure that the `dos` does exist before fetching the data, right?

superstar54 (changes requested):

aiidalab_qe/node_view.py
unkcpz: I move the `if "dos"` to `export_data` below. I think it would be more easy to maintain if the if-else checks are in one place.

aiidalab_qe/node_view.py
unkcpz: Yes, that's true. Thanks!

aiidalab_qe/node_view.py
superstar54: This will make the `export_pdos_data` function incomplete. If one calls the function from other places, there is a risk that they will get an error. Better to keep `if` here.

Then in the `export_data` part, you can remove the `if` by:
```
dos = export_pdos_data(work_chain_node)
fermi_energy = dos["fermi_energy"] if dos else None
```

superstar54 (changes requested):

superstar54 (approved):
LGTM!

PR #307, opened 2022-10-21
Adapt to support uuid traitlets

unkcpz:
after https://github.com/aiidalab/aiidalab-widgets-base/pull/375, these changes are needed here.

### DIFF
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -11,7 +11,7 @@ import ipywidgets as ipw
 import traitlets
 from aiida.common import NotExistent
 from aiida.engine import ProcessBuilderNamespace, ProcessState, submit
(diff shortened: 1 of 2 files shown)


unkcpz:
> I wonder if the CalcJobNodeViewer and its dependencies shouldn't be merged to AWB as it is quite useful, and there seems to be some overlap with existing widgets in aiidalab_widgets_base/process.py

I am fully on it. I think the initial idea from @csadorf is we need to first have the widget implemented in App and after finding it useful and having APIs well-defined, move the widget to the aiidalab-widgets-base. The widgets shared by qeApp and your app match this condition. I'll decouple it from here.

unkcpz:
Since this is anyway needed and tested in the Mig2.x. I close this PR and put all things to https://github.com/aiidalab/aiidalab-qe/pull/271

PR #332, opened 2022-12-12
Bump aiidalab-widgets-base to 2.0.0 beta

unkcpz:


setup.cfg
```diff
@@ -26,7 +26,7 @@ install_requires =
     aiida-core~=2.1
```
danielhollas, line 26:
```suggestion
    aiida-core>=2.1,<3
```

This should be quite safe I think. Otherwise you'll need to update this each time we update aiida-core in the Docker image.

danielhollas (commented):
One suggestion.

setup.cfg
unkcpz: It is safe, but I think `~=2.1` will install `2.2` if it is released and the same as what you proposed the `aiida-core>=2.1,<3`, no?

danielhollas: I am never sure about this, but I think in this case you need to use the carot, not tilde.

```suggestion
    aiida-core^=2.1
```

https://stackoverflow.com/a/22345808

yakutovicha: Why are we discussing this, by the way? The PR is about AWB, I guess :)

yakutovicha (approved):

setup.cfg
danielhollas: > But the link you share is for npm package, for python we have PEP440 and please check https://peps.python.org/pep-0440/#compatible-release.
Also https://peps.python.org/pep-0440/#examples the examples.

Oh, that's a very good point. Now I understand the source of my confusion, I didn't realize there's a difference. Thanks!

danielhollas (approved):

PR #333, opened 2022-12-12
The api change after bump the widget base

unkcpz:
I just made a test and find I forget the update when using the new `aiidalab-widget-base`.


unkcpz:
I am wondering this too! Will try 😄 
But can we get this merged first? Since I want to release a version that can be tested before we announce this to THEOS group for further tests. There is a deadline, have to do these sorts of nasty hotfixes.

danielhollas (approved):

PR #345, opened 2023-02-06
Final screenshot fixture

### LINKED ISSUE #343: Implement screenshot fixture as AWB take screenshot anyway for every test cases

unkcpz:
fixes #343 

Using the fixture `final_screenshot` to take the screenshot for smoke tests at the end of tests. 
In this PR, I also try to use the `WebDriverWait` to load the element without explicitly waiting.

### DIFF
--- a/tests/test_qe_app.py
+++ b/tests/test_qe_app.py
@@ -26,16 +26,13 @@ def test_qe_app_select_silicon_and_confirm(
 ):
     driver = selenium_driver("qe.ipynb", wait_time=30.0)
     driver.set_window_size(1920, 1485)
-    time.sleep(15)
-
-    driver.find_element(
-        By.XPATH, "//*[text()='From Examples']"
-    ).click()  # click `From Examples` tab for input structure
 
     element = WebDriverWait(driver, 60).until(
-        EC.element_to_be_clickable((By.XPATH, "//option[@value='Diamond']"))
+        EC.presence_of_element_located((By.XPATH, "//*[text()='From Examples']"))
     )
     element.click()
+
+    driver.find_element(By.XPATH, "//option[@value='Diamond']").click()
 
     driver.get_screenshot_as_file(
         str(Path.joinpath(screenshot_dir, "qe-app-select-diamond-selected.png"))
--- a/tests/test_qe_app.py
+++ b/tests/test_qe_app.py
@@ -27,12 +27,13 @@ def test_qe_app_select_silicon_and_confirm(
     driver = selenium_driver("qe.ipynb", wait_time=30.0)
     driver.set_window_size(1920, 1485)
 
-    element = WebDriverWait(driver, 60).until(
+    element = WebDriverWait(driver, 10).until(
         EC.presence_of_element_located((By.XPATH, "//*[text()='From Examples']"))
     )
     element.click()
 
     driver.find_element(By.XPATH, "//option[@value='Diamond']").click()
+    time.sleep(10)
 
     driver.get_screenshot_as_file(
         str(Path.joinpath(screenshot_dir, "qe-app-select-diamond-selected.png"))
--- a/tests/test_qe_app.py
+++ b/tests/test_qe_app.py
@@ -2,7 +2,9 @@ import time
 from pathlib import Path
 
 import requests
+import selenium.webdriver.support.expected_conditions as EC
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support.wait import WebDriverWait
 
 
 def test_notebook_service_available(notebook_service):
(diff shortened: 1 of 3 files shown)


tests/test_qe_app.py
```diff
@@ -11,29 +13,38 @@ def test_notebook_service_available(notebook_service):
...
 
-def test_qe_app_select_silicon(selenium_driver, screenshot_dir):
+def test_qe_app_select_silicon_and_confirm(
+    selenium_driver,
+    screenshot_dir,
+    final_screenshot,
+):
     driver = selenium_driver("qe.ipynb", wait_time=30.0)
     driver.set_window_size(1920, 1485)
+    time.sleep(15)
```
danielhollas, line 29: Why is this sleep needed? The element below should wait implicitly, no?

tests/test_qe_app.py
```diff
@@ -11,29 +13,38 @@ def test_notebook_service_available(notebook_service):
...
+
     driver.find_element(
         By.XPATH, "//*[text()='From Examples']"
     ).click()  # click `From Examples` tab for input structure
-    driver.find_element(By.XPATH, "//option[@value='Diamond']").click()
-    time.sleep(10)
+
+    element = WebDriverWait(driver, 60).until(
```
danielhollas, line 35: Same question here, is this fixing some flakiness that you saw?

danielhollas (approved):

tests/test_qe_app.py
unkcpz: Since the below I didn't use `WebDriverWait` for "from examples", somehow the From Examples is not clickable. I'll try again.

tests/test_qe_app.py
unkcpz: I don't quite understand what you mean, this `60` is the timeout, it will check if the element is valid in a certain period (500ms if I remember correctly).

danielhollas: The `find_element` waits implicitly (we set implicit wait in conftest), so I don't think this code is an improvement. What may happen I guess is that the element is located, but for some reason is not yet clickable. So I think you should use `EC.element_to_be_clickable` (which you already do below) instead of merely `EC.presence_of_element_located`.

unkcpz: I see your point. I changed all find_element to wait at first commit without too much thinking. You are correct, for the select of Structure, this explicit wait is needless. Changed.

unkcpz: However, for the "From examples", if I understand correctly, the `WebDriverWait` is the explicit wait in order to make sure the "From examples" tab is valid, it is not the same as the implict wait we set in conftest.

tests/test_qe_app.py
```diff
@@ -11,29 +13,35 @@ def test_notebook_service_available(notebook_service):
...
-        By.XPATH, "//*[text()='From Examples']"
-    ).click()  # click `From Examples` tab for input structure
+
+    element = WebDriverWait(driver, 60).until(
+        EC.presence_of_element_located((By.XPATH, "//*[text()='From Examples']"))
+    )
+    element.click()
+
     driver.find_element(By.XPATH, "//option[@value='Diamond']").click()
-    time.sleep(10)
```
danielhollas, line 28: I think you need to sleep here, although probably less than 10s, to ensure the molecule is displayed for the screenshot below.

tests/test_qe_app.py
```diff
@@ -11,29 +13,35 @@ def test_notebook_service_available(notebook_service):
...
+    screenshot_dir,
+    final_screenshot,
+):
     driver = selenium_driver("qe.ipynb", wait_time=30.0)
     driver.set_window_size(1920, 1485)
-    driver.find_element(
-        By.XPATH, "//*[text()='From Examples']"
-    ).click()  # click `From Examples` tab for input structure
+
+    element = WebDriverWait(driver, 60).until(
```
danielhollas, line 30:
```suggestion
    element = WebDriverWait(driver, 10).until(
```
I'd say 60 is an overkill (same below) but up to you.

tests/test_qe_app.py
```diff
@@ -11,29 +13,35 @@ def test_notebook_service_available(notebook_service):
...
+        EC.presence_of_element_located((By.XPATH, "//*[text()='From Examples']"))
```
danielhollas, line 31: Why not

```suggestion
        EC.element_to_be_clickable((By.XPATH, "//*[text()='From Examples']"))
```

danielhollas (changes requested):

tests/test_qe_app.py
danielhollas: I don't understand what you mean. From examples tab should be clickable as well, no?

tests/test_qe_app.py
unkcpz: Fair, 60 is too much.

tests/test_qe_app.py
```diff
@@ -11,29 +13,35 @@ def test_notebook_service_available(notebook_service):
...
+    )
+    element.click()
+
     driver.find_element(By.XPATH, "//option[@value='Diamond']").click()
-    time.sleep(10)
+
```
unkcpz, line 36:
```suggestion
    time.sleep(10)
    
```

tests/test_qe_app.py
unkcpz: I see, I add it back.

tests/test_qe_app.py
unkcpz: This was what I try but with no luck, the test failed, and report the element is not loaded. I assume the reason is the `element_to_be_clickable` is for element that change from unclickable to clickable such as the `Confirm` button below. If the `presence_of_element_located` works fine, I think there is no harm to keep it.

tests/test_qe_app.py
unkcpz: I don't know to be honest, I tried `element_to_be_clickable` but it failed and the screenshot is still the loading page.

danielhollas (approved):

Issue #346, opened 2023-02-09
The lockfile of installation processes not updated if the page refresh too early


unkcpz:
This is the issue bring up by @giovannipizzi. 

> The first time I opened (on theospc7), I had two quickly running bars. There was the blue error message for both things (I think the other was that the dos and projwfc codes weren't yet there). Since it wasn't clear if something was happening, I just refreshed. I could find the codes, but then I always get the error below.
If I run that aiida-pseudo command on the command line, it just tells me that it's already installed.

I guess the issue come from the lock files of pseudo libraries installation and quantum-espresso installation are not updated properly when the page refreshed.

mbercx:
Not sure if this is related, but when I tried to reproduce https://github.com/aiidateam/aiida-quantumespresso/issues/887 on `theospc7`, I got stuck in Step 3:

![Screenshot 2023-02-09 at 18 24 46](https://user-images.githubusercontent.com/25607960/217891784-75fcd9b9-0d80-4ace-9e5c-20a57c4adf29.png)

Refreshing the page and trying again only fixed the codes issue:

![Screenshot 2023-02-09 at 18 31 41](https://user-images.githubusercontent.com/25607960/217892117-bbe3f1ed-9d9a-4942-9bfa-ba8984758c79.png)

Opening a terminal and checking the installed pseudos, it does seem to be installed properly, however:

```console
$ aiida-pseudo list
Label                       Type string         Count
--------------------------  ------------------  -------
SSSP/1.2/PBEsol/efficiency  pseudo.family.sssp  85
SSSP/1.2/PBEsol/precision   pseudo.family.sssp  85
SSSP/1.2/PBE/efficiency     pseudo.family.sssp  85
SSSP/1.2/PBE/precision      pseudo.family.sssp  85
```

unkcpz:
Hi @mbercx, I release the latest version `v23.02.1`, please update and I think the issue solved.

mbercx:
Thanks for the quick fix @unkcpz! I'll try it again later.

unkcpz:
The issue appears again after bumping the SSSP version in aiida-pseudo and forgetting to do it here in QeApp. The solution is further hardcoded the pseudo version in pseudo library installation command as implemented by https://github.com/aiidalab/aiidalab-qe/pull/425

PR #347, opened 2023-02-09
Bump version using SSSP v1.2

unkcpz:
fixes https://github.com/aiidalab/aiidalab-qe/issues/346

### DIFF
--- a/aiidalab_qe/parameters/qeapp.yaml
+++ b/aiidalab_qe/parameters/qeapp.yaml
@@ -18,4 +18,4 @@ electronic_type: metal
 ## Calculation settings
 protocol: moderate
 kpoints_distance_override:
-pseudo_family: SSSP/1.1/PBEsol/efficiency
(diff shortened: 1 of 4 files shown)


unkcpz:
@mbercx I merge this and make the release directly since I may not have time working on AiiDAlab until next Wed. Since it is all related to SSSP stuff, I think nothing to review from your side, hope you don't mind :)

unkcpz:
I just find that the change in this PR is not enough if the QeApp is already installed in the AiiDAlab. 
There is no problem if the QeApp is installed from scratch, the aiida-pseudo `1.0.0` will be installed <- `aiida-quantumespresso~=4.1.0` which actually install `4.2.0` and having latest aiida-pseudo installed. 

That means I need to pin the aiida-quantumespresso version to `4.2.0`.

Next time the test need to run for install qeapp in a brand new evironment and installed when old version is there.

Issue #338, opened 2023-01-12
Adding a total charge of the system option


AndresOrtegaGuerrero:
Hello, I think it could be useful for the app to have an option where you can select the total charge of the system.
This could be useful if the user is trying to study a system with a different charge state. The issue of charge can also be important if by using the widget of structure, the modifications lead to a charging system( on purpose).

I was thinking of something like this , where the options are limited from -2, -1, 0 , 1 and 2 charge
![Screenshot 2023-01-12 at 15 53 52](https://user-images.githubusercontent.com/34098967/212100011-e2041e06-a874-4dbe-a1cc-e382a4cab5b7.png)

unkcpz:
@AndresOrtegaGuerrero thanks! How does this parameter influences the QE input? Can you post the pw.in with this setting? 
I think for all such kinds of change, we need first to have workflow supported in aiida-quantumespresso and then we wrapping the workflow in the QeApp.

AndresOrtegaGuerrero:
@unkcpz  I did it like this:

add an extra variable name tot_charge, so it can be also part of the GUI as an input
I can do a PR of my implementation if you want 

```
  @classmethod
    def get_builder_from_protocol(
        cls,
        structure,
        pw_code,
        dos_code=None,
        projwfc_code=None,
        protocol=None,
        overrides=None,
        relax_type=RelaxType.NONE,
        pseudo_family=None,
        tot_charge=0,
        **kwargs,
    ):
```
and then just use this variable to modify the input

`relax.base.pw.parameters["SYSTEM"]["tot_charge"] = tot_charge #add_the_charge`
`bands.scf.pw.parameters["SYSTEM"]["tot_charge"] = tot_charge`
`bands.bands.pw.parameters["SYSTEM"]["tot_charge"] = tot_charge #add_the_charge`
`pdos.scf.pw.parameters["SYSTEM"]["tot_charge"] = tot_charge #add_the_charge`
`pdos.nscf.pw.parameters["SYSTEM"]["tot_charge"] = tot_charge`

unkcpz:
@AndresOrtegaGuerrero thanks a lot for the details!
However, I think we'd better hold on for the moment, this definitely introduces more complicity to the workchain. The `get_builder_from_protocol` function of QeAppWorkChain is already quite unmaintainable. 
[...]

PR #377, opened 2023-03-31
Feature: tot charge

AndresOrtegaGuerrero:
This PR addresses #338 adds the feature to set the total charge of the system in the simulation. 

This implementation avoids the modification of the workchain, and includes the tot_charge parameter in the results

### DIFF
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -42,6 +42,7 @@ def _generate_report_dict(qeapp_wc: WorkChainNode):
     # Material settings
     yield "material_magnetic", builder_parameters["spin_type"]
     yield "electronic_type", builder_parameters["electronic_type"]
+    yield "tot_charge", builder_parameters["tot_charge"]
 
     # Calculation settings
     yield "protocol", builder_parameters["protocol"]
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -42,8 +42,8 @@ def _generate_report_dict(qeapp_wc: WorkChainNode):
     # Material settings
     yield "material_magnetic", builder_parameters["spin_type"]
     yield "electronic_type", builder_parameters["electronic_type"]
-   
-    #Advanced settings
+
+    # Advanced settings
     yield "tot_charge", builder_parameters.get("tot_charge", None)
 
     # Calculation settings
--- a/aiidalab_qe/report.py
+++ b/aiidalab_qe/report.py
@@ -42,7 +42,9 @@ def _generate_report_dict(qeapp_wc: WorkChainNode):
     # Material settings
     yield "material_magnetic", builder_parameters["spin_type"]
     yield "electronic_type", builder_parameters["electronic_type"]
-    yield "tot_charge", builder_parameters["tot_charge"]
+   
+    #Advanced settings
+    yield "tot_charge", builder_parameters.get("tot_charge", None)
 
     # Calculation settings
     yield "protocol", builder_parameters["protocol"]
--- a/aiidalab_qe/steps.py
+++ b/aiidalab_qe/steps.py
@@ -108,14 +108,6 @@ class WorkChainSettings(ipw.VBox):
             style={"description_width": "initial"},
         )
 
-        # tot_charge: total charge of the simulations
-        self.tot_charge = ipw.IntSlider(
-            value=DEFAULT_PARAMETERS["tot_charge"],
-            min=-2,
-            max=2,
-            step=1,
-        )
-
         # Checkbox to see if the band structure should be calculated
         self.bands_run = ipw.Checkbox(
             description="",
@@ -167,17 +159,6 @@ class WorkChainSettings(ipw.VBox):
                         self.spin_type,
                     ]
                 ),
-                ipw.HBox(
-                    children=[
-                        ipw.Label(
-                            "Total charge:",
-                            layout=ipw.Layout(
-                                justify_content="flex-start", width="120px"
-                            ),
-                        ),
-                        self.tot_charge,
-                    ]
-                ),
                 self.properties_title,
                 ipw.HTML("Select which properties to calculate:"),
                 ipw.HBox(children=[ipw.HTML("<b>Band structure</b>"), self.bands_run]),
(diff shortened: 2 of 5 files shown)


superstar54:
Hi @AndresOrtegaGuerrero, Thanks for the implementation. 

In my option, it's better to add this parameter to the advanced setting tab. A DFT calculation for a charging system is more likely a calculation for an advanced user. In the main workflow setting, it should only show the simple "control" parameters (e.g. protocol, materials type).

AndresOrtegaGuerrero:
Hello @superstar54 , thank you. I agree it could be placed in the advanced tab. I put the tot_charge in the main menu, since our users (at EMPA) will be using this feature quite often  since they will be computing point defects (So basically routinely). @unkcpz  what do you think?.

unkcpz:
@AndresOrtegaGuerrero thanks a lot! I and @superstar54 had a discussion on how to integrate new settings into step2. 
I can image that there will be more parameters setting widget comes to QeApp after you and Xing start to add more properties. It is not very clear how such a widget can be independently displayed, not only easy to be found by user but should not entangle with the other widgets (either the base widget from the current QeApp, or the widgets from other properties, such as Xing will also add new widgets for XPS calculations.)

There are multiple options:
1. the widgets can be added to a new tab parallel to "advanced settings", only when the property is selected, the tab show up. This will make the widgets related only to the new property can be fully independent of widgets of base functionality or other widgets. The cons are it is not easy to find by users and that those are in other tabs. 
2. Same as above but the new widgets are showing in the same tab of the basic setting (the "Workflow" table of Step 2), the widgets are in a parent `VBox` widget and will be hidden if the property is not selected. One drawback of this strategy is that the widget is entangled in this table in the `ConfigureQeAppWorkChainStep`. 

As for the `total_charge` in this PR, however it is a vanilla QE parameter which can be set for all PW calculations I think it makes more sense to be added as an override of the base protocol, and the widget is displayed in "Advance setting" tab.

aiidalab_qe/steps.py
```diff
@@ -856,6 +877,23 @@ def update_builder(buildy, resources, npools):
...
+        ]
+        builder.bands.scf.pw.parameters["SYSTEM"]["tot_charge"] = parameters[
+            "tot_charge"
+        ]
+        builder.pdos.scf.pw.parameters["SYSTEM"]["tot_charge"] = parameters[
+            "tot_charge"
+        ]
+        builder.pdos.nscf.pw.parameters["SYSTEM"]["tot_charge"] = parameters[
+            "tot_charge"
+        ]
```
unkcpz, line 895: I find this is error-prone. Can you try to using protocol override to override the base pw protocol and let it propagate to all other sub-workchain? Let me know if this is not clear and we can arrange a meeting to discuss.

aiidalab_qe/report.py
```diff
@@ -42,6 +42,7 @@ def _generate_report_dict(qeapp_wc: WorkChainNode):
     # Material settings
     yield "material_magnetic", builder_parameters["spin_type"]
     yield "electronic_type", builder_parameters["electronic_type"]
+    yield "tot_charge", builder_parameters["tot_charge"]
```
unkcpz, line 45: I think this is a great example to push us to have an interface for adding a new widget that controls an input parameter that will be read from the output report widget. In your PR, there are four files required to change. The idealist goal is only to implement in one place and add somewhere to reference the newly added widget.

unkcpz (changes requested):
@AndresOrtegaGuerrero Thanks a lot for making the ball rolling. My comments are slightly general, I'll work in parallel to have a generic interface for similar parameter control widgets. What you can try in this PR in parallel is my above comment, moving this widget to "Advance settings" tab.

aiidalab_qe/steps.py
AndresOrtegaGuerrero: > @AndresOrtegaGuerrero Thanks a lot for making the ball rolling. My comments are slightly general, I'll work in parallel to have a generic interface for similar parameter control widgets. What you can try in this PR in parallel is my above comment, moving this widget to "Advance settings" tab.

@unkcpz Thank you I will proceed then , One question , should I create then one specific class for this "Advanced Seetings" , since the idea will be to have Tot_charge, but I guess we want in this new tab to have other advanced settings like Periodicity (I submitted the PR in the aiida-quantumespresso) or Hubbard settings?

AndresOrtegaGuerrero: > 

@unkcpz  Maybe I would like if we can have a small talk regarding this one. I am not sure if this requires the modification of the workchain?

aiidalab_qe/report.py
AndresOrtegaGuerrero: @unkcpz Maybe we can discuss about this one, I think this will require modification in the workchain , right?

aiidalab_qe/steps.py
unkcpz: > One question , should I create then one specific class for this "Advanced Seetings" , since the idea will be to have Tot_charge, but I guess we want in this new tab to have other advanced settings like Periodicity.

Do you mean the class of widget for "Advance setting" specifically? I didn't look at the code, but maybe it is a good idea.

AndresOrtegaGuerrero:
I have done the modifications you guys suggested, the option of tot_charge is now in the Advanced Setting tab, 
I implemented a new class in steps `class PwAdvancedSettings(ipw.VBox):` , where we can include other advanced features. I was thinking in the future implementing in this class the starting_magnetization and Hubbard here. 
For total charge i use an extra checkbox since I think in the future there would be more settings and each should have its control.

aiidalab_qe/steps.py
```diff
@@ -781,6 +855,43 @@ def get_input_parameters(self):
...
+                        "base_final_scf": pw_base,
+                    },
+                    "scf": pw_base,
+                    "bands": pw_base,
+                },
+                "pdos": {
+                    "scf": pw_base,
+                    "nscf": pw_base,
+                },
+            }
```
unkcpz, line 891: Thanks! This is exactly what I was thinking. I hope you don't mind if my refactoring PR include similar code. 
My goal is to make this PR become very small for adding a parameter setting widget just needs to edit a reasonable amount of lines.

AndresOrtegaGuerrero (commented):
there are the changes including a tab in AdvancedSettings

unkcpz:
Hi @AndresOrtegaGuerrero, https://github.com/aiidalab/aiidalab-qe/pull/382 is merged, maybe you can rebase this or (maybe much easier to do) open a new PR to supersede this one?

PR #382, opened 2023-04-07
♻️ decouple override settings of builder

unkcpz:
address #383 

- [x] settle all comments with `XXX` notations.
- [x] codecov badge.
- [ ] ~~second test for insulator be more specific, and test only some of the parameters.~~
- [x] having two versions of node viewer in order to show the old workchain results backward compatibly. 
- [x] test pass but have an exception same as https://github.com/aiidalab/aiidalab-widgets-base/pull/471#pullrequestreview-1357821001. See if can work around it or find out why that is happening (memory leak issue?). The test of py-3.10 hang up, passed but not exit.
- [x] The builder_parameters should include information therefore can be passed to `set_parameters` for loading the process to widget.
- [ ] ~~_update_settings method is added for each setting, but these methods are not called. Thus, when loading a previous calculation, widget values are not restored in the GUI.~~ will be addressed in https://github.com/aiidalab/aiidalab-qe/issues/397 or by new plugin design of @superstar54


unkcpz:
Hi @superstar54 @AndresOrtegaGuerrero, I think this PR is starting shaping. Before go into more deep test, I hope you two can have a look if this make everything much clear.

aiidalab_qe/workflows.py
```diff
@@ -155,67 +159,42 @@ def get_builder_from_protocol(
...
     def setup(self):
-        """Perform the initial setup of the work chain."""
+        """Perform the initial setup of the work chain, setup the input structure
+        and the logic to determine which sub workchains to run.
+        """
         self.ctx.current_structure = self.inputs.structure
         self.ctx.current_number_of_bands = None
         self.ctx.scf_parent_folder = None
 
+        # logic based on the properties input
```
AndresOrtegaGuerrero, line 183: Maybe add some logic if properties is None, in that scenario is just a single point ?

unkcpz: Yes, if the properties list is an empty list, there is nothing actually running.
I try to move the logic outside of the workflow. In this workflow, the sub-workflows are run as independently as possible.
Not sure this is a good strategy, worth to discuss.

aiidalab_qe/steps.py
```diff
@@ -806,89 +822,236 @@ def set_pdos_status(self):
...
-        # sub-workchain where the SCF calculation will be run inside.
-        # This potentially increase the complexibility of the logic report widget,
-        # we need to refactoring the QeAppWorkChain and clear the logic here.
-        if RelaxType(parameters["relax_type"]) is RelaxType.NONE and (
-            parameters["run_bands"] or parameters["run_pdos"]
-        ):
-            builder.pop("relax")
+            self.process.base.extras.set("builder_parameters", builder_parameters)
+
+    def _create_builder(self) -> tuple[ProcessBuilderNamespace, dict[str, t.Any]]:
```
superstar54, line 835: This function is too long. There are four steps inside the function,
- step 1, get input parameters
- step2, create the builder
- step3, post-update the builder
- step4, generate report parameters
I suggest creating a function for each step (at least steps 1 and 4)

superstar54 (changes requested):

aiidalab_qe/steps.py
unkcpz: Good suggestion! I also find it quite long and a bit lost in second look, I'll separate in functions.

unkcpz:
@superstar54 thanks! 

> store the parameters to restore all the widgets settings in the four steps. This is important but has not been implemented yet. Please consider this in the PR.

True, I didn't add it yet. Will add it.

aiidalab_qe/workflows/__init__.py
```diff
@@ -99,18 +96,25 @@ def get_builder_from_protocol(
...
         overrides = overrides or {}
         builder = cls.get_builder()
-        builder.structure = structure
 
+        # Set the structure.
+        if isinstance(structure, StructureData):
+            builder.structure = structure
+        else:
+            raise TypeError(
+                f"the `structure` must be of type `StructureData`, get {structure} of type {type(structure)}."
```
mbercx, line 113: Will this really ever happen? I mean we're in control of what inputs are passed to the `get_builder_from_protocol()` method, no?

aiidalab_qe/workflows/__init__.py
```diff
@@ -120,32 +124,32 @@ def get_builder_from_protocol(
             relax_type=relax_type,
             **kwargs,
         )
+
+        # pop the inputs that are exclueded from the expose_inputs
```
mbercx, line 128:
```suggestion
        # pop the inputs that are excluded from the expose_inputs
```

aiidalab_qe/workflows/__init__.py
```diff
@@ -120,32 +124,32 @@ def get_builder_from_protocol(
...
-            bands_overrides.setdefault("scf", {})["pseudo_family"] = pseudo_family
-            bands_overrides.setdefault("bands", {})["pseudo_family"] = pseudo_family
         bands = PwBandsWorkChain.get_builder_from_protocol(
             code=pw_code,
             structure=structure,
             protocol=protocol,
             overrides=bands_overrides,
             **kwargs,
         )
+        # pop the inputs that are exclueded from the expose_inputs
```
mbercx, line 144:
```suggestion
        # pop the inputs that are excluded from the expose_inputs
```

qe.ipynb
```diff
@@ -141,10 +141,10 @@
     "app_with_work_chain_selector = ipw.VBox(children=[work_chain_selector, app])\n",
     "\n",
     "output = ipw.Output()\n",
-    "install_create_github_issue_exception_handler(\n",
-    "    output,\n",
-    "    url='https://github.com/aiidalab/aiidalab-qe/issues/new',\n",
-    "    labels=('bug', 'automated-report'))\n",
+    "# install_create_github_issue_exception_handler(\n",
```
mbercx, line 144: Why is this commented?

aiidalab_qe/app/legacy_report.py
```diff
@@ -0,0 +1,132 @@
+"""Legacy report for the `QeAppWorkChain`, only used for the old workchains run before QeApp version < 23.10
```
mbercx, line 1: So this one is only here for backwards compatibility? Since we use dated versioning, when would we remove this? Do we really _have_ to worry about backwards compatibility at this stage? 😉

aiidalab_qe/app/steps.py
```diff
@@ -36,6 +38,28 @@
...
+    "precise": "SSSP/1.2/PBE/precision",
+}
+
+
+# The static input parameters for the QE App WorkChain
+# The dataclass does not include codes and structure which will be set
+# from widgets separately.
+# Relax type, electronic type, spin type, are str because they are used also
+# for serialized input of extras attributes of the workchain
+@dataclass(frozen=True)
```
mbercx, line 54: Me likey data class ❤️

mbercx (approved):
Quite a bit PR, so I didn't look into every line of code in much details, but it all looks great! Much cleaner, great work @unkcpz!

I still left some comments/questions, but will already approve.

aiidalab_qe/workflows/__init__.py
```diff
@@ -120,32 +124,32 @@ def get_builder_from_protocol(
             relax_type=relax_type,
             **kwargs,
         )
+
+        # pop the inputs that are exclueded from the expose_inputs
         relax.pop("structure", None)
         relax.pop("clean_workdir", None)
-        relax.pop("base_final_scf", None)
+        relax.pop("base_final_scf", None)  # never run a final scf
```
AndresOrtegaGuerrero, line 131: I wonder why to also do this step , doesnt the workchain already removes this ?

aiidalab_qe/app/legacy_report.py
unkcpz: Yes, it is for backward compatibility only. To be honest, I don't know when can we remove this and how can this be removed if there are workflows run with old workchain. My answer probably be this code will stay there forever or we simply don't care about it and having all workchain requires a re-run with new extra attributes attached to the node. 

@superstar54 @AndresOrtegaGuerrero, you have some node run with old workchain, do you think it is okay to not have backward compatibility considered?

aiidalab_qe/workflows/__init__.py
unkcpz: You are correct, removed.

qe.ipynb
unkcpz: For test purposes, I forgot to bring it back. :sweat_smile:

aiidalab_qe/steps.py
```diff
@@ -806,89 +822,236 @@ def set_pdos_status(self):
...
+                ).setdefault("SYSTEM", {})[
+                    "degauss"
+                ] = self.smearing_settings.degauss.value
+
+        parameters_overrides = {
+            "relax": {
+                "base": pw_overrides["base"],
+            },
+            "bands": {
+                "scf": pw_overrides["scf"],
```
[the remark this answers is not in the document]
unkcpz: Yes, bands workchain contains `relax` workchain but pdos does not. I want to keep the bands sub-workchain in qeapp workchain as similar as possible to pdos sub-workchain. 

I think the better design would be even more verbose for workchain that relax sub-workchain will always having a final_scf calculation. The caching is turned on for using this final_scf as the first step of bands/pdos workchain.

aiidalab_qe/app/legacy_report.py
unkcpz: Just have an offline discuss with @superstar54, we think there is no harm to keep the code, I need it since otherwise I need to rerun the workchain which I did for couple of large components for MARVEL review. I think it is also good to keep it as a placeholder for future backward compatibility support of report builder parameters. We can try to keep one legacy version supported in the future and give a more verbose warning to ask to re-run the workchain if we deprecate the support for the output.

AndresOrtegaGuerrero: Maybe add extra comments giving instructions that new features and modifications should not be consider in the legacy and only in node_view  ?? , or how will it work the addition of new features ? if not here maybe in the README ?? (what you do you think @unkcpz  of adding  few design guidelines for future PRs) (or maybe this is not really needed ? )

unkcpz: > Maybe add extra comments giving instructions that new features and modifications should not be consider in the legacy and only in node_view ??

Good point, I add a comment on both legacy_report.py and on the lines where the condition happened. 

> what you do you think @unkcpz of adding few design guidelines for future PRs)

You mean the guidelines for how to write workflow and add new properties? Or the guideline for the PRs?

AndresOrtegaGuerrero: Yes the former

mbercx:
To supplement @AndresOrtegaGuerrero's message: I mainly raised the fact that maintaining backwards compatibility is perhaps an undue burden. I don't think there are that many heavy users yet who would lose any important data with this change. I think our policy until further notice (or until public release, tbd) should be to throw backwards compatibility to the wind so we are not limited in making improvements to the interface, and the code does not become more difficult to maintain.

tests/conftest.py
```diff
@@ -1,121 +1,231 @@
...
 @pytest.fixture(scope="session")
-def screenshot_dir():
-    sdir = Path.joinpath(Path.cwd(), "screenshots")
-    try:
-        os.mkdir(sdir)
-    except FileExistsError:
-        pass
-    return sdir
+def generate_upf_data():
+    """Return a `UpfData` instance for the given element a file for which should exist in `tests/fixtures/pseudos`."""
```
superstar54, line 98: Where is this folder?

superstar54 (changes requested):
Hi @unkcpz, Thanks for implementing this. I added a few special suggestions to the review.
Here are some general comments:
- `_update_settings` method is added for each setting, but these methods are not called. Thus, when loading a previous calculation, widget values are not restored in the GUI.
- In the test, the app tested as a whole in the submit step (kind of integration test). It's nice if you also add tests for each setting (unit test), e.g. workchain, smearing, especially to override the value and then test it.

aiidalab_qe/app/steps.py
```diff
@@ -807,82 +834,252 @@ def set_pdos_status(self):
...
+
+    def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
+        """Get the parameters of the `QeWorkChain` from widgets."""
+        # create the override parameters for sub PwBaseWorkChain
+        pw_overrides = {"base": {}, "scf": {}, "nscf": {}, "band": {}}
+        for key in ["base", "scf", "nscf", "band"]:
+            if self.pseudo_family_selector.override_protocol_pseudo_family.value:
+                pw_overrides[key]["pseudo_family"] = self.pseudo_family_selector.value
+
+        for key in ["base", "scf"]:
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: @superstar54 you mean that we do not to override de` dgauss` or just the `smearing` type ?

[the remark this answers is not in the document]
superstar54: For the `nscf` calculation, `occupations` is set to `tetrahedra` in the pdos workchain. In this case, `smearing` and `degauss` is not used. I think @mbercx knows better about this.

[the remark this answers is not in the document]
AndresOrtegaGuerrero: But sometimes depending of the degauss, that value will be the one use for projwfc, so in case the user download the data it cant decrease the degauss to do a new 'gaussian broadening'. So i guess the degauss is only needed for the `scf` ? @superstar54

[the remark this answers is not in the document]
superstar54: Thanks for pointing out this. In this case,  we need to pass the `degauss` to the `nscf` workchain, because the `dos` and 
the `projwfc` will read the input data from the `nscf` calculation.

It will be useful to add the setting for the `degauss` of dos and projwfc, and we could do this in a future PR.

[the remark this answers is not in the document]
AndresOrtegaGuerrero: just to clarify, if you use a degauss in pw.x (say a small number) , you can later use a higher value for degauss in projwfc.x , and that is fine. What you cant do is to use a high degauss in pw.x  and later in projwfc decrease it. Or at least that is my experience , maybe @mbercx  could give us your opinion?

tests/conftest.py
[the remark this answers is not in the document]
unkcpz: It is useless, at least for test here. I copy the fixture from qe plugin and I think it is also not used there. Open a PR at https://github.com/aiidateam/aiida-quantumespresso/pull/933, let's see if it passes all tests.

tests/conftest.py
```diff
@@ -1,121 +1,231 @@
...
-    # `port_for` takes a container port and returns the corresponding host port
-    port = docker_services.port_for("aiidalab", 8888)
-    url = f"http://{docker_ip}:{port}"
-    token = os.environ["JUPYTER_TOKEN"]
-    docker_services.wait_until_responsive(
-        timeout=30.0, pause=0.1, check=lambda: is_responsive(url)
-    )
-    return url, token
+    with tempfile.TemporaryDirectory() as dirpath:
+        for values in elements.values():
```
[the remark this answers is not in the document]
unkcpz: Since there is no much code for this function and the upf file is actually a fake file of every element, I think it is okay to keep it as it is.

tests/conftest.py
```diff
@@ -1,121 +1,231 @@
...
+        smearing_settings = SmearingSettings()
+        smearing_settings._update_settings(**kwargs)
+        return smearing_settings
+
+    return _smearing_settings_generator
+
+
+@pytest.fixture()
+def kpoints_settings_generator():
+    """Return a function that generates a kpoints settings dictionary."""
```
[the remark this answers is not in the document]
unkcpz: It can I assume but will make the code even not easy to understand. It can be a `settings_generator` with passing `workchain`, `smearing` or `kpoints_settings` with the kargs. Sometimes it is not bad pattern to repeat some code if it makes things clear and easy to implement. I'd argue to keep in this way.

tests/conftest.py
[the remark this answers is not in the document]
unkcpz: The fixture is used in `aiida-quantumespresso` for sure, but not necessary here, so I update the docstring only.

aiidalab_qe/app/steps.py
```diff
@@ -807,82 +834,252 @@ def set_pdos_status(self):
...
+            if self.kpoints_settings.override_protocol_kpoints.value:
+                pw_overrides[key][
+                    "kpoints_distance"
+                ] = self.kpoints_settings.kpoints_distance.value
+
+            if (
+                self.smearing_settings.override_protocol_smearing.value
+                and self.workchain_settings.electronic_type.value == "metal"
```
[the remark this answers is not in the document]
unkcpz: > By the way, the smearing should also work for the insulator, right?

No, if select insulator, the `occupations` is set to `fixed`, the smearing and degauss are not in the input.

AndresOrtegaGuerrero (approved):
Everything looks fine to me!

unkcpz:
Per @superstar54's comment:
> _update_settings method is added for each setting, but these methods are not called. Thus, when loading a previous calculation, widget values are not restored in the GUI.

I tried with using builder_parameters  to recover the widgets, but it is a bit more complicate than I thought. The problem is the key store in the builder_parameters  extra attributes are not same for the key used in setting widgets. There are a further step to convert the paramters from setting widget to builder_parameters, to bring it back I think we need a more systematical way to do it and I prefer to do it in a separate PR. If you agree, I’ll merge it.

Moreover, when I look at the refactored architecture in https://github.com/aiidalab/aiidalab-qe/issues/383, setting widgets directly from builder_parameters  is not ideal since it bring the entanglment between viewer and controller which is we want to avoid from first place. Instead, we need introduce and use the dataclass  for all the settings object and having bilateral converters between setting wdigets and builder_parameters  and having this dataclass in the middle.

@superstar54 is working on introducing the new "plugin-like"way of adding new properties calculation, and it will be handled in the new design by the get and set methods of the panel.

superstar54 (approved):

Issue #386, opened 2023-04-25
Adapt installation process of pseudo families


mbercx:
AFAIK, currently the QE app installs the pseudopotential files in the background when the container starts. If the user comes to the submission part of the app before the installation is finished, the submission is blocked until the pseudos are installed.

One point is that we can work on already having the pseudos installed in the image, and/or have the pseudo archives in the image and install without having to download the files once https://github.com/aiidateam/aiida-pseudo/pull/135 is merged.

Perhaps we don't have to install _all_ configurations right away though. Maybe we can install the most commonly used ones by default, and if the user selects a more esoteric one, install the family from a local archive or download it upon submission.

From a discussion with @giovannipizzi and @unkcpz.

unkcpz:
>  install the family from a local archive

The pseudo families are just the aiida groups so I think this is a very good idea to go with, this even not requires the `aiida-pseudo`, we can have the archives exported from groups for every pseudo library and import on requests.

danielhollas:
Just noting as a non-QeApp user, I would be against significantly increasing the size of the aiidalab Docker image because of a single App. :-)

mbercx:
> Just noting as a non-QeApp user, I would be against significantly increasing the size of the aiidalab Docker image because of a single App. :-) 

I think we can make it part of a QE-specific part of the stack? ^^ it's mainly vital that we find ways to improve the startup time for demo purposes.

unkcpz:
> Just noting as a non-QeApp user, I would be against significantly increasing the size of the aiidalab Docker image because of a single App. :-)

Don't worry :smile: We'll keep it as a separate image, let's say `full-demo`? Then add QeApp from the start and then everything that will highlight its power, maybe include also your app @danielhollas ?

danielhollas:
Sure, wasn't sure which image you have in mind. The problem with my app is that it depends on ORCA, which needs to be installed manually since it requires a license (so not great for automatic deployment unfortunately).

unkcpz:
We now have pseudo-dojo and having a specific docker image for qeapp with pseudopotential libraries are downloaded to the container and installed when the app first-time start. 
The image size is sure a bit fat. 
-  download size `full-stack`: 794.06 MB and `qe`: 1.28 GB
- container size (after pull to docker) `full-stack`: 2.48GB and `qe`: 3.82 GB

For the online deployment, the image will download once and used for all running container, thus the download size is not such a huge problem. 
[...]

PR #387, opened 2023-04-26
add PseudoDojo

superstar54:
Here, two pseudo families added:
- PseudoDojo/0.4/PBE/SR/standard/upf
- PseudoDojo/0.4/PBEsol/SR/standard/upf

I didn't touch the installation process, thus, it still uses the `aiida-pseudo install ` method like that for `SSSP`.

## GUI
The GUI of pseudo selection is now:

Screenshot from 2023-04-26 16-22-18


AndresOrtegaGuerrero:
I think it would be worth to also include FR and  also to add the stringent accuracy

superstar54:
Hi @AndresOrtegaGuerrero, Thanks for your suggestions. It's good to add more pseudo-family, but this leads to a longer time for the installation, thus the user needs to wait in the first run. The idea here is only to install the most commonly used pseudo, and trigger additional downloads (FR, stringent) if an advanced calculation is needed. This may be added in a further PR. See discussion https://github.com/aiidalab/aiidalab-qe/issues/386.

AndresOrtegaGuerrero:
@superstar54  I agree, though it might be a good idea that the user gets some description that the PseudoDojo that they are using is SR, and that the accuracy is the standard

superstar54:
> @superstar54 I agree, though it might be a good idea that the user gets some description that the PseudoDojo that they are using is SR, and that the accuracy is the standard

Thanks!. A description for Pseudodojo has been added.

unkcpz:
Hi @superstar54, I change the base to branch `main` since I decided to deprecate `master` after the change in app registry is merged, I hope you don't mind. 

Is there any block for this PR? Do we need more discussion on this? I remember we want to double-check the versioning of pseudo-dojo?

AndresOrtegaGuerrero:
I think in a previous meeting, we discussed as well that it could be useful to have normal and stringent precision

superstar54:
> it could be useful to have normal and stringent precision

Done

AndresOrtegaGuerrero:
@superstar54  Since the code was re factored, you should do a  rebased on this branch with main , since now there is a folder named app for the files you modified

aiidalab_qe/pseudos.py
```diff
@@ -77,11 +78,16 @@ def __init__(self, **kwargs):
...
+        pseudo_family_type = DEFAULT_PARAMETERS["pseudo_family"].split("/")[0]
+        if pseudo_family_type.upper() == "SSSP":
+            pseudo_family_type += (
+                " " + DEFAULT_PARAMETERS["pseudo_family"].split("/")[-1]
+            )
         self.protocol_selection = ipw.ToggleButtons(
-            options=["efficiency", "precision"],
-            value=DEFAULT_PARAMETERS["pseudo_family"].split("/")[3],
-            layout=ipw.Layout(max_width="80%"),
+            options=["SSSP efficiency", "SSSP precision", "PseudoDojo"],
```
AndresOrtegaGuerrero, line 88: The ToogleButton should contained the option to differenciate with PseudoDojo you are using , stringent or standard

superstar54:
@AndresOrtegaGuerrero Thanks for the review. I addressed the issues and rebased the branch with the main branch.

AndresOrtegaGuerrero:
@superstar54  Thank you it looks nice, I tested and didnt observe any issues, I only have one opinion an is regarding the widget. 
Screenshot 2023-06-08 at 09 36 51
I was wondering if we should use a Dropdown ? or maybe to make like a 2x2 grid to organize the buttons? What do you think?

superstar54:
> I was wondering if we should use a Dropdown ? or maybe to make like a 2x2 grid to organize the buttons? What do you think?

I updated the layout so that they are shown in a 2x2 grid on a large screen and 4x1 on a small one.

AndresOrtegaGuerrero:
@superstar54  It looks great to me !  thank you

AndresOrtegaGuerrero (approved):

aiidalab_qe/app/report.py
```diff
@@ -39,14 +40,20 @@ def _generate_report_dict(builder_parameters: dict):
...
-    functional = builder_parameters["functional"]
+    pseudo_family = builder_parameters["pseudo_family"]
+    yield "pseudo_family", pseudo_family
+    pseudo_family_list = pseudo_family.split("/")
+    pseudo_library = pseudo_family_list[0]
     yield "pseudo_library", pseudo_library
+    yield "pseudo_version", pseudo_family_list[1]
+    functional = pseudo_family_list[2]
     yield "functional", functional
+    if pseudo_library == "SSSP":
```
unkcpz, line 51: We shouldn't put the logit in `report.py`, very error-prone.

aiidalab_qe/app/steps.py
```diff
@@ -1114,14 +1114,9 @@ def _create_extra_report_parameters(self) -> dict[str, t.Any]:
...
             pseudo_family = PROTOCOL_PSEUDO_MAP[qe_workchain_parameters.protocol]
 
-        pseudo_family_info = pseudo_family.split("/")
         extra_report_parameters.update(
             {
                 "pseudo_family": pseudo_family,
-                "pseudo_library": pseudo_family_info[0],
-                "pseudo_version": pseudo_family_info[1],
-                "functional": pseudo_family_info[2],
-                "pseudo_protocol": pseudo_family_info[3],
```
unkcpz, line 1124: TBH, removing this and putting the logic to "viewer" the `report.py` is going back to where before I did the refactoring.

aiidalab_qe/app/sssp.py
```diff
@@ -42,8 +53,36 @@ def run_(*args, **kwargs):
     mult = 1 / len(pseudo_set)
     for i, pseudo in enumerate(pseudo_set):
         yield mult * i
-        p_family, p_version, p_func, p_type = pseudo.split("/")
-        run_(["aiida-pseudo", "install", p_family.lower(), "-x", p_func, "-p", p_type])
+        if pseudo.startswith("SSSP"):
+            p_family, p_version, p_func, p_type = pseudo.split("/")
```
unkcpz, line 57: Instead of if..else and calling`run_` in two place, please just construct the command options list and call `run_` once.

unkcpz (changes requested):
Thanks @superstar54, some change requests.

superstar54:
@unkcpz thanks for the review, I have addressed all the requests.

aiidalab_qe/app/pseudos.py
```diff
@@ -118,11 +131,18 @@ def __init__(self, **kwargs):
...
+                pseudo_family = f"PseudoDojo/0.4/{self.dft_functional.value}/SR/{self.protocol_selection.value.split()[1]}/upf"
+            elif "SSSP" in self.protocol_selection.value:
+                pseudo_family = f"SSSP/1.2/{self.dft_functional.value}/{self.protocol_selection.value.split()[1]}"
+            else:
+                raise ValueError(
+                    f"Unknown pseudo family {self.override_protocol_pseudo_family.value}"
+                )
+            self.value = pseudo_family
+        else:
+            self.value = DEFAULT_PARAMETERS["pseudo_family"]
```
unkcpz, line 145: I am a bit worried about the complex if..else nest here. Can you check it again if you can make it more clear?

AndresOrtegaGuerrero: @superstar54  perhaps you can put the option in a dictionary and evaluate the key ?

unkcpz: @AndresOrtegaGuerrero, I don't quite understand what you mean. @superstar54, can you try to address it? 
Then we can merge this PR.

AndresOrtegaGuerrero: @superstar54 you can ignore my suggestion and merge if you want

superstar54: Thanks for the suggestions. I updated the code to make the string pretty and hopefully more clear. A test is also added for the `PseudoFamilySelector` widget.

PR #394, opened 2023-05-08
bump AWB to released 2.0

unkcpz:
Jinja2 and importlib are from aiida-quantumespresso, therefore the version is loosely pinned. The pybtex has to be added to work around the installation issue.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,9 +23,11 @@ project_urls =
 packages = find:
 install_requires =
     aiida-core~=2.2,<3
+    Jinja2~=3.0
     aiida-quantumespresso~=4.2
     aiidalab-widgets-base~=2.0
     filelock~=3.8
+    importlib-resources~=5.2
     widget-bandsplot~=0.5.1
     pybtex==0.24.0
     pymatgen==2022.9.21
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,13 +22,12 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    Jinja2~=3.0
     aiida-core~=2.2,<3
     aiida-quantumespresso~=4.2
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -22,13 +22,12 @@ project_urls =
...
 install_requires =
-    Jinja2~=3.0
     aiida-core~=2.2,<3
     aiida-quantumespresso~=4.2
-    aiidalab-widgets-base==2.0.0b6
+    aiidalab-widgets-base~=2.0
     filelock~=3.8
-    importlib-resources~=5.2.2
     widget-bandsplot~=0.5.1
+    pybtex==0.24.0
```
unkcpz, line 30: This somehow has to be pinned, because of the issue https://github.com/aiidalab/aiidalab-qe/actions/runs/4917462707/jobs/8782576404. It also fails the test of aiidalab-docker-stack in https://github.com/aiidalab/aiidalab-docker-stack/pull/374

setup.cfg
```diff
@@ -22,13 +22,12 @@ project_urls =
 [options]
 packages = find:
 install_requires =
-    Jinja2~=3.0
```
superstar54, line 25: Please keep jinjia2 and importlib-resources here. Otherwise, the dependency is not clear.

superstar54 (changes requested):

setup.cfg
unkcpz: Yes, make sense. I delete it for test. You are correct, since it is used in the package, better to keep in the dependencies list.

unkcpz: But I loose the version a bit so it more flexible to the change from upstream packages, e.g aiida-core and aiida-quantumespresso.

superstar54 (approved):

Issue #388, opened 2023-05-04
WorkChainSelector: Remove background thread


danielhollas:
In https://github.com/aiidalab/issues/issues/13 I discovered that running a background thread may lead to memory leak when the Jupyter frontend becomes unavailable. In QeApp there is a background thread in WorkChainSelector that periodically updates the list of processes from AiiDA DB. 

I've tested that the leak is indeed present by setting the refresh interval to 0.1s. 

![image](https://user-images.githubusercontent.com/9539441/236155634-50ec1249-e460-4844-bc8f-9ddb6aad52f7.png)

In practice the leak will not be as severe since the refresh interval is much larger, but I have nevertheless decided to remove the background thread in my app since I did not see much benefit anyway. You might want to do the same in QeApp. Please see https://github.com/ispg-group/aiidalab-ispg/pull/177 for the code changes and additional justification.

PR #401, opened 2023-05-18
seprate workchain selector to be extendable

unkcpz:
fixes https://github.com/aiidalab/aiidalab-qe/issues/388
This PR is also an attempt of https://github.com/aiidalab/aiidalab-widgets-base/issues/382.


aiidalab_qe/app/process.py
```diff
@@ -1,32 +1,31 @@
...
     # widget to its default stage (no work chain selected), because we cannot
     # use `None` as setting the widget's value to None will lead to "no selection".
     _NO_PROCESS = object()
 
-    FMT_WORKCHAIN = "{wc.pk:6}{wc.ctime:>10}\t{wc.state:<16}\t{wc.formula} \t {wc.relax_info} \t {wc.properties_info}"
+    FMT_WORKCHAIN = "{wc.pk:6}{wc.ctime:>10}\t{wc.state:<16}"
 
-    def __init__(self, **kwargs):
+    BASE_FIELDS = [("pk", int), ("ctime", str), ("state", str)]
+    EXTRA_FIELDS = None
```
danielhollas, line 25: Since this class attribute is meant to be overriden, it probably shouldn't be in all caps? Same with `FMT_WORKCHAIN`.

aiidalab_qe/app/process.py
```diff
@@ -1,32 +1,31 @@
...
     # Note: We use this class as a singleton to reset the work chains selector
     # widget to its default stage (no work chain selected), because we cannot
     # use `None` as setting the widget's value to None will lead to "no selection".
     _NO_PROCESS = object()
 
-    FMT_WORKCHAIN = "{wc.pk:6}{wc.ctime:>10}\t{wc.state:<16}\t{wc.formula} \t {wc.relax_info} \t {wc.properties_info}"
+    FMT_WORKCHAIN = "{wc.pk:6}{wc.ctime:>10}\t{wc.state:<16}"
 
-    def __init__(self, **kwargs):
+    BASE_FIELDS = [("pk", int), ("ctime", str), ("state", str)]
```
danielhollas, line 24: Perhaps to make this more readable and safer, you could define a namedtuple tor the elements of this list. Maybe something like

```
WorkChainField= namedtuple("WorkChainField", "field type")

BASE_FIELDS = [WorkChainField("pk", int), WorkChainField("ctime", str), WorkChainField("state", str)] 
```

But perhaps that's too verbose, I dunno...

aiidalab_qe/app/process.py
```diff
@@ -182,3 +126,42 @@ def _observe_value(self, change):
...
         self.work_chains_selector.value = new
+
+
+class QeAppWorkChainSelector(WorkChainSelector):
+    EXTRA_FIELDS = [("formula", str), ("relax_info", str), ("properties_info", str)]
+
+    def __init__(self, **kwargs):
+        super().__init__(process_label="QeAppWorkChain", **kwargs)
+
+    def parse_extra_info(self, uuid):
```
danielhollas, line 137: In the parent class, you're passing `pk`, not `uuid`.

qe.ipynb
danielhollas, line 1: If you look closely at the corresponding PR in my app, I've added one extra hack, where I trigger the workchain selector refresh when a submitted workflow finishes. That way the process status in the selector is updated from "Running" to "Finished". This is a bit ugly since I need to directly patch the ProcessMonitor object in one of the Step classes, the but it works. 

https://github.com/ispg-group/aiidalab-ispg/pull/177/files#diff-5c45c68ed738fa714abe5583a6dabb14053e4cc0c93d8b253292d1cca0464a6aR162

danielhollas (changes requested):
@unkcpz this looks great, thanks! Let's try it out here, and then later we can upstream to AWB.

Except for some minor comments / suggestions that you can feel free to ignore, there is one subtle thing about triggering a refresh upon submitted WorkChain completion.

danielhollas:
@unkcpz since there is already an associated issue, I've removed this PR from the AiiDAlab project to avoid duplication. Hope that's okay.

unkcpz:
Sure, thanks! If you mean it is duplicated with https://github.com/aiidalab/aiidalab-qe/issues/388 which is already in the current sprint. I was adding this to the project in order to set it to the current sprint so I didn't forget to bring it up in the meeting (I admit this is not the original way we plan to use the sprint, but for myself it a bit easy to track everything happened. I start to have a very bad memory among a couple of things 😿 ).

aiidalab_qe/app/process.py
unkcpz: Yes, my carelessness.

qe.ipynb
unkcpz: Interesting, this is cool. I also mad `process_monitor` the interface of `ViewQeAppWorkCHainStatusAndStep` so it is designed to be used as such, then it is not very hacked.

aiidalab_qe/app/process.py
unkcpz: It is a bit redundant to me, in design the `WorkChainSelector` will be used as the parent class, and having the extra_fields override, it means the `WorkchainField` also need to be imported, not very convenient in my opinion.

aiidalab_qe/app/process.py
unkcpz: Yep, changed and add a docstring for how to subclass and extend it.`

danielhollas (approved):
Just a two short suggestions, otherwise I am approving....

qe.ipynb
danielhollas: Fair enough. The `on_sealed` callbacks are called when the process finishes. Normally one should passed them via the class constructor, so what I am doing is definitely hacky, since I am touching the internal implementation of ProcessMonitor.

aiidalab_qe/app/process.py
```diff
@@ -182,3 +143,46 @@ def _observe_value(self, change):
             self.refresh_work_chains()
 
         self.work_chains_selector.value = new
+
+
+class QeAppWorkChainSelector(WorkChainSelector):
+    EXTRA_FIELDS = [("formula", str), ("relax_info", str), ("properties_info", str)]
```
danielhollas, line 149:
```suggestion
    extra_fields = [("formula", str), ("relax_info", str), ("properties_info", str)]
```

danielhollas (changes requested):
@unkcpz have you tested your latest changes? Found a bug while I was transplanting this into my app.

aiidalab_qe/app/process.py
```diff
@@ -57,121 +68,71 @@ def __init__(self, **kwargs):
...
-        state: str
-        formula: str
-        relax_info: str
-        properties_info: str
-
-    @classmethod
-    def find_work_chains(cls):
+        self.refresh_work_chains()
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
danielhollas: Indeed, I just tested that the `@abc.abstractmethod` decorator does not work in this case (i.e. I can instantatiate a subclass that does not implement this method). For it to work, you'd need to inherit from the ABC class.

https://docs.python.org/3/library/abc.html#abc.abstractmethod

aiidalab_qe/app/process.py
unkcpz: oops! I just did a quick test by launching a calculation, and didn't find this. BIG FAILURE 😿

aiidalab_qe/app/process.py
```diff
@@ -1,32 +1,42 @@
...
-    busy = traitlets.Bool(read_only=True)
+    busy = tl.Bool(read_only=True)
 
     # Note: We use this class as a singleton to reset the work chains selector
     # widget to its default stage (no work chain selected), because we cannot
     # use `None` as setting the widget's value to None will lead to "no selection".
     _NO_PROCESS = object()
 
-    FMT_WORKCHAIN = "{wc.pk:6}{wc.ctime:>10}\t{wc.state:<16}\t{wc.formula} \t {wc.relax_info} \t {wc.properties_info}"
+    FMT_WORKCHAIN = "{wc.pk:6}{wc.ctime:>10}\t{wc.state:<16}"
```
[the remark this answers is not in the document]
unkcpz: Good idea, changed.

aiidalab_qe/app/process.py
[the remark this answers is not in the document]
unkcpz: Yeah, I try to use it as a placeholder. The parent class here is not an `ABCMeta`.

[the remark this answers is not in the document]
unkcpz: But seems this is a misuse 🙅

danielhollas (approved):
Awesome. I've transplanted this to my app and everything is much cleaner now, thanks!

PR #402, opened 2023-05-23
Feature/adv tot charge

AndresOrtegaGuerrero:
This PR updates #377 to be compatible with the new re factoring


AndresOrtegaGuerrero:
@mbercx  I was wondering if maybe you could give me a hint of what could be the sources in the failure of our test with python 3.9 and 3.8 ??

mbercx:
@AndresOrtegaGuerrero thanks for the ping! Yeah our nightly builds also started failing, not sure what's going on, but will prioritise fixing it! Very weird that even the released version `4.3.0` is suddenly failing...

mbercx:
Probably related to https://github.com/explosion/spaCy/issues/12659

mbercx:
Was most likely due to a bug in `pydantic~=1.10`, now fixed in `1.10.8`. I've opened a PR to adapt the dependency:

https://github.com/aiidateam/aiida-quantumespresso/pull/941

I reran some tests that were failing for an `aiida-quantumespresso` PR, which now pass. But here the `pydantic` version seems somehow restricted, since the CI is installing version `1.10.2`. Not sure why, but let's see what happens after my PR is merged and release a hotfix version `3.4.1`.

AndresOrtegaGuerrero:
> Was most likely due to a bug in `pydantic~=1.10`, now fixed in `1.10.8`. I've opened a PR to adapt the dependency:
> 
> [aiidateam/aiida-quantumespresso#941](https://github.com/aiidateam/aiida-quantumespresso/pull/941)
> 
> I reran some tests that were failing for an `aiida-quantumespresso` PR, which now pass. But here the `pydantic` version seems somehow restricted, since the CI is installing version `1.10.2`. Not sure why, but let's see what happens after my PR is merged and release a hotfix version `3.4.1`.

Thank you !

aiidalab_qe/app/steps.py
```diff
@@ -1077,6 +1166,11 @@ def _extract_report_parameters(
         ] = builder.bands.bands_kpoints_distance.value
         parameters["nscf_kpoints_distance"] = builder.pdos.nscf.kpoints_distance.value
 
+        if builder.relax.base["pw"]["parameters"]["SYSTEM"]["tot_charge"]:
```
unkcpz, line 1169:
```suggestion
        tot_charge = builder.relax.base["pw"]["parameters"]["SYSTEM"]["tot_charge"]
        if total_charge:
```
This better move to above under the comment where said why the value getting from `builder.relax.base` is feasible.  

I also don't quite understand, the `tot_charge` can be a `None`?

aiidalab_qe/app/steps.py
```diff
@@ -856,6 +939,12 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
         for key in ["base", "scf", "nscf", "band"]:
             if self.pseudo_family_selector.override_protocol_pseudo_family.value:
                 pw_overrides[key]["pseudo_family"] = self.pseudo_family_selector.value
+            if self.pw_advanced_settings.override_pw_advanced_settings.value:
+                pw_overrides[key]["pw"] = {"parameters": {"SYSTEM": {}}}
+                if self.pw_advanced_settings.override_tot_charge.value:
+                    pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
+                        "tot_charge"
+                    ] = self.pw_advanced_settings.tot_charge.value
```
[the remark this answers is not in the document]
unkcpz: Can we always have this value in the workflow input, the default value is `0.0`. When pass from widget to workchain, it assure it is not `None`. This will make the workflow (model) part more clear.

aiidalab_qe/app/steps.py
```diff
@@ -210,6 +210,85 @@ def _update_settings(self, **kwargs):
                 getattr(self, key).value = kwargs[key]
 
 
+class PwAdvancedSettings(ipw.VBox):
+    title = ipw.HTML(
+        """<div style="padding-top: 0px; padding-bottom: 10px">
+        <h4>Pw Advanced Settings</h4></div>"""
```
unkcpz, line 216: "Pw" is not easy to be understood by people who don't very familiar with QE. I was wondering, can we have the widget called `AdvancedSettings` and having all tricky parameters setting included parameters of dos.x and projwfc.x if we have it in the future.

aiidalab_qe/app/static/workflow_summary.jinja
```diff
@@ -92,6 +92,10 @@
         <td>{{ nscf_kpoints_distance }} &#8491;<sup>-1</sup></td>
       </tr>
       {% endif %}
+      {% if tot_charge %}
```
unkcpz, line 95: I don't think this logic is needed, if tot_charge not set, it is `0` by default, we can show it as well.

aiidalab_qe/app/steps.py
```diff
@@ -210,6 +210,85 @@ def _update_settings(self, **kwargs):
...
+            indent=False,
+            value=False,
+        )
+        self.override_tot_charge = ipw.Checkbox(
+            description="",
+            indent=False,
+            value=False,
+        )
+
+        self.tot_charge = ipw.IntSlider(
```
unkcpz, line 234: The total change can be a real number accourding to https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm289. I also prefer using the text input widget which is conformed with smearing width and k-distance widgets.

unkcpz (changes requested):
@AndresOrtegaGuerrero thanks a lot! 

I am wondering how we categorize parameters as "advanced". One option is anything beyond protocol is advanced. Or the smearing and kpoints overrides are categorized as advance setting as well. 

And what confuses me most is that we already have the `Advance Setting` tab but it has another advance setting. I think it is true we need a parent widget to collect all such "advance setting" widgets in one. Why not we put all the widgets in "Advance Settings" tab to a widget called `AdvanceSettings` and manage the display there solely?

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: Yeah it can be done, it they do not apply advanced settings

aiidalab_qe/app/static/workflow_summary.jinja
AndresOrtegaGuerrero: Is needed, since if the user doesnt activate the tot_charge , that value should be None. Thats why the widget has a general override, so the user can determine if advanced setting will be use, and then tot_charge ,( or any future option) should have also an override (to determine if you want to use it)

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: Ok, I will do this change. (i need to check , if there should be some logic in case the user user letters instead of number)

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: Yes, I will do this change as well

aiidalab_qe/app/steps.py
```diff
@@ -856,6 +939,12 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
         for key in ["base", "scf", "nscf", "band"]:
             if self.pseudo_family_selector.override_protocol_pseudo_family.value:
                 pw_overrides[key]["pseudo_family"] = self.pseudo_family_selector.value
+            if self.pw_advanced_settings.override_pw_advanced_settings.value:
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: Yes, for sure I will modify this name

aiidalab_qe/app/steps.py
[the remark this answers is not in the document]
AndresOrtegaGuerrero: I would prefer that is applied in all the steps, in any case i think is save. regarding the `None` that is why I use an override for the tot_charge, so it only checks or assigns if the user wanted to defined a tot_charge, otherwise why to assign it as `0.0`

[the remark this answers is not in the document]
AndresOrtegaGuerrero: Regarding having all widgets, would you prefer to be now included in `AdvancedSettings` ?

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero thanks! Look at the failed test on py3.10, the change on total charge is expected but I am worried about the kpoints_distance also changing. Can you check if the widget worked as expected that the kpoints can be correctly overrided?
> 
> To get the new regression tast data you just need to run `pytest tests/ --force-regen` (https://pytest-regressions.readthedocs.io/en/latest/overview.html#force-regen).

@unkcpz  I just updated the conftest

AndresOrtegaGuerrero:
@unkcpz  , there was a change in the [557a460](https://github.com/aiidalab/aiidalab-qe/pull/402/commits/557a460901fe99dd42f7d694a16397dacdf0ea69) I need to another commit

AndresOrtegaGuerrero:
@unkcpz  Now it should be fine

aiidalab_qe/app/static/workflow_summary.jinja
unkcpz: > if the user doesnt activate the tot_charge , that value should be None.

If tot_charge widget not activate, can it be set to `0.0`, then we don't need this `if tot_charge`. We can always have tot_charge in pw input file.

unkcpz (changes requested):
@AndresOrtegaGuerrero Thanks again! Still, some minor requests, but the overall shape is very satisfactory.

aiidalab_qe/app/static/workflow_summary.jinja
AndresOrtegaGuerrero: @unkcpz  The issue (i feel, not sure about it) is about compatibility (if we care at this point) , if calculations do not have the tot_charge in the parameters, an error will always be present to visualize the results. What do you think ?

AndresOrtegaGuerrero: I removed the logic

aiidalab_qe/app/steps.py
```diff
@@ -856,30 +970,30 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
...
+                        self.advanced_settings.smearing.override.value
+                        and self.workchain_settings.electronic_type.value == "metal"
+                    ):
+                        # smearing type setting
+                        pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
+                            "smearing"
+                        ] = self.advanced_settings.smearing.smearing.value
+
+                        # smearing degauss setting
+                        pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
```
unkcpz, line 994: Why you remove the `setdefault` clause? Python dictionary method `setdefault()` is similar to get(), but will set dict[key]=default if key is not already in dict.

aiidalab_qe/app/steps.py
```diff
@@ -1077,6 +1191,12 @@ def _extract_report_parameters(
         ] = builder.bands.bands_kpoints_distance.value
         parameters["nscf_kpoints_distance"] = builder.pdos.nscf.kpoints_distance.value
 
+        tot_charge = builder.relax.base["pw"]["parameters"]["SYSTEM"]["tot_charge"]
```
unkcpz, line 1194: It will fail if SYSTEM namelist don't have `tot_charge`, correct?

aiidalab_qe/app/steps.py
```diff
@@ -1077,6 +1191,12 @@ def _extract_report_parameters(
...
+        if tot_charge:
+            parameters["tot_charge"] = tot_charge
+        else:
+            parameters["tot_charge"] = 0.0
+
```
unkcpz, line 1199:
```suggestion
        parameters["tot_charge"] = builder.relax.base["pw"]["parameters"]["SYSTEM"].get("tot_charge", 0.0)
```

And probably better having a `_DEFAULT_TOT_CHARGE` for the class.

unkcpz (changes requested):
Almost there, three tiny requests.

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: @unkcpz  Because if `self.advanced_settings.override.value` can control if you are going to use overrides , and the next line already creates the dictionary `pw_overrides[key]["pw"] = {"parameters": {"SYSTEM": {}}}`.

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: Yes, I will modify this

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: @unkcpz I  solve it but i didnt implement the _DEFAULT_TOT_CHARGE for the class, since that is a function

aiidalab_qe/app/steps.py
```diff
@@ -210,6 +210,102 @@ def _update_settings(self, **kwargs):
...
 
+class AdvancedSettings(ipw.VBox):
+    title = ipw.HTML(
+        """<div style="padding-top: 0px; padding-bottom: 10px">
+        <h4>Advanced Settings</h4></div>"""
+    )
+    description = ipw.HTML("""Select the advanced settings for the <b>pw.x</b> code.""")
+    # set here defaul values for the advanced settings
+    tot_charge_default = 0
+
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: if i use the traitlets.Instance , it complaints is not defined , However it is not needed since the traitlets is now on `AdvancedSettings,` like it was for SmearingSettings and KpointsSettings in `ConfigureQeAppWorkChainStep` and `SubmitQeAppWorkChainStep`. So there shouldn't be an issue here

aiidalab_qe/app/report.py
```diff
@@ -65,6 +65,8 @@ def _generate_report_dict(builder_parameters: dict):
         yield "degauss", builder_parameters["degauss"]
         yield "smearing", builder_parameters["smearing"]
 
+    yield "tot_charge", builder_parameters.get("tot_charge", None)
```
unkcpz, line 68:
```suggestion
    yield "tot_charge", builder_parameters["tot_charge"]
```
We can do this now.

aiidalab_qe/app/steps.py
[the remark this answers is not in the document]
unkcpz: Okay, no problem from my side. Thanks!

aiidalab_qe/app/steps.py
unkcpz: I see, this is more clear, nice!

tests/conftest.py
```diff
@@ -183,10 +196,12 @@ def submit_step_widget_generator(
     workchain_settings_generator,
     smearing_settings_generator,
     kpoints_settings_generator,
+    tot_charge_generator,
+    # workchain_advanced_settings_generator,
```
unkcpz, line 200:
```suggestion
```

unkcpz (approved):
Two nitpicks, then it is fine to go. I approved it.

AndresOrtegaGuerrero:
@unkcpz  I accepted your changes, thank you for all the input and help!

unkcpz:
@AndresOrtegaGuerrero thanks a lot for such nice work! Looking forward to reviewing more for new features.

Issue #238, opened 2022-05-18
Add a widget to assign the spin guesses.


yakutovicha:
**Is your feature request related to a problem? Please describe.**

One of the most powerful features of QE is the ability to deal with spins. The QE app should provide an interface to enter the initial spin guesses.

**Describe the solution you'd like**
Add an "Assign spins" structure editor.

AndresOrtegaGuerrero:
I think there should be the option to rather use the starting_magnetization option, but also starting_ns_eigenvalue , so if you have a transition metal , you can select how many electrons you want to have in alpha and beta channels.

PR #409, opened 2023-06-01
Feature: Starting Magnetization

AndresOrtegaGuerrero:
This PR implements a class  `MagnetizationSettings` for Advanced Settings in the QE App, to be able to set initial_magnetic_moments in the calculations. Address #238 
The user can select the magnetization per kind defined in the `input_structure` 
Screenshot 2023-06-01 at 13 06 46

The will be another PR for an AddingTags Editor, so the user can create new tags for the atoms in the structure.


mbercx:
Hi @AndresOrtegaGuerrero and @unkcpz! I will be very busy the next few weeks on other projects, so unless there is something specific needed from my end (i.e. on the `aiida-quantumespresso` plugin), I will not have time to review PRs on the QEapp.

In general, I've been less and less involved on the development there. I am simply spread too thin otherwise and then no work gets done anywhere. Apologies.

unkcpz:
A quick comment, if I didn't select structure, there is a dangling "override" checkbox there which looks a bit weird. I'd suggest having a text under the "Total charge" showing it is the placeholder for setting the initial magnetic moments.

image

unkcpz (changes requested):
@AndresOrtegaGuerrero thanks for adding this! I only have some minor requests, the overall implementation looks fine, cheers!

aiidalab_qe/app/steps.py
```diff
@@ -318,6 +323,79 @@ def reset(self):
...
+                        self.kinds_widget_out,
+                    ],
+                ),
+            ],
+            layout=ipw.Layout(justify_content="space-between"),
+            **kwargs,
+        )
+        self.override.observe(self._disable_kings_widgets, "value")
+
+    def _disable_kings_widgets(self, _=None):
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: hahahahah!

aiidalab_qe/app/static/workflow_summary.jinja
```diff
@@ -96,6 +96,10 @@
         <td>Total Charge</td>
         <td>{{ tot_charge }}</td>
       </tr>
+      <tr>
+        <td>Init. Magnetic Moments</td>
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: I will update this

aiidalab_qe/app/steps.py
```diff
@@ -975,6 +1061,14 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
                     pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
                         "tot_charge"
                     ] = self.advanced_settings.tot_charge.charge.value
+                if (
+                    self.advanced_settings.magnetization.override.value
+                    and self.workchain_settings.spin_type.value == "collinear"
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: To define if you are using spin is 
```
        self.spin_type = ipw.ToggleButtons(
            options=[("Off", "none"), ("On", "collinear")],
            value=DEFAULT_PARAMETERS["spin_type"],
        )
```

aiidalab_qe/app/steps.py
```diff
@@ -1034,6 +1128,7 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
             spin_type=spin_type,
             electronic_type=electronic_type,
             overrides=overrides,
+            initial_magnetic_moments=starting_magnetization,
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: i will update this

AndresOrtegaGuerrero:
@unkcpz  Thank you for your suggestions and help. I just added another test for advanced settings and adapted your suggestions. Let me know what you think.

aiidalab_qe/app/steps.py
```diff
@@ -318,6 +325,107 @@ def reset(self):
             self.override.value = False
 
 
+class MagnetizationSettings(ipw.VBox):
+    """Widget to set the initial_magnetic_moments from each Kind in StructureData
```
unkcpz, line 329:
```suggestion
    """Widget to set initial magnetic moments for each `Kind` (usually the element name, but you can use `Fe1`, `Fe2` to distinguish iron atoms of different sites) of `StructureData`
```

Am I rephrasing it correctly? Please feel free to polish what I wrote.

aiidalab_qe/app/steps.py
```diff
@@ -318,6 +325,107 @@ def reset(self):
...
+        if hasattr(self.kinds, "children") and self.kinds.children:
+            for i in range(1, len(self.kinds.children)):
+                self.kinds.children[i].children[0].value = 0.0
+
+    def create_kinds_widget(self):
+        if self.input_structure_labels:
+            widgets_list = []
+            for label in self.input_structure_labels:
+                hbox = ipw.HBox()
+                float_widget = ipw.BoundedFloatText(
```
unkcpz, line 377: No a big issue since it as a local variable, but I prefer `bounded_float` or simply `w`. `float_widget` seems a bit like a widget that can float on the webpage. Feel free to ignore this comment.

aiidalab_qe/app/steps.py
```diff
@@ -318,6 +325,107 @@ def reset(self):
...
+                    description=label,
+                    min=-1,
+                    max=1,
+                    step=0.1,
+                    value=0.0,
+                    disabled=True,
+                )
+                hbox.children = [float_widget]
+                widgets_list.append(hbox)
+            kinds_widget = ipw.VBox([ipw.HTML("Define magnetization")] + widgets_list)
```
unkcpz, line 387: Here you didn't use HTML tags for formatting, is `ipw.HTML` really needed?

aiidalab_qe/app/steps.py
unkcpz: And why not do like this?

```python
hbox = ipw.HBox(
    children=[
        ipw.BoundedFloatText(...
    ]
)
```

aiidalab_qe/app/steps.py
```diff
@@ -318,6 +325,107 @@ def reset(self):
...
+
+    def _disable_kinds_widgets(self, _=None):
+        for i in range(1, len(self.kinds.children)):
+            self.kinds.children[i].children[0].disabled = not self.override.value
+
+    def reset(self):
+        self.override.value = False
+        if hasattr(self.kinds, "children") and self.kinds.children:
+            for i in range(1, len(self.kinds.children)):
+                self.kinds.children[i].children[0].value = 0.0
```
unkcpz, line 370: I understand why it has to set the value like this, from range 1 to skip the HTML widget of index 0, and second children for the `HBox`.  But it is a bit complex and error-prone when looking at this loop independently. Not sure there is a better way, I think having a Kind widget might be too much. 
Just think about it, if there is not better way to make it clear, leave it as this and add a comment for why start loop from index 1 and what the second children refer to.

aiidalab_qe/app/steps.py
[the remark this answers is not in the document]
unkcpz: I see, fair. 👍

tests/test_submit_qe_workchain.py
```diff
@@ -77,6 +84,59 @@ def test_create_builder_insulator(
...
+    assert report_html.count("None") == 1
+    if "initial_magnetic_moments" not in builder_parameters:
+        parsed = BeautifulSoup(report_html, "html.parser")
+        assert parsed.find("initial_magnetic_moments").text == "None"
+
+
+def test_create_builder_advanced_settings(
+    submit_step_widget_generator,
+):
+    """ "Test the creation of the workchain builder.
```
unkcpz, line 97:
```suggestion
    """Test the creation of the workchain builder.
```

tests/test_submit_qe_workchain.py
```diff
@@ -77,6 +84,59 @@ def test_create_builder_insulator(
...
     report_html = _generate_report_html(builder_parameters)
 
+    # There should be only one None from initial_magnetic_moments
+    assert report_html.count("None") == 1
+    if "initial_magnetic_moments" not in builder_parameters:
+        parsed = BeautifulSoup(report_html, "html.parser")
+        assert parsed.find("initial_magnetic_moments").text == "None"
+
+
+def test_create_builder_advanced_settings(
```
unkcpz, line 94: Thanks a lot for this test, very useful!

unkcpz (changes requested):
Thanks @AndresOrtegaGuerrero, three minor requests. It is all fine then.

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: Thank you i will expand this 👍

tests/test_submit_qe_workchain.py
AndresOrtegaGuerrero: Yes i think here we can add the test for new advanced settings :)

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: I will update in such a way that is a VBox instead and to access the valees is `self.kinds.children[i].value`

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: I will modify this

AndresOrtegaGuerrero:
@unkcpz  Let me know i adapted the code :)

aiidalab_qe/app/steps.py
```diff
@@ -318,6 +325,109 @@ def reset(self):
...
+        return kinds_widget
+
+    def update_kinds_widget(self):
+        self.input_structure_labels = self.input_structure.get_kind_names()
+        self.kinds = self.create_kinds_widget()
+        self.description.value = "Define magnetization: "
+        self.display_kinds()
+
+    def display_kinds(self):
+        if "PYTEST_CURRENT_TEST" not in os.environ and self.kinds:
```
unkcpz, line 407: Interesting, I didn't know this `PYTEST_CURRENT_TEST`. Can you double check this is really needed here? I think even is a test it should be found to clear_output and display. Am I missing something?

unkcpz (approved):
Hi @AndresOrtegaGuerrero, It is all fine. I have one question about the `PYTEST_CURRENT_TEST`, but I approve it anyway since the logic is all great for me.

aiidalab_qe/app/steps.py
AndresOrtegaGuerrero: If i dont use this , it will return the widget during the pytest, so to avoid this behaviour i use the display and clear only if i am not in a pytest , but i do the update before.

unkcpz: Sorry, I didn't understand, what will happen if the widget is returned during pytest, the test will fail? with what exception?

AndresOrtegaGuerrero: @unkcpz  The test wont fail , but during the test in the terminal it will print the widget as text

unkcpz: Did you use `pytest -s`?  Other wise it should not print things. If this is not needed, I'd not add it. 
Sorry for the back-and-forth on this PR.

AndresOrtegaGuerrero: @unkcpz  I did the test without with `pytest -s` without this and the issue occurs, the test passed but it returns the object in this case a VBox

unkcpz: > I did the test with `pytest -s`

Yep, `-s` is to print things to the stdout. I think it is fair to suppress it with your way, otherwise run with `-s` locally would be a mess. 

All good to me now. Do you mind merge this one with writing a description of the commit message? Just click the "squash and merge" below and input the message in the dialog.

PR #410, opened 2023-06-01
Feature/ Tags editor to define Kinds in StructureData

AndresOrtegaGuerrero:
This PR implements an Editor for the QE App, to enable the user to define extra kinds within a structure. For example if you have a system compose with Fe atoms, you can define Fe1, and Fe2 based on a selection. This editor is intended to facilitated spin polarisation calculations with #409 

Screenshot 2023-06-01 at 13 24 14


unkcpz:
@superstar54, can you give this one a review? Thanks!

aiidalab_qe/app/widgets.py
```diff
@@ -515,3 +523,107 @@ def _observe_value(self, change):
...
+
+    structure = traitlets.Instance(ase.Atoms, allow_none=True)
+    selection = traitlets.List(traitlets.Int, allow_none=True)
+    input_selection = traitlets.List(traitlets.Int, allow_none=True)
+
+    def __init__(self, title=""):
+        self.title = title
+        self._status_message = StatusHTML()
+        self.atom_selection = ipw.Text(
+            description="Define kind", value="", layout={"width": "initial"}
```
superstar54, line 539:
```suggestion
            description="Index of atoms", value="", layout={"width": "initial"}
```

aiidalab_qe/app/widgets.py
```diff
@@ -515,3 +523,107 @@ def _observe_value(self, change):
...
+            self._status_message.message = """
+            <div class="alert alert-info">
+            <strong>Please select atoms first.</strong>
+            </div>
+            """
+        else:
+            selection = string_range_to_list(self.atom_selection.value)
+            new_structure = deepcopy(self.structure)
+            new_tags = new_structure.get_tags()
+            new_tags[selection] = 0
```
superstar54, line 614: When clearing tags, you set the tag value to 0. It is more like `reset` , so maybe change `clear` to `reset`.

superstar54 (changes requested):
@AndresOrtegaGuerrero Thanks for adding this. It would be very useful for advanced calculation.
Could you add a section to show all tags of all atoms? In this case, the user can make sure the tag is updated successfully.

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero Thanks for adding this. It would be very useful for advanced calculation. Could you add a section to show all tags of all atoms? In this case, the user can make sure the tag is updated successfully.

@superstar54 How would you suggest to do the visualisation (Is something i have been wondering) ? Because the system can have a good amount of atoms.

superstar54:
> Because the system can have a good amount of atoms.

One option would be only to show the tags of the selected atoms.

AndresOrtegaGuerrero:
@superstar54 I just adapted the changes you suggested , let me know.

superstar54:
Thanks!  I see the tags of selected atoms. Not it shows in one line, e.g.:

`Id: 5 O4 ; Id: 1 Si ; Id: 2 Si2 ; Id: 3 O2`

Better to make a table like this:
| Index | element| tag|
|--|--|--|
|5|O|4|
|1|Si||
|2|Si|2|
|3|O|2|

superstar54 (approved):
Perfect! Thanks!

PR #416, opened 2023-06-19
Restrict pydantic to fix testing

unkcpz:
The dependency is pinned by `aiida-quantumespresso==4.4.0`, which not release yet. Revert after the aiida-quantumespresso bump.


superstar54 (approved):
LGTM!

PR #423, opened 2023-06-21
Backward compatibility for report widget

unkcpz:
I insist to bring this back, since now we are planning to announce to QE community version `23.4.3` which I hard code the version used in Quantum Mobile and in the documentation. 
This means there are potential users who will run calculations with this version and still want to check this workflow after they update the app to `23.10.0`.

### DIFF
--- a/aiidalab_qe/app/report_legacy.py
+++ b/aiidalab_qe/app/report_legacy.py
@@ -1,9 +1,15 @@
 from aiida.orm import WorkChainNode
 from aiida.plugins import WorkflowFactory
 
-from .report import PROTOCOL_PSEUDO_MAP, PSEUDO_LINK_MAP, FUNCTIONAL_LINK_MAP, _generate_report_html
+from .report import (
+    FUNCTIONAL_LINK_MAP,
+    PROTOCOL_PSEUDO_MAP,
+    PSEUDO_LINK_MAP,
+    _generate_report_html,
+)
 
 PwBaseWorkChain = WorkflowFactory("quantumespresso.pw.base")
+
 
 def _generate_report_dict(qeapp_wc: WorkChainNode):
     builder_parameters = qeapp_wc.base.extras.get("builder_parameters", {})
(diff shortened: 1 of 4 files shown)


AndresOrtegaGuerrero:
@unkcpz Thank you! I think this could also help to keep the app stable for future features. I was wondering if we should have a test for the legacy part ?

unkcpz:
> if we should have a test for the legacy part ?

I was thinking about it when I prepare this PR, but what lacks here is the legacy workchain node. The good side is the old code will not be changed so if we test by hand then it should be fine.

AndresOrtegaGuerrero:
> 

Ok , I see your point. Well LGTM !

aiidalab_qe/app/report_legacy.py
```diff
@@ -0,0 +1,126 @@
+from aiida.orm import WorkChainNode
+from aiida.plugins import WorkflowFactory
+
+from .report import (
+    FUNCTIONAL_LINK_MAP,
+    PROTOCOL_PSEUDO_MAP,
```
AndresOrtegaGuerrero, line 6: Should it be a problem with the version of the SSSP from aiida-pseudo?

unkcpz: I think you are right. 🤔 It is a problem. The problem comes from the pseudo library version information is not written in the QE workchain node but deduct from protocol. 
If so, we have to ignore the legacy node view that may give the wrong pseudo information, or we provide a legacy pseudo mapping from protocol to the old version (I'd say it is now all SSSP 1.2, and with https://github.com/aiidalab/aiidalab-qe/pull/425, the pseudo library version is hardcoded and will not be updated automatically). 
Then we need a change that also writes pseudo library information into the QE workchain so it can be directly used in the report widget.

unkcpz:
The current state for the release is we make the announcement to QE community with `v23.10.0` in October. This means will have the new interface and do not need to consider this backward compatibility. So I close this.

PR #425, opened 2023-06-22
add version when installing pseudo

superstar54:
QeApp failed when installing SSSP pseudo, because the new SSSP version (1.3) is released, and we hard-coded version 1.2 in QeApp, so we need to add the version when installing pseudo.


unkcpz (approved):
LGTM! thanks!

PR #434, opened 2023-08-03
Doc: using the new propared qe image

unkcpz:


unkcpz:
Okay, although I move the paragraph out intently. But that's for sure to be discussed. Let's use your organization of the sections and content, just keep the new content added. We can discuss how to re-organize during the documentation day.

PR #435, opened 2023-08-15
Support using customize pseudopotential

unkcpz:
The use case comes from I need to test Li pseudopotential I generated and plot the bands/pdos. 

- [x] unit test
- [ ] override or not?
- [x] the kind can be "O1", but the cutoffs are not properly set from pseudo family. Check by SiO2.
- [x] Unify the element/kind of `PseudoUploadWidget`


unkcpz:
@AndresOrtegaGuerrero this new widget borrow some idea from your magnetization setting widget. Let's see if we can converge them into a generic widget for setting things that dynamically changed with structure and depend on structure kinds.

AndresOrtegaGuerrero:
@unkcpz I really like the idea, specially because I was doing a test with PseudoDojo using the cutoff from the app, and it seems the cutoff I obtained were not well converged. So this input would be nice in case the user want to do a convergence test with the app

AndresOrtegaGuerrero:
@unkcpz I think you should add like a caption that explains the user they can upload and/or modify each Pseudo

aiidalab_qe/app/steps.py
```diff
@@ -1087,8 +1097,19 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
         for key in ["base", "scf", "nscf", "band"]:
             if self.pseudo_family_selector.override_protocol_pseudo_family.value:
                 pw_overrides[key]["pseudo_family"] = self.pseudo_family_selector.value
+
+            if self.pseudo_setter.pseudos:
+                pw_overrides[key].setdefault("pw", {"parameters": {"SYSTEM": {}}})
+                pw_overrides[key]["pw"]["pseudos"] = self.pseudo_setter.pseudos
+                pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
+                    "ecutwfc"
```
AndresOrtegaGuerrero, line 1105: The selection of ecutwfc and ecutrho, is just to take the max values per independent pseudos. I would suggest that you add two extra widgets that it display the ecutwfc and ecutwfc, used in the simulation. Since they can vary depending if it is NC or US. I think also it gives the possibility to the user to test or modify this in case the suggestion by aiida-pseudo might not work for a particular system. And also it could open the door for future implementation of SOC , where we need to modify the cutoffs

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero this new widget borrow some idea from your magnetization setting widget. Let's see if we can converge them into a generic widget for setting things that dynamically changed with structure and depend on structure kinds.

Yes maybe we can make a widget like DynamicalSettings

aiidalab_qe/app/steps.py
unkcpz: Done! See new commits. I also change the color of the text of the message to show for the input structure is not set, for your widget as well.

unkcpz:
Hi @AndresOrtegaGuerrero, can you give this a second look? It is already from my side. For making it generic, I'll open an issue for future implementation, I notice @mikibonacci also mention it in the comment of plugin UI pr. It is certainly worth doing.

AndresOrtegaGuerrero:
@unkcpz  thank you!, I just check it looks good for me. My opinion will be that you should  include two widgets to define the wfc and rho cutoff (in case the ones defined is not enough). I am saying out of experience, i tried to run something with the app using PseudoDojo, to later realized that the cutoffs were really off and not converged. Then to be forced to stop using the App since i didn't have control over it. I think if you are already modifying the pseudos you should include this.

unkcpz:
> My opinion will be that you should include two widgets to define the wfc and rho cutoff (in case the ones defined is not enough)

Ah! I see your point. Actually, you can just set the ecutwfc/ecutrho higher for one pseudo and it will be used. But you are right, this looks like a non-standard trick. I'll try to implement that.

unkcpz:
Emm.... Right after I start, I feel it does not fully make sense to have one more widget to set it. An extra cutoff widget are a but redundant as the one for setting pseudos cutoffs separately. 
It looks not bad for changing the cutoffs for pseudos from pseudo families. But in my case, I upload and use my own pseudo, and the ecutwfc/ecutrho are just showing 0 and it surely encourages the user to set it. 
Can we keep it as this, and I'll remember to bring this layout design to Nicola and ask for feedback? (there are more such design decisions, I'll make an issue for this.)

AndresOrtegaGuerrero:
@unkcpz  I would suggest that for the design, there should be only control over ecutwfc and ecutrho (of the calculation).  The user can upload a different functional. I dont think there is need to display or control the cutoff of each element. In particular if the user is going to combine US , PAW or NC pseudos. The user rather have control over the ecutwfc and ecutrho of the calculation.

AndresOrtegaGuerrero:
> > I dont think there is need to display or control the cutoff of each element.
> 
> I agree. Then it'll be after the pseudo, I display the cutoffs, if it is customized pseudo, I show "No recommended ecutwfc". The FloatText input widget is provided for the user to change. I'll do the change.
> 
> > In particular if the user is going to combine US , PAW or NC pseudos.
> 
> Why you think this is a special case? In my SSSP verification, US/PAW is used, I use dual=8 and for NC use dual=4 and run convergence. Do you mean sometimes higher cutoffs are required?

What did you mean by dual? You meant the ration between ecutwfc and ecutrho, like in SSSP is ecutrho = (8 to 12) * ecutrho and in NC is ecutrho = 4* ecutwfc ?

unkcpz:
> You meant the ration between ecutwfc and ecutrho, like in SSSP is ecutrho = (8 to 12) * ecutrho and in NC is ecutrho = 4* ecutwfc ?

Yep. But anyway, I think we have a good plan for how to improve it.

AndresOrtegaGuerrero:
@unkcpz Yes, you can have like only two widget to control ecutwfc and ecutrho. once you select the pseudopotential the value of these two widgets get set by the values given by aiida-pseudo. Then the user can decide to increase them accordingly. In a future release, once the plugin PR is merge, we can even make one tab to run a convergence workflow when the user can select a list of ecutwfc to test.

unkcpz:
Hi @AndresOrtegaGuerrero, I did the change, please have another look.

AndresOrtegaGuerrero:
@unkcpz  Thank you! , It looks great to me. I did a few test didn't find any issue.  One suggestion I would add is that if would be good that the cutoff wfn and rho, should be included in the override of pseudos

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -420,8 +421,20 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
...
+            if self.pseudo_setter.pseudos:
+                pw_overrides[key].setdefault("pseudos", self.pseudo_setter.pseudos)
+
+                pw_overrides[key].setdefault("pw", {"parameters": {"SYSTEM": {}}})
+                pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
+                    "ecutwfc"
+                ] = self.pseudo_setter.ecutwfc
+                pw_overrides[key]["pw"]["parameters"]["SYSTEM"][
+                    "ecutrho"
+                ] = self.pseudo_setter.ecutrho
```
unkcpz, line 434: See here how ecutwfc/ecutrho in pw_overrides affect the bulider.

unkcpz:
> One suggestion I would add is that if would be good that the cutoffs wfn and rho, should be included in the override of pseudos

Not quite sure what you mean. It is passed to the builder through the override and it will change the summary report when it is changed.

src/aiidalab_qe/app/submission/__init__.py
AndresOrtegaGuerrero: Yes, that i tested. What i meant was in the GUI, that the widget is not available if the checkbox is activated

unkcpz: I see~ I don't want to make this change for now. 1. It may not be proper to control the cutoffs setup with pseudo override. 2. we need more discussion on what needs to be overridden, how to display the override checkbox.
If we follow the other parameters that exist or are added by you, there in principle should be a checkbox for every parameter. But maybe that is not necessary. @superstar54 was suggesting since the user on this page would anyway want/know how to change things, maybe we don't need checkboxes.

unkcpz: Let's discuss this on the issue https://github.com/aiidalab/aiidalab-qe/issues/444

AndresOrtegaGuerrero (approved):
Look good to me

PR #438, opened 2023-08-16
DOC: add xps to howto

superstar54:
This PR adds the doc for XPS calculation using the `plugin_workchain` branch of aiidalab-qe.

I also improved the XPS plugin by:
- Added an upload button for the experimental data.
- Add the offset correction into the workflow so that the user don't need to correct the binding energy manually.


docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
+============================
+How to calculate XPS spectra
+============================
+
+Overview
+========
+This tutorial will guide you through the process of setting up and running a XPS calculation for CH\ :sub:`3`\CHO molecule.
```
PNOGillespie, line 7:
```suggestion
This tutorial will guide you through the process of setting up and running an XPS calculation for CH\ :sub:`3`\CHO molecule.
```

docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
...
+
+Steps
+=====
+
+To start, go ahead and :doc:`launch </installation/launch>` the app, then follow the steps below.
+
+
+Step 1 Select a structure
+--------------------------------
+For this tutorial task, please use the `From Exmple` tab, and select the CH\ :sub:`3`\CHO molecule structure.
```
PNOGillespie, line 18:
```suggestion
For this tutorial task, please use the `From Example` tab, and select the CH\ :sub:`3`\CHO molecule structure.
```

docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
...
+
+
+Then go to the **Advanced settings** tab, navigate to `Accuracy and precision`, tick the `Override` box on the right-hand-side and in the dropdown box under `Exchange-correlation functional` select `PBE`.
+
+.. image:: ../_static/images/XAS_Plugin-Set_Adv_Options-Alt-Annotated-Cropped.png
+   :align: center
+
+
+.. note::
+    At present, the core-hole pseudopotential for Si and O only available in PBE functional.
```
PNOGillespie, line 44:
```suggestion
    At present, core-hole pseudopotentials for Si and O are only available for the PBE functional.
```

docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
...
+
+Then go to the **XPS setting** tab, and the set the parameters which are highlighted as follows:
+
+- In the **Select core-level** section, select ``C_1s``.
```
PNOGillespie, line 48:
```suggestion
- In the **Select core-level** section, select ``C_1s`` by ticking the appropriate box.
```

docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
...
+
+
+Then, lick the **Confirm** button.
+
+
+Step 3 Choose computational resources
+---------------------------------------
+
+For this small system, we can use the local computer (``localhost``) to run the calculation. The code ``pw-7.2@localhost`` is already selected by default.
```
PNOGillespie, line 57: This is just to clarify for the user that this *should* be the case (and that there is an issue somewhere if it isn't). I know that QE fails to install on my own machine, for instance.
```suggestion
For this small system, we can use the local computer (``localhost``) to run the calculation. The code ``pw-7.2@localhost`` should already be selected by default.
```

docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
...
+
+.. figure:: /_static/images/xps_step_4.png
+   :align: center
+
+.. tip::
+
+   When comparing the calculated binding energies to the experimental data, please correct the energies by the given offset listed in the ``Offset Energies`` table.
+   Besides, the DFT calculated binding energies do not include spin-orbit splitting of the core level state.
+   We can include the spin-orbit splitting using its experimental value.
+   Take `f` orbit as a example, we need subtracting :math:`3/7` of the experimental spin-orbit splitting or adding :math:`4/7` of the DFT calculated value, to get the position of the :math:`4f_{7/2}` and :math:`4f_{5/2}` peaks, respectively.
```
PNOGillespie, line 93: This section here could be useful, but (unless I'm mistaken) there is no way to extract the exact binding energy values from AiiDALab-QE.

If there is a way to do this, then please update some part of the text with instructions on how to do so, otherwise this doesn't provide much utility and should be removed for simplicity.

docs/source/howto/xps.rst
```diff
@@ -0,0 +1,119 @@
...
+
+.. figure:: /_static/images/xps_etfa_dft.png
+   :align: center
+
+One can compared the calculated chemical shift with the experimental data.
+
+.. figure:: /_static/images/xps_etfa_exp.jpg
+   :align: center
+
+
```
PNOGillespie, line 112: This last example doesn't seem to provide much for the user's benefit, for a few reasons:
1. The calculated and experimental spectra aren't plotted together for comparison.
2. Without a set of instructions, or source for the structure, the user can't use this like some sort of complex "real-world" example as an extension of the documentation and so it doesn't really add much.

I would suggest either removing the section or replacing the example with one for Phenacetylene, which we have tested already and can provide figures/data for.

PNOGillespie (changes requested):
Hi @superstar54, good work putting together a simple documentation set.

I've added some suggested some minor changes to fix typos and grammar in some places.

The more substantive changes are:
1. Step 2 is missing a picture of the XPS Settings tab.
2. There are parts towards the end which talk about the `offset energies` and the spin-orbit splitting, but lacks a prior step which explains how to export the data (beyond reading the value from the Plotly widget)
3. The real world example of ETFA doesn't seem to give much useful information for the user about the plugin (see the relevant comment for more information).

docs/source/howto/xps.rst
superstar54: Thanks for pointing out this. I add the detail text to tell user how to find the exact binding energies.

superstar54:
@PNOGillespie thanks for the review. I revised the page based on your suggestions. 

> Step 2 is missing a picture of the XPS Settings tab.

I added a picture.

> There are parts towards the end which talk about the offset energies and the spin-orbit splitting, but lacks a prior step which explains how to export the data (beyond reading the value from the Plotly widget)

I added a detailed instruction.

> The real world example of ETFA doesn't seem to give much useful information for the user about the plugin (see the relevant comment for more information).

I replied after your comment.

Please have a second review.

PNOGillespie (changes requested):
Hi @superstar54. Looks good for the most part, particularly in providing a contrast between ETFA and Phenylacetylene. Just a few proofreading changes to make, then I think this is good to go.

Since you have added a spectrum upload tool (which seems to work fine, from what i can see) I would suggest (in a different PR) one improvement, which would be to add a UI element to scale the intensities of the computed XPS traces so as to make it easier to compare the profile of (in particular) the total XPS to the experimental data.

Again, I would just suggest you open an `improvements` issue for that and address that at a later time.

superstar54:
Thanks @PNOGillespie . I created an issue as you suggested. #627

PNOGillespie (approved):
Good to go. Nice work @superstar54! Thanks also for adding the issue page.

PR #428, opened 2023-06-28
Plugin UI

superstar54:
The first PR for #426. This PR implements all the GUI parts for the plugin design.

## Summary of the changes of GUI

- [x] Create `Panel` classes, which are used to register new pluggable tabs (panels) in each step (e.g., step 2: configure, step 4: result).
    - `Panel` class
    - `OutlinePanel` class
    - `ResultPanel` class
- [x] Add a `plugin` folder, add `relax`, `bands`, `pdos` subfolders. Move all codes related to the plugins to their folder.
    - Each folder could have the following files: 
        - setting, use the `Panel` class
        - workchain
        - result, use the `ResultPanel` class
- [x] Refactor the `steps.py` file:
    - [x] Create a `configure` folder, and move all codes related to step 2 (configure workflow) to this folder.
        -  `workchain_settings` is split into `workflow settings` and `basic settings` (spin, electronic type ...).
        - `pseudo_family_selector` and `advanced_settings` are merged into `advance_settings`
    - [x] Create a `submit` folder, and move all codes related to step 3 (submit, resources) to this folder.
    - [x] Create a `result` folder, and move all codes related to step 4 (status and results) to this folder.
        - all the code related to reports in submit step are moved here.
- [x] Create a QeApp class. The QeApp is the `parent` of all four sub-steps

## Change of QeAppWorkChain

- [x] Move the code related to creating the builder of sub-workchain to the plugins folder.
- [x] Attach the output nodes of sub-workchain directly as outputs of the QeAppWorkChain.

Note the sub-workchain is hard coded and not pluggable. I will make another PR to allow the pluggable sub-workchaion.

## Tests
Due to the above changes, the QeApp has become more modularized. Thus, it is straightforward to write the test for each module. The test coverage is > **85%** now!
- [x] services (postgres and rabbitmq) is added in the ci test, because they are now used in generating workchain.
- [x] unit test, add test for each module
- [x] end-to-end test, test the QeApp as a whole through the `QeApp` class

## For the reviewer
- after changing the branch to `plugin_ui`, you need to run `pip install -e .` to set the entry points.
- I usually added comments if I deleted or moved the original code.
- There is a `setting` for the `bands`. It is only used to illustrate that the setting panel appears when the user selects the `bands` property.

The detailed changes:
- for each step, the class has a parent attribute
- link basic protocol to all plugin-specific protocols
- set GUI widget values when loading a process from workcain selector

PR #439, opened 2023-08-17
Adapt modules and files layout for plugin implementation

unkcpz:
This PR is separated and prepared for https://github.com/aiidalab/aiidalab-qe/pull/428 to avoid mixing changes, especially for moving modules/files **without changing any implementations**. See https://github.com/aiidalab/aiidalab-qe/pull/428#issuecomment-1676301972 for the design of restructuring the modules.

Every commit serves its own purpose, so you can basically review it commit by commit.

From #428, there are still more things that should be separated:
- some unit tests which not related to the new "plugin" design can be implemented first to test widgets.
- ipynb can be simplified and moved to a module for test purposes as did in #428 

But better to use new PRs as well. Do you want to do it yourself @superstar54  (maybe better since it was all your effort and it might not fair to "plagiarise" it by me) or I can do it and put you as co-author.


superstar54 (approved):
@unkcpz Thanks for the work. LGTM!

One thing we may consider is to remove the app folder and change it to:
```console
~/apps/aiidalab-qe/src/aiidalab_qe$ ls
common         __init__.py  parameters  static     submission  workflows
configuration  __main__.py  result      structure  version.py
```
so that the `configuration`, `submission`, and `common` are at the same level as the `workflow`. @AndresOrtegaGuerrero what do you think?

For the test and `ipynb`, I think I have time to do it next week.

unkcpz:
No really, see the discussion we had at https://github.com/aiidalab/aiidalab-qe/pull/379#issuecomment-1523390921. Put workflow folder separately will make the workflow import faster without importing anything from app module.

unkcpz:
@superstar54 there are still two things that can be done separately before start introduce the "Panel" concept in:

- some unit tests which not related to the new "plugin" design can be implemented first to test widgets.
- ipynb can be simplified and moved to a module for test purposes as did in https://github.com/aiidalab/aiidalab-qe/pull/428

Can I also do it? This will make me to have a close line-by-line investigation on how you implement tests on every widgets and what are not covered.

superstar54:
> No really, see the discussion we had at [#379 (comment)](https://github.com/aiidalab/aiidalab-qe/pull/379#issuecomment-1523390921). Put workflow folder separately will make the workflow import faster without importing anything from app module.

This can be avoided. Just let the `aiidalab_qe/__init__.py` empty.

unkcpz:
> This can be avoided. Just let the aiidalab_qe/__init__.py empty.

Okay, I see your point. I'll benchmark and change if import workflow still fast.

unkcpz:
Move the modules out and clean the `__init__.py`, all good. Here is the load time check:

```
# workflows module separated from `app`
(base) jovyan@51e7296639d5:~$ time python -c "from aiidalab_qe.workflows import QeAppWorkChain"

real    0m1.897s
user    0m2.443s
sys     0m3.717s

# workflow and app in the same hierarchy, and entries in `__init__.py`
(base) jovyan@51e7296639d5:~$ time python -c "from aiidalab_qe.workflows import QeAppWorkChain"

real    0m3.201s
user    0m3.679s
sys     0m3.833s

#  workflow and app in the same hierarchy,, and no entries in `__init__.py`
(base) jovyan@51e7296639d5:~$ time python -c "from aiidalab_qe.workflows import QeAppWorkChain"

real    0m1.682s
user    0m2.176s
sys     0m3.515s
```

unkcpz:
However, I still don't think this is good. This means we cannot/should not put any module to `aiidalab_qe/__init__.py`, put any module from the current `app` will slow down the import of `workflows` module, which is what we want.

superstar54:
> This means we cannot/should not put any module to aiidalab_qe/__init__.py,

In the current `app` and `workflow` structure, you also should not put anything in the `aiidalab_qe/__init__.py`, right? You put all the app-related modules into the `aiidalab_qe/app/__init__.py`, and then import them like this:
```python
from aiidalab_qe.app import QeApp
```
you can not import a module from `aiidalab_qe` directly, right? In this case, the `app` folder is really not necessary.
e.g,  one need to always add the `app` when importing
```from aiidalab_qe.app.configuration import Configuration```

I think it's better to do it like this:
```from aidalab_qe.configuration import Configuration```

> @superstar54 if this is not what your plugin implementation needed, I propose we just keep it as it is now?

In the workchain PR, to avoid recursive import, one should not put QeApp class in the `app/__init__.py`, which means I need to create a file `app/app.py`, which is not good. So I suggest to remove the app folder and creating an `app.py` file directly.

unkcpz:
> you can not import a module from aiidalab_qe directly, right?

You miss the goal here, we separate the workflows folder to avoid importing `QeAppWorkChain` is slow. As the command, I used above for benchmarking: `time python -c "from aiidalab_qe.workflows import QeAppWorkChain"`

> e.g, one need to always add the app when importing
> `from aiidalab_qe.app.configuration import Configuration`
> I think it's better to do it like this:
> `from aidalab_qe.configuration import Configuration`

Neither of them are good, ultimately it is more clear to have `from aiidalab_qe.app import Configuration`, so I want to put things in `aiidalab_qe/app/__init__.py`.

> In the workchain PR, to avoid recursive import, one should not put QeApp class in the app/__init__.py, which means I need to create a file app/app.py

The `QeApp` class is the class that is not dependent on other class/modules, if there are recursive import, you need to check your implementation to avoid that, no? 
I suggest keeping this structure and I start also moving to `QeApp`, maybe I'll see the problem then.

AndresOrtegaGuerrero:
@unkcpz thank you ! , It looks good to me. I  am wondering what have you decided about the `apps` folder. I am pro reducing the time of loading of the app, since it could be a lil slow. I am not too knowledgable in this conversation. I was reading about parallel loading using `concurrent.futures`, Do you think that could be worth it to try (please if i am saying some dumb just let me know :D )

unkcpz:
> I was reading about parallel loading using concurrent.futures, Do you think that could be worth it to try

I don't think this is possible, it runs things asynchronously which still uses a single thread, the module load is not ID intense tasks and thus will no benefit from it. Moreover, it is not designed to import the modules.

unkcpz:
The rebase of https://github.com/aiidalab/aiidalab-qe/pull/435 on this are quite light, only 3 files need to solve the conflict. Which means the refactoring is well recognized by git.

PR #442, opened 2023-08-18
Move to App widget and call from notebook

unkcpz:
For the test purpose of https://github.com/aiidalab/aiidalab-qe/pull/428 proposed by @superstar54, we think moving the code from `qe.ipynb` into a module and creating an App class would make it not only more maintainable and easy for integrate unit test of the whole app as an widget.


unkcpz:
@superstar54 you mentioned the dlink may not necessary. I think it depends on the scenario and if you can decouple step widgets by adding another "data" layer in between. If you want to make such a change, please make a dedicated PR for that.

src/aiidalab_qe/app/main.py
```diff
@@ -0,0 +1,166 @@
...
+from aiidalab_qe.app.common import AddingTagsEditor, QeAppWorkChainSelector
+from aiidalab_qe.app.configuration import ConfigureQeAppWorkChainStep
+from aiidalab_qe.app.result import ViewQeAppWorkChainStatusAndResultsStep
+from aiidalab_qe.app.structure import Examples, StructureSelectionStep
+from aiidalab_qe.app.submission import SubmitQeAppWorkChainStep
+
+OptimadeQueryWidget.title = "OPTIMADE"  # monkeypatch
+
+
+class MainApp(ipw.VBox):
```
superstar54, line 30: `MainApp` may mean there are other apps somewhere. Better use `App` or `QeApp`.

src/aiidalab_qe/app/__init__.py
```diff
@@ -12,4 +13,5 @@
     "SubmitQeAppWorkChainStep",
     "ViewQeAppWorkChainStatusAndResultsStep",
     "WorkChainSelector",
```
superstar54, line 15: we don't need to import these sub-modules here. they will not be used anywhere.

superstar54 (changes requested):

src/aiidalab_qe/app/main.py
unkcpz: I was thinking of using the "main" as the C main function as the entry. No preference here. 

> MainApp may mean there are other apps somewhere.

So you think it will be other apps or not? `App` to me is too general `QeApp` is a bit redundant, because the whole thing is called QeApp already. 
Any preference from others @AndresOrtegaGuerrero @mikibonacci?

src/aiidalab_qe/app/__init__.py
unkcpz: True, it was there so I didn't delete it. But since we move things to a file, we can remove this.

src/aiidalab_qe/app/main.py
unkcpz: Had an offline discussion with @superstar54 (we didn't fight, but almost :smile: ) Agreed on the compromised `App`. We think in two years, there won't be any new "Other" app added.

superstar54 (approved):
LGTM!

PR #446, opened 2023-08-23
update nscf kpoints distance

AndresOrtegaGuerrero:
When doing an override of the kpoints distance, the nscf calculation of PdosWorkChain is not updated. This PR fix that behaviour

### DIFF
--- a/src/aiidalab_qe/app/submission/__init__.py
+++ b/src/aiidalab_qe/app/submission/__init__.py
@@ -447,7 +447,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
                         self.advanced_settings.magnetization.get_magnetization()
                     )
 
-                if key in ["base", "scf"]:
+                if key in ["base", "scf" , "nscf"]:
                     if self.advanced_settings.kpoints.override.value:
                         pw_overrides[key][
                             "kpoints_distance"


AndresOrtegaGuerrero:
@unkcpz  another solution should be that the value of k-points distance should change the protocol widget, if the user kept moderate, but uses a k-point distance lower or equal to 0.1, the protocol should be change to "precision" , though that changes the selection of pseudo

unkcpz:
> if the kpoint distance used is lower than the ones defined in the protocol, the scf calculation will have a higher number of kpoints , while the nscf will have the value of the protocol.

Why this is a problem? From what I understand, the nscf here is sort of independent, it uses the charge density from the scf calculation. So the nscf can be run at sparse k-points. In terms of the accuracy of the result, I also don't think it is a problem. 

I think we need use the protocol defined for pdos https://github.com/aiidateam/aiida-quantumespresso/blob/main/src/aiida_quantumespresso/workflows/protocols/pdos.yaml. Can you check if it is already the case? 

My opinion is that we should not add additional logic for the parameters in QeApp but more rely on the protocols of qe plugin. There are exceptions such as the magnetization structures/parameters that not yet properly deal with by qe plugin and we add the logic in the QeApp.

AndresOrtegaGuerrero:
@unkcpz I did test with some system in which i was defining a lower kpoint_distance , this affected the pdos, since it had less kpoints in the nscf

AndresOrtegaGuerrero:
For example if the user has protocol moderate, and the user modifies the kpoints_distance lower than the one of precise say 0.05 , the pdos (nscf) will be with 0.1. So the override is modifying only the scf calculations accordingly but not taking into account the nscf

unkcpz:
Okay, I see your point. The kpoint_distance of nscf is not settable and always gets from protocol. The problem you encountered is the widget is provided to only override the kpoint_distance of scf calculation but not nscf. 
Does that make sense that when the kpoint_distance is overridden it propagates to nscf and overrides it as well with the same value? Or more properly, since we have new things that need to be overridden, why not provide the widget for it specifically?

Ideally, for everything we/user suppose to change for builder, there should be a widget design for it. Do you agree with it?

AndresOrtegaGuerrero:
@unkcpz I am not sure. The issue is due to two factors the procotol and the kpoints_distance(override). What i propose is that if an user overrides the kpoints_distance the app should take into account two things.
1) What is the protocol being used
2)If PDOS calculations is needed the nscf should be less or equal than the kpoints_distance override but not higher

unkcpz:
I see. The issue comes from override and there is no logic between two kpoints_distance in `aiida-quantumespresso`. I agree at the moment your solution is good. But maybe we can leave this PR open and bring this to the next meeting for discussion? I can foresee some potential issues that might happen in the future.

Some thoughts and concerns below:
- I have a question for @mbercx, Is that possible to add some of this logic to the qe plugin, so that override parameters can have logic to make it more physically valid? 
- The logic you implement here requires dependencies between override parameters from two parts, e.g. advanced setting for scf kpoints_distance, pdos setting for nscf kpoints_distance. After @superstar54's plugin idea is implemented, which means this needs the widget/variable interaction between two different plugins. @superstar54 do you see any problem of this?

superstar54:
> After @superstar54's plugin idea is implemented, which means this needs the widget/variable interaction between two different plugins. @superstar54 do you see any problem of this?

no, this does not need the interaction between two plugins. Step 2 only provides the parameters, and it's the plugin developer decides how to use these two kpoint distances. 

In my option, if the user decides to override the kpoint distance, all the point distances (base, scf, nscf) should be overridden. So there is no need to compare with the protocol anymore.

unkcpz:
Thanks @superstar54 

> no, this does not need the interaction between two plugins.

For this case yes, since one override (scf kpoint_distance) will be from advanced setting, it will be visible by all plugins. What if there are parameters that from plugin A, and plugin B might need it to decide the logic? Did you consider this scenario? How

superstar54:
> What if there are parameters that from plugin A, and plugin B might need it to decide the logic? Did you consider this scenario? How

`plugin` design assumes all the plugins are independent, so that they are pluginable.

If a plugin developer still wants to add logic (dependence) between plugins. As far as I can foresee, there are three levels of logic,
- creating builder. This is easy, because all parameters (from all plugins) are visible to all plugins. Plugin developers can decide how to use the parameters from other plugins.
- setting panel. This can be implemented because a plugin can read also the setting panel of another plugin. However, this needs the user to set the plugin step by step.
- workchain. Adding dependency between sub-workchains (e.g. first run plugin A, then use the result of plugin A to run plugin B),  is not supported currently, and also not supported by the next workchain PR. Maybe supported in the future if we use the `WorkTree`.

unkcpz:
Okay, makes sense to me. But it means we need to be careful about how to separate plugins, the band structure/ pdos probably need to be one panel called "electronic structure". This will also guide how properties plugin developers were to add the widgets or create a brand new plugin.

superstar54:
Yes, a clear doc is needed.

mbercx:
> I have a question for @mbercx, Is that possible to add some of this logic to the qe plugin, so that override parameters can have logic to make it more physically valid?

Hi @unkcpz! I'm not sure what solution you propose here. Do you want the `aiida-quantumespresso` plugin to add some logic inside the work chain that in case the user changes the `kpoints_distance` of the SCF via the `overrides`, the `kpoints_distance` of the NSCF is adapted as well?

unkcpz:
> Do you want the aiida-quantumespresso plugin to add some logic inside the work chain that in case the user changes the kpoints_distance of the SCF via the overrides, the kpoints_distance of the NSCF is adapted as well?

Yes, that's what I thought. Do you have a follow-up discussion on how to implement override parameters with Lorenzo and Giovanni? My point is for the QeApp, if there are things that need more scientific logic, it better go to `aiida-quantumespresso`.

mbercx:
> Yes, that's what I thought. Do you have a follow-up discussion on how to implement override parameters with Lorenzo and Giovanni? My point is for the QeApp, if there are things that need more scientific logic, it better go to aiida-quantumespresso.

I'm going to be quite resistant to including any work chain logic that overrides parameters in the steps of the outline. I'm even opposed to setting defaults in the work chain, see https://github.com/aiidateam/aiida-quantumespresso/pull/902. ^^

If I understand correctly, you are even proposing overriding inputs that are set in the work chain. It would be hard to convince me that this is a good idea.

> Do you have a follow-up discussion on how to implement override parameters with Lorenzo and Giovanni?

Not really. Most work on that side is on hold until I am released from my scientific responsibilities. Happy to restart this once I move to PSI and have the time. 🙃

unkcpz:
> I'm going to be quite resistant to including any work chain logic that overrides parameters in the steps of the outline. I'm even opposed to setting defaults in the work chain, see https://github.com/aiidateam/aiida-quantumespresso/pull/902. ^^

Nice, if we have unanimous agreement on the design principle, I am all good to stick with it. If I understand correctly, the aiida-quantumespresso is supposed to be made as flexible as possible to get even very "un-scientific" inputs and handle the errors. 

> Most work on that side is on hold until I am released from my scientific responsibilities. Happy to restart this once I move to PSI and have the time. 🙃

May force be with you 🤌 Looking forward to seeing you here!!

mbercx:
> If I understand correctly, the aiida-quantumespresso is supposed to be made as flexible as possible to get even very "un-scientific" inputs and handle the errors.

Exactly! Who are we to say what's reasonable? The work chain simply implements the logic, i.e. steps of the workflow. The protocol defined what we think is a reasonable set of inputs, but the user can change everything before submission. What the user sees in the builder is what the work chain will actually run with. And it is their responsibility to make sure the inputs are correct.

The only caveats are: 

1. Some inputs are dynamically changed, e.g. through the use of an error handler.
2. If the inputs will simply not work at all: validate and raise an error, e.g. [not providing a pseudo for one of the elements in the structure.](https://github.com/aiidateam/aiida-quantumespresso/blob/7263f92ccabcfc9f828b9da5473e1aefbc4b8eca/src/aiida_quantumespresso/calculations/__init__.py#L171-L175)
3. If the inputs will run but are most likely incorrect: validate and warn the user, e.g. [providing a parent folder for certain calculations that will not set inputs to restart from the parent folder by default in `pw.x`](https://github.com/aiidateam/aiida-quantumespresso/blob/7263f92ccabcfc9f828b9da5473e1aefbc4b8eca/src/aiida_quantumespresso/calculations/pw.py#L193-L207).

unkcpz:
Have this a second look, according to what @mbercx replied, the logic is better to add in here in QeApp. I'll rebase this after the advance setting is re-organized and get this merge. We don't need a further discussion I think.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -449,9 +458,18 @@ def _get_qe_workchain_parameters(self) -> QeWorkChainParameters:
...
                 if key in ["base", "scf"]:
                     if self.advanced_settings.kpoints.override.value:
-                        pw_overrides[key][
-                            "kpoints_distance"
-                        ] = self.advanced_settings.kpoints.distance.value
+                        kpoints_distance = self.advanced_settings.kpoints.distance.value
+                        pw_overrides[key]["kpoints_distance"] = kpoints_distance
+
+                        if protocol == "fast" and (
+                            kpoints_distance < 0.5 and kpoints_distance > 0.15
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: @mbercx I should modify this logic it should be 
` if protocol == "fast" and kpoints_distance < 0.5`
My idea is that is kpoint_distance is set less to 0.5 (protocol) then nscf should be the same the user is setting. (or less but i dont know at what percentage we should decrease)

mbercx:
@AndresOrtegaGuerrero not sure what happened, but it seems there are no longer any changes in this PR?

AndresOrtegaGuerrero:
@mbercx I am also puzzle , i just merged with the main, i will check

AndresOrtegaGuerrero:
@mbercx Now that the app is using plugin, the logic of override is done differently , from what I noticed the kpoints_distance in nscf is ignore. Am I wrong on this one @superstar54 ?

tests/test_submit_qe_workchain/test_create_builder_default.yml
```diff
@@ -64,7 +64,7 @@ pdos:
       DOS:
         DeltaE: 0.02
   nscf:
-    kpoints_distance: 0.1
+    kpoints_distance: 0.12
```
mbercx, line 67: Why this change? What's this file, actually?

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -11,12 +11,17 @@ def get_builder(codes, structure, parameters, **kwargs):
...
-    #
+    kpoints_distance = parameters["advanced"]["kpoints_distance"]
+
     scf_overrides = deepcopy(parameters["advanced"])
     nscf_overrides = deepcopy(parameters["advanced"])
-    nscf_overrides.pop("kpoints_distance", None)
-    nscf_overrides["pw"]["parameters"]["SYSTEM"].pop("smearing", None)
-    nscf_overrides["pw"]["parameters"]["SYSTEM"].pop("degauss", None)
+    if protocol == "fast" and (kpoints_distance < 0.5):
+        nscf_overrides["kpoints_distance"] = kpoints_distance
```
mbercx, line 19: As previously noted, I think approach is a little strange as it defeats the purpose of the NSCF (restarting from the charge density with a denser k-points mesh).

My vote would go to: if the user decides to override the k-points, they should have access to all densities.

As a side note (warning: incoming scope creep), I wonder how many users can think in terms of k-points distance. Maybe some guidelines or examples might be helpful, if not there already.

mbercx (commented):
Thanks @AndresOrtegaGuerrero! Just a few comments.

tests/test_submit_qe_workchain/test_create_builder_default.yml
AndresOrtegaGuerrero: this file is a yml to compare the builder generated from the app. This change is necessary since the nscf logic is being modify by this PR

src/aiidalab_qe/plugins/pdos/workchain.py
AndresOrtegaGuerrero: So another approach could be to get a pdos `Settings` tab , and there the user can select the kpoints-distance for the nscf calculation. Regarding the k-points distance , the app already includes a widget that it displays the kpoint mesh that the calculation will be using

mbercx: I'd have to check the current interface (on my phone at the moment ^^), but either sounds fine by me.

AndresOrtegaGuerrero:
@mbercx  I re-implemented the PR, by using Setting Panel in the pdos pluggin

src/aiidalab_qe/plugins/pdos/setting.py
```diff
@@ -0,0 +1,79 @@
...
+    protocol = tl.Unicode(allow_none=True)
+
+    def __init__(self, **kwargs):
+        self.settings_title = ipw.HTML(
+            """<div style="padding-top: 0px; padding-bottom: 0px">
+            <h4>Settings</h4></div>"""
+        )
+        # nscf kpoints setting widget
+        self.nscf_kpoints_distance = ipw.BoundedFloatText(
+            min=0.0,
```
superstar54, line 33:
```suggestion
            min=0.001,
```
and remove the `if else` in lines 61-69.

src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -81,6 +86,13 @@ def __init__(self, **kwargs):
                 self.workchain_settings.properties[identifier].run.observe(
                     self._update_panel, "value"
                 )
+            # link the trailets if they exist in the plugin specific settings
+            for trailet in trailets_list:
+                if hasattr(self.settings[identifier], trailet):
+                    ipw.dlink(
+                        (self.advanced_settings, trailet),
+                        (self.settings[identifier], trailet),
+                    )
```
superstar54, line 95: since the protocol is added here, please remove lines 79-84.

superstar54 (changes requested):

superstar54:
@AndresOrtegaGuerrero Thanks. very clear implementation. Two mirror changes are needed. I will approve it after your changes.

src/aiidalab_qe/plugins/pdos/setting.py
```diff
@@ -0,0 +1,79 @@
...
+import ipywidgets as ipw
+import traitlets as tl
+from aiida import orm
+from aiida_quantumespresso.calculations.functions.create_kpoints_from_distance import (
+    create_kpoints_from_distance,
+)
+
+from aiidalab_qe.common.panel import Panel
+
+# Periodicity
```
superstar54, line 12:
```suggestion
# 
```

Please add a comment on why you use these map values.

src/aiidalab_qe/app/configuration/__init__.py
AndresOrtegaGuerrero: I was wondering how to use it as a trailets so it takes into account the changes in settings ?

superstar54 (approved):
LGTM!

PR #447, opened 2023-08-23
increasing the range of the magnetization floattext

AndresOrtegaGuerrero:
This PR increases the value of the magnetization widget, the values -1, 1 were too limited leading to low starting_magnetization values in the pw.x calculation


unkcpz (approved):
Thanks @AndresOrtegaGuerrero, the change looks good to me. 

Just curious why it was set to `+-1` in the first place. I am ignorance of the magnetization calculation, is `4` too high, or can be higher, let's see `10`?

AndresOrtegaGuerrero:
@unkcpz So if you check the code it uses input from the workchain `initial_magnetic_moments` and not directly the values of starting_magnetization. https://aiida-quantumespresso.readthedocs.io/en/latest/tutorials/magnetism.html

PR #449, opened 2023-08-23
Move docker stack build from docker stack to aiidalab-qe

unkcpz:
Attempt to solve https://github.com/aiidalab/aiidalab-docker-stack/issues/406

I adapt most of the codes I used in `aiidalab-docker-stack`, except in this repo, I call `docker buildx bake` directly instead of using `doit` script. 
It reveals that with this implementation, the integration test can be easily run with the image build. Also, provide a preview container on ghcr for immediate test of new feature implementation.

- [x] Move the notebook test to using the built image directly.


src/aiidalab_qe/app/configuration/pseudos.py
```diff
@@ -347,7 +347,7 @@ def _get_pseudos_family(self, pseudo_family: str) -> orm.Group:
                 .one()[0]
             )
         except exceptions.NotExistent as exception:
-            raise ValueError(
+            raise exceptions.NotExistent(
```
unkcpz, line 350: This change is separated as a single commit https://github.com/aiidalab/aiidalab-qe/pull/449/commits/6c194c6a6f9e396bcf37f4e14a30b1971ffa3ef6. The exception needs to be propagated up and not raise, otherwise if the pseudopotential library not installed, the app will crash when structure selected and confirm.

yakutovicha (commented):
Oh, wow, I wasn't expecting the PR to look so big.

Do we need all this complicated machinery? I thought we just made a tiny docker file + 2 bash scripts.

I believe it shouldn't be as complicated as the base repository where there are multiple stages/versions.

If you want, I can try to make a concurrent PR with what I have in mind.

unkcpz:
> Oh, wow, I wasn't expecting the PR to look so big.

I try to use all the state-of-art pr review feature from what I introduced to `aiidalab-docker-stack`, since we have it there, these codes are just duplicate some of it and create images not only for amd64 but images for arm64 and run tests on it. 

> If you want, I can try to make a concurrent PR with what I have in mind.

Sure, please try it. But I am doubt there is a simple one with all the useful features exists here, see the checkboxes below:
- [ ] release arm64/amd64 and combine these into one tag. check `docker-merge.yml`
- [ ] release for ghcr and for dockerhub as well when making new release.
- [ ] The PR can be reviewed in the ghcr. (I realize this is quite useful when reviewing PRs, the newly implemented features can quickly be seen by the reviewer.)

.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,92 @@
...
+              with:
+                  images: |
+                      name=${{ inputs.registry }}/${{ env.OWNER }}/${{ inputs.image }}
+                  tags: |
+                      type=edge,enable={{is_default_branch}}
+                      type=sha,enable=${{ github.ref_type != 'tag' }}
+                      type=ref,event=pr
+                      type=match,pattern=v(\d{4}\.\d{4}),group=1
+                      type=raw,value={{tag}},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+                      type=raw,value=qe-${{ env.QE_VERSION }},enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```
unkcpz, line 65: FYI, the tags are created here when making release. For the pr, there is "pr" and "sha" activated.

unkcpz:
Hi @yakutovicha, I deployed the macOS action runner on PSI Machine with Marnik, and all tests are passed. I believe to support ARM64 especially run on self-hosted runner on macOS the complicities here are difficult to avoid. But if your app doesn't need support arm64, then it is possible to have a light CI to build and test, you can do the test on your machine. 

Feel free to have a go-through on the implementation, if you have no objection I'll merge it tonight.

.github/actions/create-dev-env/action.yml
```diff
@@ -0,0 +1,39 @@
...
+              pip install --upgrade -r docker/requirements-dev.txt
+          shell: bash
+
+        - name: Install Dev Dependencies 📦
+          if: ${{ inputs.architecture == 'arm64' }}
+          run: |
+              echo $(whoami)
+              echo $PATH
+              echo $(docker context ls)
+              echo $(docker ps -a)
```
unkcpz, line 36:
```suggestion
```

PR #450, opened 2023-08-30
Docker stack full baked

unkcpz:
simplify #449 with tuck codes in one workflow CI.


unkcpz:
Hi @yakutovicha, I gave it a try this morning and have to give up here. I really think it brings more confusing on how every github action is run in sequence. There is another caveat on using multi-arch bake action that build arm64 on amd64 is very slow, it finished 5 mins in arm machine but took around 20 mins to finish.

unkcpz:
I removed the parameters for qe version and base image in https://github.com/aiidalab/aiidalab-qe/pull/449 as we discussed. For the GitHub action, if you take a close look, it is well-modulized and understandable in my opinion (every command can reproduced locally not like using `docker/bake` action it is very hard to debug since we have to use `jq` to parse parameters from other action steps).

unkcpz:
Close for the moment, not able to only use docker/back action to make the build and push.

PR #451, opened 2023-08-30
Standardize setting-like widgets with tests

unkcpz:


unkcpz:
I had a discussion with @superstar54 yesterday on how to standardize all such "settings" widgets to make them more uniformed and easy to test. 
This PR is an attempt. Let's try to make these two setting widgets as perfect as possible and we can then apply it to other settings. 

For the override, I plan to have another PR to remove it for sub-widgets and only keep one for the pw parameters override.

AndresOrtegaGuerrero:
> I had a discussion with @superstar54 yesterday on how to standardize all such "settings" widgets to make them more uniformed and easy to test. This PR is an attempt. Let's try to make these two setting widgets as perfect as possible and we can then apply it to other settings.
> 
> For the override, I plan to have another PR to remove it for sub-widgets and only keep one for the pw parameters override.

unkcpz:
I continue on changing `TotalCharge` widget as well, by this commit https://github.com/aiidalab/aiidalab-qe/pull/451/commits/6fe5ff349308413692841e920d9eba93f3786d68
I think this one is very worth discussing. @superstar54 was proposed that KpointSettings is not needed but can directly use `FloatText` widget in `AdvancedSettings`. During refactoring it I find since it makes more sense that kpoint_settings depend on protocol directly thus with logic it is better to make it a widget to deal with the interaction between widgets. 

For the `TotalCharge` widget, I'll let @AndresOrtegaGuerrero and @superstar54 vote on whether it worth being constructed as a widget. I personally think it should, otherwise the override logic is not that easy to implement properly, as well as `reset`.

AndresOrtegaGuerrero:
> 

But are we keeping to modify kpoints_distance ?

unkcpz:
Thanks for the green light! I'll have a follow-up PR to bundle things to advanced settings and at the end keep only one override for all such settings. 
Now we sort of have an agreement on when to construct a widget from small ones. The rule of thumb is if the small widget has its logic controlled by a parent widget (in our case, the kpoint setting widget is included by Advanced setting and having reset and override logic handled by the advanced setting widget) then there is litte need to create a dedicate widget for it. Only where there is more complex logic that is not able to be managed by the parent widget, create a widget to deal with interactions (in our case, the smearing needs an interface to tell its caller if it is activated).

AndresOrtegaGuerrero (approved):
Looks good to me!

PR #456, opened 2023-09-04
Plugin band

superstar54:
This PR uses the plugin design to implement a basic version of `bands`. It only adds the `UI` part:
- setting panel, It is only used to illustrate that the setting panel appears when the user selects the bands property. No parameters from this panel are used in the workchain.
- result panel. This is a panel only to show the bands. The code for `electronic_structure` tab is still kept.

### DIFF
--- a/src/aiidalab_qe/app/common/panel.py
+++ b/src/aiidalab_qe/app/common/panel.py
@@ -1,0 +1,127 @@
+# -*- coding: utf-8 -*-
+"""Class to .
+
+Authors:
+
+    AiiDAlab Team
+"""
+import ipywidgets as ipw
+
+DEFAULT_PARAMETERS = {}
+
+
+class Panel(ipw.VBox):
+    """Base class for all the panels.
+
+    The base class has a method to return the value of all the widgets in
+    the panel as a dictionary. The dictionary is used to construct the
+    input file for the calculation. The class also has a method to load a dictionary to set the value of the widgets in the panel.
+
+    title: the title to be shown in the GUI
+    identifier: which plugin this panel belong to.
+
+    """
+
+    title = "Panel"
+
+    def __init__(self, parent=None, identifier="panel", **kwargs):
+        """Initialize the panel.
+
+        :param kwargs: keyword arguments to pass to the ipw.VBox constructor.
+        """
+        self.parent = parent
+        if identifier:
+            self.identifier = identifier
+        super().__init__(
+            children=self.children,
+            **kwargs,
+        )
+
+    def get_panel_value(self):
+        """Return the value of all the widgets in the panel as a dictionary.
+
+        :return: a dictionary of the values of all the widgets in the panel.
+        """
+        return {}
+
+    def set_panel_value(self, parameters):
+        """Set the value of the widgets in the panel.
+
+        :param parameters: a dictionary of the values of all the widgets in the panel.
+        """
+        for key, value in parameters.items():
+            if key in self.__dict__:
+                setattr(self, key, value)
+
+    def reset(self):
+        """Reset the panel to the default value."""
+        self.set_panel_value(DEFAULT_PARAMETERS)
+
+    def _update_state(self):
+        """Update the state of the panel."""
+        pass
+
+
+class OutlinePanel(Panel):
+    title = "Outline"
+    description = ""
+
+    def __init__(self, **kwargs):
+        # Checkbox to see if the property should be calculated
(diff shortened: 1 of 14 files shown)


src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -155,3 +177,18 @@ def _default_state(self):
     def reset(self):
         with self.hold_trait_notifications():
             self.set_input_parameters(DEFAULT_PARAMETERS)
+
+    def _update_panel(self, _=None):
+        """Dynamic add/remove the panel based on the selected properties."""
+        # only keep basic and advanced settings
+        self.tab.children = self.tab.children[:2]
```
unkcpz, line 184: Can you define `children[:2]` as a more obvious variable in the `__init__` function, defined here seems a bit error-prone.

src/aiidalab_qe/app/common/panel.py
```diff
@@ -0,0 +1,127 @@
...
+Authors:
+
+    AiiDAlab Team
+"""
+import ipywidgets as ipw
+
+DEFAULT_PARAMETERS = {}
+
+
+class Panel(ipw.VBox):
```
unkcpz, line 13: The `Panel` serve not only as a viewer but also a controller for setting the data, is there a way to decouple this two and use the sub-class as:

```python
class ResultPanel(Panel, ipw.VBox):
```

unkcpz (commented):
Thanks a lot for separating this out! The logic makes sense to me. Just some technical problem to be further discussed. 
I think it is also possible to have tests for the Panel.

tests/test_submit_qe_workchain/test_create_builder_default.yml
```diff
@@ -126,8 +126,8 @@ pdos:
       pseudos:
         Si: Si.upf
 properties:
-- bands
 - pdos
+- bands
```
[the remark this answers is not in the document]
superstar54: Nothing special happened here. This is just the order when appending the properties. Since the `bands` is a plugin, it is added after the `pdos`, which is hardcoded.

src/aiidalab_qe/app/common/panel.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+
+    title = "Panel"
+
+    def __init__(self, parent=None, identifier="panel", **kwargs):
+        """Initialize the panel.
+
+        :param kwargs: keyword arguments to pass to the ipw.VBox constructor.
+        """
+        self.parent = parent
```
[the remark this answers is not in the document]
superstar54: The `parent` of a panel is where the panel is attached. This is useful when sharing data inside the `parent`. When the `parent` is `None`, means the panel is not attached to any step. This happened when we testing the individual panel.

src/aiidalab_qe/app/common/panel.py
```diff
@@ -0,0 +1,127 @@
...
+            {self.description}</div>"""
+        )
+        self.accordion = ipw.Accordion(children=[self.description_html])
+        self.accordion.selected_index = None
+        # self.children = [self.run, self.accordion]
+        self.children = [self.run, self.description_html]
+        super().__init__(**kwargs)
+
+    def get_panel_value(self):
+        return {f"{self.identifier}_run": self.run.value}
```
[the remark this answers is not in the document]
superstar54: This is from the old `bands_run` and `pdos_run`.

src/aiidalab_qe/app/configuration/__init__.py
superstar54: I use `builtin_settings` them.

src/aiidalab_qe/app/common/panel.py
```diff
@@ -0,0 +1,126 @@
...
+            if key in self.__dict__:
+                setattr(self, key, value)
+
+    def reset(self):
+        """Reset the panel to the default value."""
+        self.set_panel_value(DEFAULT_PARAMETERS)
+
+    def _update_state(self):
+        """Update the state of the panel."""
+        pass
```
unkcpz, line 62:
```suggestion
```

tests/test_entry_point.py
```diff
@@ -0,0 +1,6 @@
+def test_entry_point():
+    from aiidalab_qe.app.utils import get_entries
+
+    entries = get_entries()
+    print(entries)
```
unkcpz, line 5:
```suggestion
```

unkcpz (approved):

PR #458, opened 2023-09-05
Add plugin entry point to workchain 

superstar54:
This PR uses `bands` property as an example to implement plugin design into the workchain.

- to avoid cyclic import, I moved the `plugins` and `common` folders out from the app. This is also reasonable since both of them do not belong to the app. 
- Expose input and output of plugin workchain
- `run_plugin` method to run all plugins in parallel.
  - pros: run plugins in parallel
  - cons: can not set dependency between plugins, e.g. share the `scf` foder between `bands` and `pdos`.


src/aiidalab_qe/workflows/__init__.py
```diff
@@ -39,41 +61,53 @@ def define(cls, spec):
...
+                    "populate_defaults": False,
+                    "help": f"Inputs for the {name} plugin.",
+                },
+            )
+            spec.expose_outputs(plugin_workchain,
+                                namespace=name,
+                                namespace_options={"required": False},
+                                )
+            spec.exit_code(
+                404 + 1,
```
unkcpz, line 91: ? :smile_cat:

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -16,14 +16,36 @@
 BandsData = DataFactory("core.array.bands")
 Orbital = DataFactory("core.orbital")
 
-PROPERTIES_LIST = ["relax", "bands", "pdos"]
 
+# I removed the `validate_properties`, because the `properties` to be
+# calculated are not fixed anymore. It depends on the installed plugins.
```
unkcpz, line 21: Remove this before merging.

unkcpz (approved):
Perfect! Very clear now, thanks!

unkcpz:
Wait, seems there is a failed test, and it is caused by the plugin design. You can go and check the screenshot of the module loading failure in https://github.com/aiidalab/aiidalab-qe/actions/runs/6085956137?pr=458

superstar54:
> Wait, seems there is a failed test, and it is caused by the plugin design.

Thanks for pointing out this. It was solved by adding `__init__.py` file to the plugins folder.

PR #459, opened 2023-09-06
make pdos a plugin

superstar54:
Follow PR #456 and PR #458 , this PR makes the `pdos` as a plugin. It implements:
- UI part. Only the outline and result panel are added.
- Workchain part.

### DIFF
--- a/src/aiidalab_qe/plugins/pdos/result.py
+++ b/src/aiidalab_qe/plugins/pdos/result.py
@@ -1,0 +1,213 @@
+"""PDOS results view widgets
+
+"""
+
+import ipywidgets as ipw
+from aiida.orm import ProjectionData
+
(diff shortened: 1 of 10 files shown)


src/aiidalab_qe/plugins/pdos/result.py
unkcpz, line 1: I think most of the functions in this file moved from the `result.electronic_structure` module, but I didn't see the original codes were deleted.

unkcpz (changes requested):
Just one minor request, the refactoring looks very nice! Thanks!

unkcpz (approved):
As we discussed offline, the original code are required by both pdos and bands plugins. It'll be better to find a better way to put the shared modules of multiple plugins. @superstar54 will take care of it after and open a dedicate issue for it. I approve this PR.

PR #462, opened 2023-09-08
move get_configuration_parameters from submission to configuration

superstar54:
- move get_configuration_parameters from submission to configuration
- link the value of the `configuration  step` to the `input_parameters` of the submit step

### DIFF
--- a/src/aiidalab_qe/app/main.py
+++ b/src/aiidalab_qe/app/main.py
@@ -83,12 +83,8 @@ class App(ipw.VBox):
             (self.submit_qe_app_work_chain_step, "previous_step_state"),
         )
         ipw.dlink(
-            (self.configure_qe_app_work_chain_step, "workchain_settings"),
-            (self.submit_qe_app_work_chain_step, "workchain_settings"),
-        )
-        ipw.dlink(
-            (self.configure_qe_app_work_chain_step, "advanced_settings"),
-            (self.submit_qe_app_work_chain_step, "advanced_settings"),
+            (self.configure_qe_app_work_chain_step, "value"),
+            (self.submit_qe_app_work_chain_step, "input_parameters"),
         )
 
         ipw.dlink(
@@ -156,3 +152,7 @@ class App(ipw.VBox):
                         WizardAppWidgetStep.State.SUCCESS
                     )
                     self.submit_qe_app_work_chain_step.process = process
+
+    @property
+    def steps(self):
+        return self._wizard_app_widget.steps
--- a/src/aiidalab_qe/app/main.py
+++ b/src/aiidalab_qe/app/main.py
@@ -142,7 +142,3 @@ class App(ipw.VBox):
                     self.structure_step.confirmed_structure = process.inputs.structure
                     self.configure_step.state = WizardAppWidgetStep.State.SUCCESS
                     self.submit_step.process = process
-
-    @property
-    def steps(self):
-        return self._wizard_app_widget.steps
--- a/src/aiidalab_qe/app/main.py
+++ b/src/aiidalab_qe/app/main.py
@@ -48,58 +48,52 @@ class App(ipw.VBox):
             storable=False,
             configuration_tabs=["Cell", "Selection", "Appearance", "Download"],
         )
-        self.structure_selection_step = StructureSelectionStep(
+        self.structure_step = StructureSelectionStep(
             manager=self.structure_manager_widget, auto_advance=True
         )
-        self.structure_selection_step.observe(
-            self._observe_structure_selection, "structure"
-        )
+        self.structure_step.observe(self._observe_structure_selection, "structure")
 
-        self.configure_qe_app_work_chain_step = ConfigureQeAppWorkChainStep(
-            auto_advance=True
-        )
-        self.submit_qe_app_work_chain_step = SubmitQeAppWorkChainStep(
+        self.configure_step = ConfigureQeAppWorkChainStep(auto_advance=True)
+        self.submit_step = SubmitQeAppWorkChainStep(
             auto_advance=True, qe_auto_setup=qe_auto_setup
         )
-        view_qe_app_work_chain_status_and_results_step = (
-            ViewQeAppWorkChainStatusAndResultsStep()
-        )
+        self.results_step = ViewQeAppWorkChainStatusAndResultsStep()
 
         # Link the application steps
         ipw.dlink(
-            (self.structure_selection_step, "state"),
-            (self.configure_qe_app_work_chain_step, "previous_step_state"),
+            (self.structure_step, "state"),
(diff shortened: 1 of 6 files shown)


tests/conftest.py
```diff
@@ -209,57 +219,51 @@ def submit_step_widget_generator(
...
         )
-
         # mimic the behavior of the smearing widget set up
-        submit_step.advanced_settings.smearing.smearing.value = smearing
-        submit_step.advanced_settings.smearing.degauss.value = degauss
+        configure_step.advanced_settings.smearing.smearing.value = smearing
+        configure_step.advanced_settings.smearing.degauss.value = degauss
+        configure_step.confirm()
+        #
+        submit_step = app.steps[2][1]
```
unkcpz, line 261:
```suggestion
        submit_step = app.submit_qe_app_work_chain_step
```

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -399,136 +378,38 @@ def submit(self, _=None):
...
             dos_code=orm.load_code(dos_code),
             projwfc_code=orm.load_code(projwfc_code),
-            protocol=parameters.protocol,
-            relax_type=RelaxType(parameters.relax_type),
-            properties=parameters.properties,
-            spin_type=SpinType(parameters.spin_type),
-            electronic_type=ElectronicType(parameters.electronic_type),
-            overrides=parameters.overrides,
-            initial_magnetic_moments=parameters.initial_magnetic_moments,
+            protocol=parameters["protocol"],
```
unkcpz, line 406: As we discussed offline, we agree keep on using dataclass is better for the moment.

unkcpz (changes requested):
Thanks! Two minor comments.

src/aiidalab_qe/app/submission/__init__.py
superstar54: After another discussion, it is better to use `dict` directly. Because the `value` of the configure step is now a traitlet and linked to the `input_parameters` (also a trailet) of the submit step. 

```
    # output dictionary
    value = tl.Dict()
```

superstar54:
@unkcpz Thanks for the comments. I updated:
- In the report, remove `if` and use `.get(key, None)`, and remove all `None` values in the final step.
- Use the name of the step directly instead of steps[0][1]. Also, I renamed the step using a short name.

Please have another look.

src/aiidalab_qe/app/submission/__init__.py
unkcpz: Can we name it `configuration_parameters` as the function name has. The `value` of the step makes no much sense and is confusing.

src/aiidalab_qe/app/main.py
```diff
@@ -143,16 +133,16 @@ def _observe_process_selection(self, change):
...
-                    )
-                    self.submit_qe_app_work_chain_step.process = process
+                    self.structure_step.structure = process.inputs.structure
+                    self.structure_step.confirmed_structure = process.inputs.structure
+                    self.configure_step.state = WizardAppWidgetStep.State.SUCCESS
+                    self.submit_step.process = process
+
+    @property
+    def steps(self):
+        return self._wizard_app_widget.steps
```
unkcpz, line 148: Not sure I understand this. Where is this property used?

src/aiidalab_qe/app/result/report.py
```diff
@@ -27,47 +27,49 @@ def _generate_report_dict(builder_parameters: dict):
...
     """
     # Workflow logic
-    yield "relax_method", builder_parameters["relax_type"]
-    yield "relaxed", builder_parameters["run_relax"]
-    yield "bands_computed", builder_parameters["run_bands"]
-    yield "pdos_computed", builder_parameters["run_pdos"]
+    yield "relax_method", builder_parameters.get("relax_type", None)
+    yield "relaxed", builder_parameters.get("run_relax", None)
+    yield "bands_computed", builder_parameters.get("run_bands", None)
+    yield "pdos_computed", builder_parameters.get("run_pdos", None)
```
unkcpz, line 33: The `None` is the default return value if the key not defined when using `.get`. So just do `builder_parameters.get("relax_type")`

src/aiidalab_qe/app/result/report.py
```diff
@@ -91,6 +93,7 @@ def _fmt_yes_no(truthy):
     )
     template = resources.read_text(static, "workflow_summary.jinja")
     style = resources.read_text(static, "style.css")
+    report = {key: value for key, value in report.items() if value is not None}
```
unkcpz, line 96: :+1:

tests/test_submit_qe_workchain.py
```diff
@@ -47,11 +46,11 @@ def test_create_builder_default(
...
-    # There should be only one None from initial_magnetic_moments
-    assert report_html.count("None") == 1
-    if "initial_magnetic_moments" not in builder_parameters:
-        parsed = BeautifulSoup(report_html, "html.parser")
-        assert parsed.find("initial_magnetic_moments").text == "None"
+    # All None value is removed in the final report
+    assert report_html.count("None") == 0
+    # if "initial_magnetic_moments" not in builder_parameters:
+    #     parsed = BeautifulSoup(report_html, "html.parser")
+    #     assert parsed.find("initial_magnetic_moments").text == "None"
```
unkcpz, line 53: This actually change the behaviour of the original test and skipped the thing we want to test initially, that when the initial_magnetic_moment not set the report dict will not contain it.

Instead, of the changes, you can call `_generate_report_dict` and still have this behaviour tested.

unkcpz (changes requested):
@superstar54 Thanks! Just minor requests.

src/aiidalab_qe/app/main.py
superstar54: It was used for tests before, e.g. self.steps[1][1]. Now, we can use self.configure_step directly. So I removed it.

src/aiidalab_qe/app/result/report.py
superstar54: :+1: , i removed the `None`.

tests/test_submit_qe_workchain.py
superstar54: I find that there is nothing needs to be changed here, so I revert it.

superstar54:
> @superstar54 Thanks! Just minor requests.

Hi @unkcpz  Thanks! I changed as you suggested.

src/aiidalab_qe/app/submission/__init__.py
superstar54: yes, i changed it to `configuration_parameters`

unkcpz (approved):
All good! thanks!

PR #464, opened 2023-09-14
use get and set for WorkChain settings

superstar54:
The `get_configuration_parameters` method of `ConfigureQeAppWorkChainStep` is hard coded. In principle, it should only need to call the method from sub-settings (plugin's panel). e.g
```
        workchain_settings = self.workchain_settings.get_setting_parameters()
        advanced_settings = self.advanced_settings.get_setting_parameters()
```
Same for the `setting` method.
In this PR, I make this change for the `WorkChainSettings`. Next PR will be the `AdvancedSettings`.

### DIFF
--- a/src/aiidalab_qe/app/configuration/__init__.py
+++ b/src/aiidalab_qe/app/configuration/__init__.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 import ipywidgets as ipw
 import traitlets as tl
 from aiida import orm
-from aiida_quantumespresso.common.types import RelaxType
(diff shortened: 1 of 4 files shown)


unkcpz:
> In this PR, I make this change for the WorkChainSettings. Next PR will be the AdvancedSettings.

Go ahead, no need to be a separated PR. You can add more commits here.

superstar54:
> Thanks! The change looks all good. Please go ahead, no need to be a separated PR. You can add more commits here.

Thanks! Then, I will merge this first. Because the change in the `advanced setting` will be a little bit complicated, it is better to add it in another PR.

PR #465, opened 2023-09-15
add get and set for advanced settings

superstar54:
Follow #464 

- add get and set_setting_parameters for `advaced settings`.
- move the logic of getting overrides of `base`, `scf`, `bands` and `scf` to the individual plugin.
- link the `spin_type` and `electronic_type` to the `advanced setting`. Generally, we should only link the `protocol` of the `workchain_settings` to the plugin's setting. However, in the advanced setting, we also need the `spin_type` and `electronice_type`, which we hard-coded. If other plugins also need these values, they should directly access the `self.parent.workchain_settings.spin_type.value`.

### DIFF
--- a/src/aiidalab_qe/plugins/pdos/workchain.py
+++ b/src/aiidalab_qe/plugins/pdos/workchain.py
@@ -7,15 +7,26 @@ def get_builder(codes, structure, overrides, protocol, **kwargs):
     pw_code = codes.get("pw_code", None)
     dos_code = codes.get("dos_code", None)
     projwfc_code = codes.get("projwfc_code", None)
+    from copy import deepcopy
+
+    pw_code = codes.get("pw_code", {})
+    scf_overrides = deepcopy(overrides)
+    nscf_overrides = deepcopy(overrides)
+    nscf_overrides.pop("kpoints_distance", None)
+    nscf_overrides["pw"]["parameters"]["SYSTEM"].pop("smearing", None)
+    nscf_overrides["pw"]["parameters"]["SYSTEM"].pop("degauss", None)
+    overrides = {
+        "scf": scf_overrides,
+        "nscf": nscf_overrides,
+    }
     if dos_code is not None and projwfc_code is not None:
-        pdos_overrides = overrides.get("pdos", {})
         pdos = PdosWorkChain.get_builder_from_protocol(
             pw_code=pw_code,
             dos_code=dos_code,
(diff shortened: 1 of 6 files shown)


src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -7,15 +7,26 @@ def get_builder(codes, structure, overrides, protocol, **kwargs):
     pw_code = codes.get("pw_code", None)
     dos_code = codes.get("dos_code", None)
     projwfc_code = codes.get("projwfc_code", None)
+    from copy import deepcopy
+
+    pw_code = codes.get("pw_code", {})
+    scf_overrides = deepcopy(overrides)
+    nscf_overrides = deepcopy(overrides)
+    nscf_overrides.pop("kpoints_distance", None)
+    nscf_overrides["pw"]["parameters"]["SYSTEM"].pop("smearing", None)
```
unkcpz, line 16:
```suggestion
    nscf_overrides["pw"]["parameters"]["SYSTEM"].pop("smearing")
```

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -7,15 +7,26 @@ def get_builder(codes, structure, overrides, protocol, **kwargs):
     pw_code = codes.get("pw_code", None)
     dos_code = codes.get("dos_code", None)
     projwfc_code = codes.get("projwfc_code", None)
+    from copy import deepcopy
+
+    pw_code = codes.get("pw_code", {})
+    scf_overrides = deepcopy(overrides)
+    nscf_overrides = deepcopy(overrides)
+    nscf_overrides.pop("kpoints_distance", None)
```
unkcpz, line 15: Please open an issue for how to validate the parameters passed in for future improvement.

unkcpz (approved):
Thanks! All good, just two minor requests.

src/aiidalab_qe/plugins/pdos/workchain.py
superstar54: done, #466

src/aiidalab_qe/plugins/pdos/workchain.py
superstar54: I just recall that, in some cases, e.g. an insulator, there is no `smearing` value in the parameters. So the `None` is needed here.

PR #471, opened 2023-09-18
load_profile explicitly in qe.ipynb

unkcpz:
It is recommended to call `load_profile` explicitly. After `aiidalab-widgets-base` if not loaded a warning will throw up.


superstar54 (approved):
LGTM.

PR #472, opened 2023-09-18
add test for workchain

superstar54:
This PR adds tests for the QeAppWorkChain, and also includes the plugins: `bands` and `pdos`.

The test generates the `projection` data and `bands` data, then mounts them to the QeAppWorkChain. The final result is then checked.

### DIFF
--- a/src/aiidalab_qe/app/structure/__init__.py
+++ b/src/aiidalab_qe/app/structure/__init__.py
@@ -2,7 +2,6 @@
 
 Authors: AiiDAlab team
 """
-import os
 import pathlib
 import warnings
 
@@ -29,14 +28,14 @@ structures are supported.
 
 # The Examples list of (name, file) tuple curretly passed to
 # StructureExamplesWidget.
-file_path = pathlib.Path(__file__).parent.resolve()
+file_path = pathlib.Path(__file__).parent
 Examples = [
-    ("Silicon (diamond)", os.path.join(file_path, "examples/Si.xyz")),
-    ("Silicon oxide", os.path.join(file_path, "examples/SiO2.xyz")),
-    ("Diamond", os.path.join(file_path, "examples/diamond.cif")),
-    ("Gallium arsenide", os.path.join(file_path, "examples/GaAs.xyz")),
-    ("Gold (fcc)", os.path.join(file_path, "examples/Au.cif")),
-    ("Cobalt (hcp)", os.path.join(file_path, "examples/Co.cif")),
+    ("Silicon (diamond)", file_path / "examples" / "Si.xyz"),
+    ("Silicon oxide", file_path / "examples" / "SiO2.xyz"),
+    ("Diamond", file_path / "examples" / "diamond.cif"),
+    ("Gallium arsenide", file_path / "examples" / "GaAs.xyz"),
+    ("Gold (fcc)", file_path / "examples" / "Au.cif"),
+    ("Cobalt (hcp)", file_path / "examples" / "Co.cif"),
 ]
 
 


src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -25,15 +26,17 @@
...
 # The Examples list of (name, file) tuple curretly passed to
 # StructureExamplesWidget.
 Examples = [
-    ("Silicon (diamond)", "miscellaneous/structures/Si.xyz"),
-    ("Silicon oxide", "miscellaneous/structures/SiO2.xyz"),
-    ("Diamond", "miscellaneous/structures/diamond.cif"),
-    ("Gallium arsenide", "miscellaneous/structures/GaAs.xyz"),
-    ("Gold (fcc)", "miscellaneous/structures/Au.cif"),
-    ("Cobalt (hcp)", "miscellaneous/structures/Co.cif"),
+    ("Silicon (diamond)", os.path.join(file_path, "miscellaneous/structures/Si.xyz")),
```
unkcpz, line 34: Try using `pathlib.Path`?

unkcpz: BTW, why is this change needed?

tests/conftest.py
```diff
@@ -51,6 +73,91 @@ def _generate_structure_data(name="silicon"):
...
+            "magnetic_number": 0,
+            "radial_nodes": 1,
+            "position": [0.0, 0.0, 0.0],
+        }
+        orbitals = [OrbitalCls(**state_dict)]
+        # projections = np.array([[1]])
+        energy_arrays = np.array([1])
+        pdos_arrays = np.array([1])
+
+        projection_data = ProjectionData()
```
unkcpz, line 145: Here the datatype is not created from the data factory, please try to use the same way to create the data type.

tests/conftest.py
```diff
@@ -51,6 +73,91 @@ def _generate_structure_data(name="silicon"):
...
+        bands_data = generate_bands_data()
+        projection_data.set_reference_bandsdata(bands_data)
+        projection_data.set_projectiondata(
+            orbitals,
+            # list_of_projections=projections,
```
unkcpz, line 150: If it is not used, remove it.

tests/conftest.py
```diff
@@ -133,24 +240,25 @@ def _generate_upf_data(element, filename=None):
 @pytest.fixture
 def pw_code(aiida_local_code_factory):
     """Return a `Code` configured for the pw.x executable."""
+
     return aiida_local_code_factory(
-        label="pw@localhost", executable="bash", entry_point="quantumespresso.pw"
+        label="pw", executable="bash", entry_point="quantumespresso.pw"
```
unkcpz, line 245: The label is the full label, why the computer is removed?

tests_notebooks/test_qe_app.py
```diff
@@ -32,17 +32,19 @@ def test_qe_app_select_silicon_and_confirm(
...
+    try:
+        driver.find_element(By.XPATH, "//option[@value='Diamond']").click()
+        time.sleep(10)
+        element = WebDriverWait(driver, 60).until(
+            EC.element_to_be_clickable((By.XPATH, "//button[text()='Confirm']"))
+        )
+        element.click()
+    except Exception:
+        driver.find_element(By.TAG_NAME, "summary").click()
+        time.sleep(10)
```
unkcpz, line 44: Very useful! thanks!

unkcpz (changes requested):
@superstar54 thanks a lot! These are very useful tests. I only have minor requests good to me overall.

.github/workflows/ci.yml
```diff
@@ -51,11 +51,15 @@ jobs:
 
         runs-on: ubuntu-latest
         timeout-minutes: 30
+
         services:
+            postgres:
+                image: postgres:12
```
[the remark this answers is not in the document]
superstar54: Yes, you are right. Postgres is not needed here.

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -27,13 +29,14 @@
...
 # StructureExamplesWidget.
+file_path = pathlib.Path(__file__).parent.resolve()
 Examples = [
-    ("Silicon (diamond)", "miscellaneous/structures/Si.xyz"),
-    ("Silicon oxide", "miscellaneous/structures/SiO2.xyz"),
-    ("Diamond", "miscellaneous/structures/diamond.cif"),
-    ("Gallium arsenide", "miscellaneous/structures/GaAs.xyz"),
-    ("Gold (fcc)", "miscellaneous/structures/Au.cif"),
-    ("Cobalt (hcp)", "miscellaneous/structures/Co.cif"),
+    ("Silicon (diamond)", os.path.join(file_path, "examples/Si.xyz")),
```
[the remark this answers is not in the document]
superstar54: I don't know how this `/` works on a Windows machine. So I will use `os.path.join` here, which is safe for sure.

tests/conftest.py
```diff
@@ -51,6 +73,91 @@ def _generate_structure_data(name="silicon"):
...
+    """Return an ``XyData`` instance."""
+
+    def _generate_xy_data(xvals=None, yvals=None, xlabel=None, ylabel=None):
+        """Return an ``XyData`` node."""
+        from aiida.orm import XyData
+
+        xvals = xvals  # [1, 2, 3]
+        yvals = yvals  # [10, 20, 30]
+        xlabel = xlabel  # "X"
+        ylabel = ylabel  # ["dos"]
```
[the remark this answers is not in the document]
superstar54: I removed the comments and added a short description on the docstring.

tests/conftest.py
```diff
@@ -51,6 +73,91 @@ def _generate_structure_data(name="silicon"):
...
+    """Return a `BandsData` node."""
+
+    def _generate_bands_data():
+        """Return a `BandsData` instance with some basic `kpoints` and `bands` arrays."""
+        import numpy as np
+        from aiida.plugins import DataFactory
+
+        kpoints = np.array([[0.0, 0.0, 0.0]])
+        bands = np.array([[-5.64024889]])
+        BandsData = DataFactory("core.array.bands")
```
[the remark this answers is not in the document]
superstar54: Sure

tests/conftest.py
```diff
@@ -51,6 +73,91 @@ def _generate_structure_data(name="silicon"):
...
+        OrbitalCls = OrbitalFactory("core.realhydrogen")
+        state_dict = {
+            "kind_name": "C",
+            "angular_momentum": 0,
+            "magnetic_number": 0,
+            "radial_nodes": 1,
+            "position": [0.0, 0.0, 0.0],
+        }
+        orbitals = [OrbitalCls(**state_dict)]
+        # projections = np.array([[1]])
```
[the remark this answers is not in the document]
superstar54: Yes, I removed the lines.

tests/conftest.py
superstar54: Yes, I removed it.

tests/conftest.py
superstar54: The label is the code label, not the full label.

tests/conftest.py
superstar54: Sure, I used the `DataFactory` for all.

src/aiidalab_qe/app/structure/__init__.py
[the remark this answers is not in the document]
unkcpz: No, we never need to consider the windows, the AiiDAlab is supposed to run in a container, no matter whether we use aiidalab-launch or deployed somewhere else.

tests/conftest.py
unkcpz: But why `pw@localhost` was worked before? It must be somewhere else you change the corresponding code to use the label, no?

src/aiidalab_qe/app/structure/__init__.py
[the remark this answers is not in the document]
unkcpz: Meanwhile, the pathlib is also abstract for windows, it creates a path object that can easily convert to windows path later on. See https://docs.python.org/3/library/pathlib.html

tests/conftest.py
superstar54: It works even when you use `pw@localhost` as the label, but when you print out the code, it shows `pw@localhost@localhost`.

unkcpz: I see, thanks for the explanation.

src/aiidalab_qe/app/structure/__init__.py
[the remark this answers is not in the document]
superstar54: > the AiiDAlab is supposed to run in a container, no matter whether we use aiidalab-launch or deployed somewhere else.

Running in a container is a nice way, also the recommended way, to use AiiDAlab. However, since AiiDAlab is pure Python. The users can use it in a normal notebook. Especially after we define the aiidalab-qe app as an `App Class`, the user can easily import it and use it in a normal notebook as long as they have a aiida environment. 

>  the pathlib is also abstract for windows, it creates a path object that can easily convert to windows path later on. 

Unless we test it in Windows. Otherwise, and for the moment, I'd like to use `os.path`. Please open an issue to test it and update it if you prefer `pathlib` with `/`.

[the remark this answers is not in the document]
unkcpz: As I said, using os.path is not windows support but pathlib will.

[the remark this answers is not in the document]
unkcpz: > Unless we test it in Windows

To be honest, we will never do this. And I don't think we should mix using os.path and pathlib. You can change everything back to os.path, I am okay with that.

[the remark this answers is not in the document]
superstar54: Thanks! I searched for the difference. The `pathlib` is indeed a more modern method.  I changed it to use `pathlib` here.

unkcpz (approved):
Thanks for the work! @superstar54 
The chrome test failed with an error, I retrigger it and let's see if it is because of timeout.

unkcpz:
:clinking_glasses:

PR #474, opened 2023-09-19
Remove hard coded codes from workflow

superstar54:
The codes (`pw_code`, `dos_code`, `projwfc_code`) are hard coded in the builder, which is not good for the plugin design. This PR uses `get_selected_codes`. 
The `get_resources` method is also used to make the code more modularized.

### DIFF
--- a/src/aiidalab_qe/app/submission/__init__.py
+++ b/src/aiidalab_qe/app/submission/__init__.py
@@ -226,14 +226,16 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     def _auto_select_code(self, change):
         if change["new"] and not change["old"]:
             for code in [
-                "pw_code",
-                "dos_code",
-                "projwfc_code",
+                "pw",
+                "dos",
+                "projwfc",
             ]:
                 try:
-                    code_widget = getattr(self, code)
+                    code_widget = getattr(self, f"{code}_code")
                     code_widget.refresh()
-                    code_widget.value = orm.load_code(DEFAULT_PARAMETERS[code]).uuid
+                    code_widget.value = orm.load_code(
+                        DEFAULT_PARAMETERS["codes"][code]
+                    ).uuid
                 except NotExistent:
                     pass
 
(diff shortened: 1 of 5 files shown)


src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -412,30 +433,23 @@ def _create_builder(self) -> ProcessBuilderNamespace:
...
-        pw_code = self.pw_code.value
-        dos_code = self.dos_code.value
-        projwfc_code = self.projwfc_code.value
-
-        parameters = deepcopy(self.input_parameters)
+        self.builder_parameters = deepcopy(self.input_parameters)
+        # add codes info into input_parameters
+        self.builder_parameters["codes"] = self.get_selected_codes()
+        self.builder_parameters["resources"] = self.get_resources()
+        #
```
unkcpz, line 440:
```suggestion

```

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -346,6 +359,14 @@ def _on_submit_button_clicked(self, _):
         self.submit_button.disabled = True
         self.submit()
 
+    def get_selected_codes(self):
+        codes = {
```
unkcpz, line 363:
```suggestion
        """Return the dict of code name as key and code node UUID as value"""
        codes = {
```

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -346,6 +359,14 @@ def _on_submit_button_clicked(self, _):
...
+            "pw_code": self.pw_code.value,
+            "dos_code": self.dos_code.value,
+            "projwfc_code": self.projwfc_code.value,
```
unkcpz, line 366:
```suggestion
            "pw": self.pw_code.value,
            "dos": self.dos_code.value,
            "projwfc": self.projwfc_code.value,
```

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -134,7 +132,7 @@ def get_builder_from_protocol(
         relax_overrides = {"base": parameters["advanced"]}
         protocol = parameters["workchain"]["protocol"]
         relax_builder = PwRelaxWorkChain.get_builder_from_protocol(
-            code=pw_code,
+            code=codes.get("pw_code"),
```
unkcpz, line 135:
```suggestion
            code=codes.get("pw"),
```

unkcpz (changes requested):
Thanks! Just minor requests. @superstar54

superstar54:
@unkcpz I changed as you suggested. Since we replaced the `pw_code` with `pw`, the `qeapp.yaml` and the workchains are also updated.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -226,14 +226,16 @@ def _toggle_install_widgets(self, change):
...
-                "pw_code",
-                "dos_code",
-                "projwfc_code",
+                "pw",
+                "dos",
+                "projwfc",
             ]:
                 try:
-                    code_widget = getattr(self, code)
+                    code_widget = getattr(self, f"{code}_code")
```
unkcpz, line 234: Ah, I didn't notice this is directly used in `getattr`. Then the original code is good enough.

unkcpz (approved):
Thanks! Looks all good. 

I just merge https://github.com/aiidalab/aiidalab-qe/pull/449, and what to use it as a test. So I'll rebase this PR and see if from PR everything is built correctly. Hope you don't mind.

unkcpz:
Hi @superstar54, after rebasing the tests failed. But I compared the two branches and didn't find the difference. Do you have any clue why it is happening, can you check the changes are all correct?

PR #475, opened 2023-09-19
use uuid for pseudo

superstar54:
To reload all parameters in the configuration step, one has to save the `configuration_parameters` as an extra attribute of the `QeAppWorkChain` process, which means the `parameters` can not include an AiiDA Node, so the uuid is used.


superstar54:
> can you add the implementation to use this for loading the workflow pseudo setting back?

I added a `set_pseudos` method to load the parameters back. The test is also updated.

> The change itself seems not necessary, and I assume when it is used in the actual scenario more change may required, 

 The final scenario, which loads all configuration parameters back, can not be done now. It depends on other widgets. It's not good to mix the pseuods with other widgets.

unkcpz (approved):
Thanks! Looks all good to me.

PR #476, opened 2023-09-19
fix override in get_panel_value for advanced

superstar54:
#477 
We should output the values even if the `override` is not selected. In this case, the default values are used. Since the default values are also with `protocol`, so it safe to use them.


AndresOrtegaGuerrero (approved):
I tested and it works!,  LGTM!

Issue #211, opened 2022-03-01
Support bandstructure calculation for two-dimensional materials


unkcpz:
The main blocker of this feature is that the seekpath not yet support the kpath finding for all types of two-dimensional structures. Trace back to the discussion:

That implementation was not general enough. It required the cell to be already oriented in a given direction.
The difficult part is to standardise the orientation of the cell in the 2D case. 
If this is done, it's easy to then implement the definition of high-symmetry points for 2D.

_Originally posted by @giovannipizzi in https://github.com/giovannipizzi/seekpath/issues/76#issuecomment-618851600_

For the input structure which already has periodic_type=[True, True, False] the QeApp now raise the exception that it is not supported. We should in the future, create a input structure with proper orientation and use seekpath to get the correct k-path. While for the input structure that already has a vacuum layer set, the pymatgen detect it is 2D structure, besides raise a warning, the structure needs to be converted to the proper orientation so the seekpath can work properly.

giovannipizzi:
Probably this can be (almost) closed. Before closing, though, a few notes/request. 

1. Now the text "If your system has periodicity xy. Please select one of the five 2D Bravais lattices corresponding to your system." and the dropdown always appears, also for 3D systems. I think it's confusing, can this be hidden and one says simply "The path is selected automatically using seekpath [link]" if the system is 3D? 
2. When 2D, as discussed, use the logic in AiiDA/tools/data/array/kpoints/legacy
3. Both in 2D and 3D (for the future maybe, to move to a different issue?) can we have a way to specify optionally a list of k-points for the path? in the form 
   
    kpt_start kpt_end num_points
    kpt_start kpt_end num_points
    ...

unkcpz:
Thanks @giovannipizzi.

@AndresOrtegaGuerrero, let's find a time together to try to address them.

AndresOrtegaGuerrero:
Thank you , for the code in aiidalab_qe I implemented the same as here https://pubs.acs.org/doi/10.1021/acs.jpclett.2c02972 
I checked the function for 2d, https://github.com/aiidalab/aiidalab-qe/blob/98ebc43e66fbd0a49e47c89e2d24e20ebc084cc0/src/aiidalab_qe/plugins/bands/workchain.py#L59 

I noticed that there is difference in the oblique and in the  centered rectangular ,

PR #479, opened 2023-09-19
Allow structures with different periodicity

AndresOrtegaGuerrero:
The current PR allows QE app to use 1D along x-axis and 2D -xy materials 
- [X] Updata aiidalab-widget-base to newer version to allow periodicity view
- [X] Modify AddTagEditor to StructureData to allow to change of periodicity of StructureData
- [X] Removing constraint of 3D systems only
- [X] Generation of kpoints for 1D and 2D points based on kpoints_distance 
- [X] Selection of 2D symmetry path in Settings


superstar54 (changes requested):
@AndresOrtegaGuerrero Thanks for the PR. Two comments:
- periodicity belongs to the cell. I think moving it to the `Edit Cell` tab is better.
- The test coverage is down, but don't worry. I have a draft PR to test the plugin. So, for now, it's fine leave it

src/aiidalab_qe/plugins/bands/setting.py
```diff
@@ -50,6 +61,7 @@ def get_panel_value(self):
         return {
             "path": Str(self.path.value),
             "npoint": Int(self.npoint.value),
+            "two_dim_kpoints_path": self.two_dim_kpoints_path.value,
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: @superstar54 Thank you i will do the modifications, the only thing about the tab, if i do that change it will mean to do a PR in aiidalab widgets base , and then it means to close this PR until that one is approved. What do you think ?

src/aiidalab_qe/app/main.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, qe_auto_setup=True):
             editors=[
                 BasicCellEditor(title="Edit cell"),
                 BasicStructureEditor(title="Edit structure"),
-                AddingTagsEditor(title="Edit tags"),
+                AddingTagsEditor(title="Edit StructureData"),
```
unkcpz, line 45: The `StructureData` is aiida specific and I'd suppose QeApp user don't understand it, so maybe call it "Edit site tags"?

src/aiidalab_qe/app/result/report.py
```diff
@@ -14,6 +14,12 @@
     "precise": "SSSP/1.2/PBE/precision",
 }
 
+PROTOCOL_BANDS_KPOINTS_MAP = {
+    "fast": 0.1,
+    "moderate": 0.025,
+    "precise": 0.1,
+}
+
```
unkcpz, line 22: Why this appear here? I guess it is from https://github.com/aiidalab/aiidalab-qe/pull/446 but that PR is now empty?

src/aiidalab_qe/plugins/bands/setting.py
```diff
@@ -36,12 +36,23 @@ def __init__(self, **kwargs):
             disabled=False,
             style={"description_width": "initial"},
         )
-
+        self.two_dim_kpoints_path = ipw.Dropdown(
+            description="Select 2D symmetry path:",
```
unkcpz, line 40: I remember there is a paper about how to select kpath for 2D: https://www.sciencedirect.com/science/article/pii/S0927025610002697?via%3Dihub

Is the method from this paper? Anyway, I think we should add a detailed description of what this path is about and how it is created. I'll be better an `ipw.HTML` widget for it dedicately.

src/aiidalab_qe/plugins/bands/setting.py
```diff
@@ -36,12 +36,23 @@ def __init__(self, **kwargs):
             disabled=False,
             style={"description_width": "initial"},
         )
-
+        self.two_dim_kpoints_path = ipw.Dropdown(
```
unkcpz, line 39:
```suggestion
        self.kpath_2d = ipw.Dropdown(
```
I guess this is better?

src/aiidalab_qe/plugins/bands/workchain.py
```diff
@@ -1,8 +1,162 @@
...
+from aiida.plugins import DataFactory, WorkflowFactory
 from aiida_quantumespresso.common.types import ElectronicType, SpinType
 
 PwBandsWorkChain = WorkflowFactory("quantumespresso.pw.bands")
 
+KpointsData = DataFactory("core.array.kpoints")
+
+
+# function to set the band path for one and two dimensional systems
+def one_two_dim_kpoints_path(structure, kpoints_distance, two_dim_kpoints_path):
```
unkcpz, line 11: Why this is 1d or 2d but only gets `two_dim_kpoints_path` as input? Is that possible to have a function for 1D and function for 2d seperated?

unkcpz (changes requested):
This PR will fix https://github.com/aiidalab/aiidalab-qe/issues/211.

Thanks a lot for the nice work! I am looking forward to getting this new feature so we can get rid of the 2d dimension warning.

I have some minor requests, I think this is a feature that requires a clear description in UI, so that user will not get confused.

unkcpz:
> The test coverage is down, but don't worry. I have a draft PR to test the plugin. So, for now, it's fine to leave it

@superstar54 If possible, please add a new commit to this PR, it is better to have implementation and test together.

src/aiidalab_qe/app/result/report.py
AndresOrtegaGuerrero: This is because , since the kpoints for 1D and 2D are defined via kpoints instead of bands_kpoints_distance, so since that information wont be in the extras of the workchain, i can call it here to print in the report.

src/aiidalab_qe/plugins/bands/setting.py
AndresOrtegaGuerrero: @unkcpz as a reference i defined the 5 symmetries , like here https://pubs.acs.org/doi/10.1021/acs.jpclett.2c02972 , i will add a description

src/aiidalab_qe/plugins/bands/workchain.py
AndresOrtegaGuerrero: yes i can split this into two functions, to make it clear , I will modify the code to have two independent functions

src/aiidalab_qe/app/main.py
AndresOrtegaGuerrero: The issues is that i added the periodicity in this editor since a change in cell editor will require a PR to aiidalab-widget-base

src/aiidalab_qe/plugins/bands/setting.py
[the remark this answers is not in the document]
superstar54: For me, it's OK to keep it here for the moment. Then, maybe you can add an issue to update it when it is added in aiidalab widget base.

superstar54:
> @superstar54 If possible, please add a new commit to this PR, it is better to have implementation and test together.

I can do this. I need to first fix the problem in the draft PR #472.  @AndresOrtegaGuerrero Would you mind if I push some common this branch?

AndresOrtegaGuerrero:
> > @superstar54 If possible, please add a new commit to this PR, it is better to have implementation and test together.
> 
> I can do this. I need to first fix the problem in the draft PR #472. @AndresOrtegaGuerrero Would you mind if I push some commit to your PR?

Please go ahead, I already add the changes that @unkcpz  suggested. There is editor thing pending for discussion :D

src/aiidalab_qe/app/result/report.py
unkcpz: Okay, then we can close #446?

src/aiidalab_qe/app/main.py
unkcpz: No problem, I am fine with implement it here and open an issue on aiidalab-widgets-base. Please also add a comment here link to the issue you open so we remember to replace it in the future.

src/aiidalab_qe/plugins/bands/workchain.py
```diff
@@ -1,7 +1,179 @@
...
+            "labels": ["\u0393", "X", "H_1", "C", "H", "Y", "\u0393"],
+        }
+    points_branch = []
+    num_per_branch = []
+    path = selected_paths[two_dim_kpoints_path]["path"]
+    labels = selected_paths[two_dim_kpoints_path]["labels"]
+    branches = pairwise(path)
+
+    # Calculate the number of points per branch and generate the kpoints
+    for branch in branches:
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: you meant using for `for branch in zip(*[iter(path)*2]:` ?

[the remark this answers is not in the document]
AndresOrtegaGuerrero: I tried it doesnt work

src/aiidalab_qe/plugins/bands/workchain.py
```diff
@@ -1,7 +1,179 @@
...
+    )
+    # Generate the kpoints
+    points = np.linspace(
+        start=[0.0, 0.0, 0.0],
+        stop=[0.5, 0.0, 0.0],
+        endpoint=True,
+        num=num_points_per_branch,
+    )
+    kpoints.set_kpoints(points.tolist())
+    kpoints.labels = [[0, "\u0393"], [len(points) - 1, "X"]]
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: Will do!

src/aiidalab_qe/plugins/bands/workchain.py
```diff
@@ -1,7 +1,179 @@
...
+                num=num_points_per_branch,
+            )
+        else:
+            points = np.linspace(
+                start=branch[0], stop=branch[1], num=num_points_per_branch
+            )
+        points_branch.append(points.tolist())
+        num_per_branch.append(num_points_per_branch)
+
+    list_kpoints = [item for sublist in points_branch for item in sublist]
```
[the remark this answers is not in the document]
AndresOrtegaGuerrero: @unkcpz  you mean like this 
`list_kpoints = []
for sublist in points_branch:
    for item in sublist:
        list_kpoints.append(item)`

because basically since points_branch is a list of list , the goal of the list comprehension is to concatanate into a single list , should i split it into two for loops or i put some comments to explain this step ?

src/aiidalab_qe/app/result/report.py
AndresOrtegaGuerrero: No this issue is not related

unkcpz:
> is not related to https://github.com/aiidalab/aiidalab-qe/pull/446 , that one is for the nscf calculation

I see, sorry for the misunderstanding. I am curious why #446 has no code changes.

```
 PROTOCOL_BANDS_KPOINTS_MAP = {
    "fast": 0.1,
    "moderate": 0.025,
    "precise": 0.1,
}
```

Is this from aiida-quantumespresso protocol or it is defined by you? If it is from QE plugin, can you simply use the value from there, so in the future these values are synchronized with QE plugin?

src/aiidalab_qe/common/widgets.py
```diff
@@ -581,3 +605,15 @@ def _reset_all_tags(self, _=None):
...
+        """Select periodicity."""
+        periodicity_options = {
+            "xyz": (True, True, True),
+            "xy": (True, True, False),
+            "x": (True, False, False),
+        }
+        new_structure = deepcopy(self.structure)
+        new_structure.set_pbc(periodicity_options[self.periodicity.value])
+        self.structure = None
+        self.structure = deepcopy(new_structure)
```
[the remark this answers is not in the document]
unkcpz: Fair, thanks for the explanation! Let's keep deepcopy for safty.

src/aiidalab_qe/plugins/bands/workchain.py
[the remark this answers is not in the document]
unkcpz: I see, then what about:

```python
for branch in zip(path[:-1], path[1:]):
    print(branch)
```

I think this will do the trick and be easy to understand.

src/aiidalab_qe/plugins/bands/workchain.py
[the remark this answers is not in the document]
unkcpz: I see, this is flat a 2D list. Then the list comprehension is the best. Can you add a comment on it? Thanks!

src/aiidalab_qe/app/result/report.py
unkcpz: Very sad, I got a thumbs down 😂

AndresOrtegaGuerrero: sorry!  hahah

src/aiidalab_qe/plugins/bands/workchain.py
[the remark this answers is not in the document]
AndresOrtegaGuerrero: It did! thanks!

src/aiidalab_qe/app/result/report.py
```diff
@@ -57,7 +71,10 @@ def _generate_report_dict(builder_parameters: dict):
     yield "energy_cutoff_rho", builder_parameters.get("energy_cutoff_rho")
     yield "scf_kpoints_distance", builder_parameters.get("scf_kpoints_distance")
     yield "bands_kpoints_distance", builder_parameters.get(
-        "bands_kpoints_distance", None
+        "bands_kpoints_distance",
+        bands_kpoints_distance(
+            builder_parameters.get("run_bands"), builder_parameters.get("protocol")
+        ),
```
unkcpz, line 77: @superstar54 Can you check if, under the current design, the bands_kpoins_distance can be stored somewhere to restore the report back, rather than use a function?  

When I created this `_generate_report_dict` function, I planned that there would be only the final value to yield, the logic process needed to happen before the parameters passed in.

src/aiidalab_qe/app/result/report.py
```diff
@@ -70,6 +87,7 @@ def _generate_report_dict(builder_parameters: dict):
     yield "initial_magnetic_moments", builder_parameters.get(
         "initial_magnetic_moments", None
     )
+    yield "periodicity", builder_parameters.get("periodicity", "xyz")
```
unkcpz, line 90: Similar here, the default value `"xyz"` should be set as default value in `builder_parameters`.

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -2,28 +2,11 @@
...
 Authors: AiiDAlab team
 """
-import warnings
 
 import aiida
 import ipywidgets as ipw
 import traitlets
 from aiidalab_widgets_base import WizardAppWidgetStep
-from pymatgen.analysis.dimensionality import get_dimensionality_larsen
-from pymatgen.analysis.local_env import CrystalNN
```
unkcpz, line 12: I am very happy that we can potentially get rid of `pymatgen`.

src/aiidalab_qe/common/widgets.py
```diff
@@ -15,7 +15,9 @@
 import ipywidgets as ipw
 import numpy as np
 import traitlets
-from aiida.orm import CalcJobNode, load_node
+from aiida.orm import CalcJobNode
+from aiida.orm import Data as orm_Data
```
unkcpz, line 19: I'd use `from aiida import orm` and `orm.CalcJobNode` and `orm.Data`.

src/aiidalab_qe/common/widgets.py
```diff
@@ -430,6 +432,7 @@ class AddingTagsEditor(ipw.VBox):
     structure = traitlets.Instance(ase.Atoms, allow_none=True)
     selection = traitlets.List(traitlets.Int, allow_none=True)
     input_selection = traitlets.List(traitlets.Int, allow_none=True)
+    structure_node = traitlets.Instance(orm_Data, allow_none=True, read_only=True)
```
unkcpz, line 435:
```suggestion
    structure_node = traitlets.Instance(orm.StructureData, allow_none=True, read_only=True)
```
This is a structure data, no?

src/aiidalab_qe/plugins/bands/setting.py
```diff
@@ -36,12 +36,29 @@ def __init__(self, **kwargs):
             disabled=False,
             style={"description_width": "initial"},
         )
-
+        self.kpath_2d_help = ipw.HTML(
+            """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 5px">
+            If your system has periodicity xy. Please select one of the five 2D Bravais lattices corresponding to your system.
```
unkcpz, line 41:
```suggestion
            If the structure is two-dimensional, please select 2-dimensional Bravais lattices corresponding to the structure.
```

unkcpz (changes requested):
Thanks for updating the PR, just some very small further changing requests.

src/aiidalab_qe/app/result/report.py
unkcpz: But you can leave this one, and I'll have a quick discussion with @superstar54.
@AndresOrtegaGuerrero if you don't mind, I can directly add commits for this?

unkcpz:
> I update the PR i am waiting for Marnik for the review. 

Do you mean the bands kpoints distance protocol? @mbercx Can you have a look at it?

> I can do this. I need to first fix the problem in the draft PR https://github.com/aiidalab/aiidalab-qe/pull/472. @AndresOrtegaGuerrero Would you mind if I push some commit to your PR?

@superstar54, I give a go on #472, you can continue adding tests after that one is merged.

src/aiidalab_qe/app/result/report.py
AndresOrtegaGuerrero: Yes go ahead!

src/aiidalab_qe/app/result/report.py
AndresOrtegaGuerrero: that or we use a function to get the pbc from the StructureData

mbercx:
> @mbercx Can you have a look at it?

I'm confused. It this PR blocked by https://github.com/aiidalab/aiidalab-qe/pull/446? I left some comments there as I remember. I'm not sure if we have reached a conclusion, but I think I was leaning towards the following:

If the user wants to override the SCF k-points distance, we shouldn't ad-hoc override the NSCF k-points distance according to a somewhat arbitrary approach. It seems better to me that there is also an override for the NSCF k-points distance in this case. 

But I suppose this should be discussed on https://github.com/aiidalab/aiidalab-qe/pull/446, not sure why I was tagged for this here. ^^

unkcpz:
If I understand correctly, in this PR it defines a new "protocol" for kpoints distance, I think @AndresOrtegaGuerrero want you to have a look? 

I was confused about #446 because there is no code change, as I mentioned many times. See the screeshot, that's why I think @AndresOrtegaGuerrero move implementations to this PR. Can you double check you push your changes? 

image

Sorry if there is any misunderstanding. :pray:

superstar54:
@AndresOrtegaGuerrero @unkcpz I added the test for 1d and 2d bands. Now the codecov +0.22%.

I also deleted the old `path` and `npoint`, which are used for illustration only.

src/aiidalab_qe/app/result/report.py
```diff
@@ -14,6 +14,12 @@
     "precise": "SSSP/1.2/PBE/precision",
 }
 
+PROTOCOL_BANDS_KPOINTS_MAP = {
+    "fast": 0.1,
+    "moderate": 0.025,
+    "precise": 0.1,
```
superstar54, line 20:
```suggestion
    "precise": 0.01,
```

unkcpz:
I removed the logic in the report part, instead, the complex logic all happened when generating the builder parameters in the submission step. Please have a look at https://github.com/aiidalab/aiidalab-qe/pull/479/commits/b76bd70e840e54f9bfc5c584547a43344f05174d for moving "periodicity" and https://github.com/aiidalab/aiidalab-qe/pull/479/commits/358cd1d9624d2f629b7210654e1214dd135ce04c for moving the `bands_kpoints_distance`. In this way, we don't need to reclaim the protocol mapping because `bands_kpoints_distance` is stored in the builder. Actually, part of the code was implemented by @superstar54 already but comment out. I think this more or less conflicts with plugin design and remember you said you'll have a further PR to fix it, can you give it a look? Thanks.

superstar54:
> I removed the logic in the report part, instead, the complex logic all happened when generating the builder parameters in the submission step. Please have a look at [b76bd70](https://github.com/aiidalab/aiidalab-qe/commit/b76bd70e840e54f9bfc5c584547a43344f05174d) for moving "periodicity" and [358cd1d](https://github.com/aiidalab/aiidalab-qe/commit/358cd1d9624d2f629b7210654e1214dd135ce04c) for moving the `bands_kpoints_distance`. In this way, we don't need to reclaim the protocol mapping because `bands_kpoints_distance` is stored in the builder. Actually, part of the code was implemented by @superstar54 already but comment out. I think this more or less conflicts with plugin design and remember you said you'll have a further PR to fix it, can you give it a look? Thanks.

In  #481 , I refactor the report code, so that all the report-related parameters are handled in, and only in, the result step. and the `builder_parameters` is not needed anymore.

unkcpz:
So maybe we can merge this one first and then #481? I think this PR is all ready @AndresOrtegaGuerrero please have a look and we can merge.

unkcpz (approved):
I think this one is ready, and just had a quick chat with @superstar54 he is okay with all the changes.

Issue #418, opened 2023-06-21
Fuzzy misc enhancement requests


unkcpz:
I find some fuzzy UX problems during the use of the app, this issue is open as the centralized place to collect such problems:

- The workflow selection widget shows the description that the text is not clearly separated. 
- Start a new workflow and reset some widgets but not all. 
- Not super clear where to start a new workflow using workflow selector widget.
- Upload Structure widget keeps one increase in the counting when the new file is uploaded.

unkcpz:
Thanks @AndresOrtegaGuerrero, I'll convert these requests to the issues so we don't forget to fix it.

> (problem - is a calculation so it will create a node when you call it , right? )

Yes, it will create nodes, but I won't worry about this duplicate node. 
Very good suggestion, though.

> Once you do a geometry optimization, it could be useful to show in the results the initial structure (with different coloring) vs final optimized, it could by overlapping?. Her comment is that there should be a way to do the visual comparison

Not sure there is an easy way to do this. It is not clear to me if the cell and atom positions are both changed, and how to align the structures to compare.

superstar54:
> Not sure there is an easy way to do this. It is not clear to me if the cell and atom positions are both changed, and how to align the structures to compare.

An animation using the trajectory from the relaxing step is what we need.

unkcpz:
> An animation using the trajectory from the relaxing step is what we need.

I don't think this is easy to do. First, is there already widget that supports showing the trajectory? Second, if it is an animation, which viewer angle should we use to show it? I guess you may have a similar component in BeautifulAtom @superstar54 ?

superstar54:
`nglview` has an animation feature. We can test it. https://projects.volkamerlab.org/teachopencadd/talktorials/T017_advanced_nglview_usage.html
```
nv.show_asetraj(...)
```

AndresOrtegaGuerrero:
> Not sure there is an easy way to do this. It is not clear to me if the cell and atom positions are both changed, and how to align the structures to compare.

Perhaps when the user does a cell optimization, the report should show the optimized geometry and a comparison with the initial cell parameter and the final cell parameters. For a geometry optimization we can just use coordinates of the initial atom as dummy atoms and color them all in blue (to overlap with the final structure), and just write a button that display the initial ones (or hide). The download one, should just download the final structure.

superstar54:
> we actually have only two frames, the initial structure and the optimized one.

No, we can get the trajectory from the relax work chain.

[...]

PR #480, opened 2023-09-19
Adding HTML widget to show the kpoint setting changes

AndresOrtegaGuerrero:
This PR adds a HTML displaying the mesh that is going to be use in the calculations based on the kpoints_distances used. This was one issue in #418

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced.py
@@ -275,6 +275,7 @@ class AdvancedSettings(Panel):
             else:
                 self.mesh_grid.value = "Please select a number higher than 0.0"
 
+    # Taken from QE-pluggin - Modified to return a list and to not use Float or Bool from aiida-core
     def create_kpoints_from_distance(self, structure, distance, force_parity):
         """Generate a uniformly spaced kpoint mesh for a given structure.
 
--- a/src/aiidalab_qe/app/configuration/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced.py
@@ -70,12 +70,14 @@ class AdvancedSettings(Panel):
         )
 
         # Kpoints setting widget
-        self.kpoints_distance = ipw.FloatText(
+        self.kpoints_distance = ipw.BoundedFloatText(
+            min=0.0,
             step=0.05,
             description="K-points distance (1/Å):",
             disabled=False,
             style={"description_width": "initial"},
         )
+        self.mesh_grid = ipw.HTML()
         ipw.dlink(
             (self.override, "value"),
             (self.kpoints_distance, "disabled"),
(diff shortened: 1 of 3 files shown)


src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -261,6 +265,53 @@ def reset(self):
             # reset the override checkbox
             self.override.value = False
 
+    def _display_mesh(self, _=None):
+        if self.input_structure is not None:
+            if self.kpoints_distance.value > 0:
+                mesh = self.create_kpoints_from_distance(
```
unkcpz, line 271: I don't think you need a modified version of this function. It is a calcification in QE plugin and aiida now support passing regular data type and aiida will convert to aiida datatype for you. So here you can import the function from QE plugin, only need to take care of the output to the correct datatype.

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -261,6 +265,53 @@ def reset(self):
             # reset the override checkbox
             self.override.value = False
 
+    def _display_mesh(self, _=None):
+        if self.input_structure is not None:
```
unkcpz, line 269:
```suggestion
        if self.input_structure is None:
        return
```
Then the left part of code in this function has no need to be indented.

unkcpz (changes requested):
Thanks! I have two minor requests. Since we have a mature test framework and get more experience on writing unit test for widgets can you add a test for this class and test the function added here.

src/aiidalab_qe/app/configuration/advanced.py
AndresOrtegaGuerrero: @unkcpz Thank you i will adapt the code and write the test

AndresOrtegaGuerrero:
@unkcpz  Do you have any suggestion on why the test is failing? , locally the advanced test is passing

unkcpz:
I think that is because you call calcfunction but you don’t have rabbitmq. Please try adding this section https://github.com/aiidalab/aiidalab-widgets-base/blob/2c29a9eac70e9f0773fd93165daccc14c304919b/.github/workflows/ci.yml#L113

tests/configuration/test_advanced.py
```diff
@@ -114,3 +114,40 @@ def test_advanced_tot_charge_settings():
...
+
+    from aiidalab_qe.app.configuration.advanced import AdvancedSettings
+
+    StructureData = DataFactory("structure")
+
+    w = AdvancedSettings()
+
+    # Create a StructureData for AdvancedSettings (Silicon)
+
+    structure = StructureData(
```
unkcpz, line 131:
```suggestion
    structure = orm.StructureData(
```
We usually have this in other packages.

unkcpz (changes requested):
Just one minor request, all good then.

tests/configuration/test_advanced.py
AndresOrtegaGuerrero: no problems i will call the StructureData via from aiida import orm,

unkcpz (approved):
Thanks! all great. @AndresOrtegaGuerrero

PR #481, opened 2023-09-20
Move report

superstar54:
Move the report part from the `submission` step to the `result` step, so that all the report-related parameters are handled in, and only in, the result step.

The old design separates the report into two parts in the `submission` and `result` steps, which is hard to maintain. One also needs to save and reload the parameters. Which is a mess, as seen from the names of the functions.
- submission step `_create_extra_report_parameters`, `_extract_report_parameters`, 
- result step: `_generate_report_dict`, `_generate_report_html`, `generate_report_html`

It also extracts parameters from the GUI and the builder to form a `builder_parameters`, but later one needs to re-extract the GUI parameters back from the `builder_parameters` (handle the logic again). Most importantly, we can not reload all GUI parameters, which is bad.

The current implementation:
==========================
- move all report-related code to the `result` step, the `summary_viewer.py` file.
- save all GUI-related data to the `ui_parameters`, and set it as an extra attribute to the node, so we can reload the GUI back.
- generate report parameters from the `ui_parameters` and inputs of the workchain node directly. 

Test
======
- add unit tests for the `result` step, including a regression test for the report parameters. +5 % now! :grinning:


unkcpz (approved):
LGTM, thanks!

PR #485, opened 2023-09-22
Reload GUI parameters

superstar54:
This PR allows the QeApp to reload GUI parameters when the user selects a process.

- use `aiida.orm.utils.serialize.serialize` to serialize the GUI parameters as yaml, then save it as an extra of the process node. Then load it using the `deserialize_unsafe` method.
- fix set_panel_value for the `advanced_settings`.
- add unit test for reloading the app.

### DIFF
--- a/src/aiidalab_qe/app/submission/__init__.py
+++ b/src/aiidalab_qe/app/submission/__init__.py
@@ -80,7 +80,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
         self.resources_config = ResourceSelectionWidget()
         self.parallelization = ParallelizationSettings()
 
-        self.set_selected_codes(DEFAULT_PARAMETERS)
+        self.set_selected_codes(DEFAULT_PARAMETERS["codes"])
         self.set_resource_defaults()
 
         self.pw_code.observe(self._update_state, "value")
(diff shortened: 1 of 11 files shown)


AndresOrtegaGuerrero:
@superstar54 , I try using the branch, but i cant upload a previous job it produces an error . On the other hand, i tried creating a new job and setting different parameters, this one works!. With the exception that the Protocol, the Smearing width and Smearing Type were not re-load

AndresOrtegaGuerrero:
I guess in order to not affect previous jobs maybe you just can check if the yaml file is reachable if not, just set the GUI parameters as default ?

superstar54:
@AndresOrtegaGuerrero Thanks! I fixed the `protocol` and `smearing`.

> I guess in order to not affect previous jobs maybe you just can check if the yaml file is reachable if not, just set the GUI parameters as default ?

I added a check if the `ui_parameters` is a yaml format or not. Note that since there is an update of the summary report (#481), it also does not show the summary for the old report.

AndresOrtegaGuerrero (approved):
LGTM!

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -471,6 +472,16 @@ def _update_builder(self, buildy, max_mpi_per_pool):
                 else:
                     self._update_builder(v, max_mpi_per_pool)
 
+    def set_submission_parameters(self, parameters):
```
unkcpz, line 475: @superstar54, can you add docstring for these newly added functions? I find it quite hard to understand when I first look at it. 
Also please try to add as many comment/docstring as possible. Sometimes the design is clear for you but may not be easy to understand for other developers.

PR #487, opened 2023-09-23
add code entry point

superstar54:
This PR adds an entry_point for code in the submission step, allowing users to add code for their plugin. Now, only the `pw` code is hardcoded. Other codes, e.g. `dos` and `projwfc`, are from the plugins.

If the workchain property is not selected, the related code will be hidden.

Here is the example of the pdos plugins with codes: `dos` and `projwfc`:

```python

dos_code = ComputationalResourcesWidget(
    description="dos.x",
    default_calc_job_plugin="quantumespresso.dos",
)

projwfc_code = ComputationalResourcesWidget(
    description="projwfc.x",
    default_calc_job_plugin="quantumespresso.projwfc",
)


pdos = {
    "outline": PdosOutline,
    "code": {"dos": dos_code, "projwfc": projwfc_code},
    "setting": Setting,
    "result": Result,
    "workchain": workchain_and_builder,
}
```


mikibonacci:
Thank Xing, this is super nice! 

I have few comments:
1 - do you think we can easily integrate some automatic code installation (if the specific version/instruction on how to do it is provided in the plugin setup specification)
2 - code-wise resource & parallelization widgets? maybe we can integrate ResourceSelectionWidget and ParallelizationSettings in the ComputationalResourcesWidget, so it is all related to the specific code? 

This may produce conflict if >1 plugins have the same codes, with different definitions. But maybe these will be very specific cases that we can solve ad hoc. 

what do you think? I can also work on comment 2. comment 1 is not so important I think, at least for now

PNOGillespie:
Just to add my perspective, having tested this system for developing a new plugin, the process to do so is now a lot easier in terms of what one has to do in order to incorporate a new plugin into the rest of the App - so, very nicely done.

No other comments really, I think it just needs some brief documentation for developers to be added at a later date - essentially just to clarify how to add a new plugin to the App.

superstar54:
@mikibonacci @PNOGillespie Thanks for the test.

> do you think we can easily integrate some automatic code installation (if the specific version/instruction on how to do it is provided in the plugin setup specification)

This is code dependent. For code that can be installed using `conda/pip`, it's easy to integrate the automatic code installation. Please check the `setup_codes.py` file.

> code-wise resource & parallelization widgets? maybe we can integrate ResourceSelectionWidget and ParallelizationSettings in the ComputationalResourcesWidget, so it is all related to the specific code?
> I can also work on comment 2. comment 1 is not so important I think, at least for now

We had such a discussion before, about adding different resources for `pw`, `dos`, etc. You can mention this in the AiiDAlab meeting, and good if you have time to work out a prototype. 


> I think it just needs some brief documentation for developers to be added at a later date

Yes, a clear doc will be added.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -68,26 +69,29 @@ def __init__(self, qe_auto_setup=True, **kwargs):
...
+            self.codes_help,
+            self.pw_code,
+        ]
+        self.code_entries = get_entry_items("aiidalab_qe.properties", "code")
+        for _, entry_point in self.code_entries.items():
+            for name, code in entry_point.items():
+                self.codes[name] = code
+                self.code_children.append(self.codes[name])
+        self.set_selected_codes(DEFAULT_PARAMETERS["codes"])
+        # s
```
unkcpz, line 94:
```suggestion
```

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -380,18 +338,22 @@ def _get_code_uuid(code):
...
-            self.projwfc_code.code_select_dropdown.disabled = False
-        else:
-            self.dos_code.code_select_dropdown.disabled = True
-            self.projwfc_code.code_select_dropdown.disabled = True
+            for name, code in self.codes.items():
+                code.value = _get_code_uuid(codes.get(name))
+
+    def set_codes_status(self):
+        """Disable code if no related property is selected."""
+        # disable all codes except pw
```
unkcpz, line 346: Instead of disable, maybe we hide? set visible -> "none"? But this is a layout-related comment, I don't want to block the PR with it. Please ignore if you don't agree.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -380,18 +338,22 @@ def _get_code_uuid(code):
...
+        for name, code in self.codes.items():
+            if name == "pw":
+                continue
```
unkcpz, line 349: I know the `pw` is essential code should always be there. So why not separated it as a new widget and only leave the entry point implementation for plug-in codes?

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -380,18 +338,22 @@ def _get_code_uuid(code):
...
+            code.code_select_dropdown.disabled = True
+        # enable code if the related property is selected.
+        properties = self.input_parameters.get("workchain", {}).get("properties", [])
+        for identifer, codes in self.code_entries.items():
```
unkcpz, line 353:
```suggestion
        # Update the layout of code widgets
        for identifer, codes in self.code_entries.items():
```
Another nitpick, possible to simplify this for..if..for ?

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -4,12 +4,40 @@
 PdosWorkChain = WorkflowFactory("quantumespresso.pdos")
 
 
+def check_codes(pw_code, dos_code, projwfc_code):
+    """Check that the codes are installed on the same computer."""
+    if (
+        not any(
```
unkcpz, line 10: Cool, much clear that this check is moved to here for pdos only.

unkcpz (changes requested):
Some minor requests.

src/aiidalab_qe/app/submission/__init__.py
superstar54: It is convenient to manipulate the code widgets, e.g. `set` and `get` the code widget values, if all widgets are stored together, and treated equally.

src/aiidalab_qe/app/submission/__init__.py
superstar54: I removed the `if`.

src/aiidalab_qe/app/submission/__init__.py
superstar54: Good idea. I used the visibility.

unkcpz (approved):
Thanks @superstar54. All good to me. Ready to :ship:

Issue #437, opened 2023-08-16
Create new workflow using workflow selector reset the functional but not update widget


unkcpz:
**Describe the bug**
When the 1st calculation is running by selecting PBE, to start the second one using "New workflow" by workflow selector, the PBEsol is used but the pseudo family selector widget shows "PBE".

This is caused by the widget is not properly reset.

superstar54:
fixed in #489

Issue #420, opened 2023-06-21
Start a new workflow and reset some widgets but not all.


unkcpz:
1. Open a qe app in a new page
2. start a new calculation, and set "Structure as it is" (without relaxation), and override pseudopotential to SSSP precision.
3. using "New workflow" of workflow selector to start a new workflow
4. You'll find some widget is reset to default such as structure set to "Full geometry", but pseudopotential is still overrided.

I encounter this with `v23.04.3`, not sure if I fix some already during refactoring.

superstar54:
fixed in #489

PR #489, opened 2023-09-26
Reset GUI

superstar54:
Reset the GUI when selecting to start a new workflow. Related to  #437 #420. 

Reset:
- step 1, Structure (structure and viewer)
- step 2, Configuration (all panels)
- step 3, computational resource. (code and resources)

Step 4 will be automatically reset when the `process` of step 3 is reset.

### DIFF
--- a/src/aiidalab_qe/app/configuration/pseudos.py
+++ b/src/aiidalab_qe/app/configuration/pseudos.py
@@ -214,7 +214,10 @@ class PseudoSetter(ipw.VBox):
     ecutwfc = tl.Float()
     ecutrho = tl.Float()
 
-    _pseudo_setter_helper_text = """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 10px">
+    _default_pseudo_setter_helper_text = """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 10px; color: red;">
+            Input structure is not set. Please set the structure first.
+            </div>"""
+    _update_pseudo_setter_helper_text = """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 10px">
         The pseudopotential for each kind of atom in the structure can be set customly.
         The default pseudopotential and cutoffs are get from the pseudo family.
         The cutoffs used for the calculation are the maximum of the default from all pseudopotentials
@@ -238,11 +241,7 @@ class PseudoSetter(ipw.VBox):
         self.ecutwfc = 0
         self.ecutrho = 0
 
-        self.pseudo_setter_helper = ipw.HTML(
-            """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 10px; color: red;">
-            Input structure is not set. Please set the structure first.
-            </div>"""
-        )
+        self.pseudo_setter_helper = ipw.HTML(self._default_pseudo_setter_helper_text)
         self.cutoff_setter_helper = ipw.HTML(self._cutoff_setter_helper_text)
         self.ecutwfc_setter = ipw.FloatText(
             description="Wavefunction cutoff (Ry)",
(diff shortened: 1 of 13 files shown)


AndresOrtegaGuerrero:
@superstar54  I did a test, and it seems the Smearing type: , Smearing width (Ry) arent reset , also Wavefunction cutoff (Ry)
Charge density cutoff (Ry) . I am also wondering if the options in the panels get also reset ? I guess each panel needs a method named reset ?

superstar54:
Thanks for the review. 

>  it seems the Smearing type: , Smearing width (Ry) arent reset , also Wavefunction cutoff (Ry)
Charge density cutoff (Ry) . 

I fixed it. Also fixed the `reset` in `pseudo_setter` and `magnetization`.

> I am also wondering if the options in the panels get also reset ? I guess each panel needs a method named reset ?

yes, each panel needs to implement a `reset` method, which will be called from the `ConfigureQeAppWorkChainStep`.

AndresOrtegaGuerrero:
@superstar54 Would it be necessary to add in the documentation of qe_app in `create your plugin` , to include a reset method, to avoid future conflicts ?

AndresOrtegaGuerrero:
@superstar54 I tested the PR, now it does update the field i told you , yet I noticed some behavior. I started to select one workchain and then another , and then the first one again. (Everything was working) after another change the app crashed.
Screenshot 2023-10-10 at 09 47 23

AndresOrtegaGuerrero:
@superstar54 another thing i noticed is that one of my workchain , (newer one) doesnt set the input parameters, but the another one i was using for testing it does. I know this is related to Load GUI PR.

AndresOrtegaGuerrero:
>Which workchain? In my test, the load_GUI is OK.

It seems my workchain is "old" so i tried with a new one an no issues , maybe that workchain is incompatible with the load_GUI, but indeed i tried a new one and everything upload correctly

src/aiidalab_qe/app/configuration/pseudos.py
```diff
@@ -278,7 +279,7 @@ def _on_cutoff_change(self, _=None):
         self.ecutrho = self.ecutrho_setter.value
 
     def _reset_cutoff_widgets(self):
-        self.ecutrho_setter.value = 0
+        self.ecutwfc_setter.value = 0
         self.ecutrho_setter.value = 0
 
     def _reset_traitlets(self):
```
AndresOrtegaGuerrero, line 285: Maybe you should include docstring to these _reset_* functions , just to keep it consistent with the rest of your changes where you added docstring

superstar54:
Hi @AndresOrtegaGuerrero , thanks for the review. I made the changes as you requested.
- add docstring to the `reset` methods.
- remove the message of `Stored in AiiDA....`

AndresOrtegaGuerrero (approved):
Thank you! LGTM. One note once you use .reset() for the settings in the panels, if a future plugin doesnt have this method it should be a problem right? I think that just be a different issue for the documentation for plugin developers

superstar54:
> One note once you use .reset() for the settings in the panels, if a future plugin doesnt have this method it should be a problem right? I think that just be a different issue for the documentation for plugin developers

Thanks for the comment. In the `Panel` class, there is an empty `reset` method to ensure it always exists.

unkcpz:
Thanks @superstar54. This is finally happened, great!

> In the Panel class, there is an empty reset method to ensure it always exists.

However, there is no guarantee that user (property plugin developers) will implement this. I think if the `reset` is the method that must be overridden otherwise the app is not working properly, you need make the `reset` of `Panel` raise `NotImplement` exception or provide the default behaviour to avoid breaking the app. 
If this is the case, please open an issue for future implementation. Thanks!

PR #494, opened 2023-10-04
Add button to start new workflow

unkcpz:
Looks like this:

image

### DIFF
--- a/src/aiidalab_qe/common/process.py
+++ b/src/aiidalab_qe/common/process.py
@@ -17,7 +17,7 @@ class WorkChainSelector(ipw.HBox):
     names of the output keys of the `parse_extra_info` method.
     """
 
-    # The PK of a 'aiida.workflows:quantumespresso.pw.bands' WorkChainNode.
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/common/process.py
```diff
@@ -55,13 +55,29 @@ def __init__(self, process_label, **kwargs):
         else:
             self.fmt_workchain = self.BASE_FMT_WORKCHAIN
 
-        self.refresh_work_chains_button = ipw.Button(description="Refresh")
+        self.new_work_chains_button = ipw.Button(
+            description="New",
+            tooltip="Start a new computation",
+            button_style="primary",
```
superstar54, line 61: I would suggest using different colors for the buttons `New` and `Refresh`.
```suggestion
            button_style="success",
```

superstar54 (approved):
Thanks for the fast work. LGTM!

I added one suggestion for the color. You can decide which color you want to use. I approved the PR. There is no need for a second review.

PR #497, opened 2023-10-06
Add clean-up button to control the remote work directory clean

### LINKED ISSUE #370: Clean the remote workdir as default and provide checkbox for not clean up
In the THEOS testing server, the storage space of active user increased dramatically. The issue comes from that the calculation folder is not cleaned up if run in localhost. 

Clean up the remote workdir if the calculation finished successfully, provide the widget to tick the option to turn off clean up.

unkcpz:
fixes #370

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced.py
@@ -54,7 +54,7 @@ class AdvancedSettings(Panel):
             description="",
             indent=False,
             value=True,
-            layout=ipw.Layout(max_width="10%"),
+            layout=ipw.Layout(max_width="20px"),
         )
         self.clean_workdir_description = ipw.HTML(
             """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 5px">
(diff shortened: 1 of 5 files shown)


mikibonacci (commented):
Hi jason, is it intended to leave some space between the checkbox and the description?


Screenshot from 2023-10-09 11-41-17

mikibonacci (approved):

unkcpz:
> Hi jason, is it intended to leave some space between the checkbox and the description?

Good point, I didn't expect the space are so large. I intended to use the `px` instead of the `%`. I polished it a bit. We need anyway go though all the layout and make everything look compact, but that is a task for the future.

AndresOrtegaGuerrero:
@mikibonacci By default the check box is always on ? This could be problematic if an user missed this and wanted to do postprocessing with some files not saved

unkcpz:
> This could be problematic if an user missed this and wanted to do postprocessing with some files not saved

Then, it will be an "advanced" user and he is supposed to go to the advance setting and un-tick it I assume?

unkcpz:
As we discussed in the last AiiDAlab meeting, we agree that the workchain should not be cleaned by default. I'll open a PR for it after.

PR #500, opened 2023-10-09
Feature/kill button - adding the kill button on top of NodeTree 

### LINKED ISSUE #161: Add button to kill the job
I'm frustrated when I start a calculation, but I realise I made an error and I want to kill it, and I don't know how to do it (well, I know that I can open the shell and do `verdi process kill` after finding the PK, but an average user does not know this).

Have a "Kill calculation" red button that shows somewhere while the calculation is running. This should also ask for confirmation to avoid a click by mistake.

mikibonacci:
Fixes #161 

adding the kill button to support live-killing of a workchain, just selecting it from the list and clicking on the button.

### DIFF
--- a/src/aiidalab_qe/common/process.py
+++ b/src/aiidalab_qe/common/process.py
@@ -86,7 +86,9 @@ class WorkChainSelector(ipw.HBox):
         ipw.dlink(
             (self, "value"),
             (self.kill_work_chains_button, "disabled"),
-            lambda workchain: orm.load_node(workchain).is_terminated if isinstance(workchain, int) else True,
+            lambda workchain: orm.load_node(workchain).is_terminated
+            if isinstance(workchain, int)
+            else True,
         )
         self.kill_work_chains_button.on_click(self._on_click_kill_work_chain)
 
@@ -144,7 +146,6 @@ class WorkChainSelector(ipw.HBox):
             child.disabled = change["new"]
 
     def refresh_work_chains(self, _=None):
-        
         try:
             self.set_trait("busy", True)  # disables the widget
 
@@ -164,9 +165,12 @@ class WorkChainSelector(ipw.HBox):
                 self.work_chains_selector.value = original_value
         finally:
             self.set_trait("busy", False)  # reenable the widget
-            
-        self.kill_work_chains_button.disabled = orm.load_node(self.value).is_terminated if isinstance(self.value, int) else True
 
+        self.kill_work_chains_button.disabled = (
+            orm.load_node(self.value).is_terminated
+            if isinstance(self.value, int)
+            else True
+        )
 
     def _on_click_new_work_chain(self, _=None):
         self.refresh_work_chains()
--- a/src/aiidalab_qe/common/process.py
+++ b/src/aiidalab_qe/common/process.py
@@ -152,7 +152,6 @@ class WorkChainSelector(ipw.HBox):
         self.refresh_work_chains()
         self.work_chains_selector.value = self._NO_PROCESS
 
-
     @tl.observe("value")
     def _observe_value(self, change):
         if change["old"] == change["new"]:
--- a/src/aiidalab_qe/common/process.py
+++ b/src/aiidalab_qe/common/process.py
@@ -80,15 +80,8 @@ class WorkChainSelector(ipw.HBox):
             tooltip="Kill the selected workflow",
             button_style="danger",
             icon="window-close",
-            disabled=False,
+            disabled=True,
             layout=ipw.Layout(width="auto"),
-        )
-        ipw.dlink(
-            (self, "value"),
-            (self.kill_work_chains_button, "disabled"),
-            lambda workchain: orm.load_node(workchain).is_terminated
-            if isinstance(workchain, int)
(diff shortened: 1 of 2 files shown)


src/aiidalab_qe/common/process.py
```diff
@@ -150,6 +162,12 @@ def _on_click_new_work_chain(self, _=None):
         self.refresh_work_chains()
         self.work_chains_selector.value = self._NO_PROCESS
 
+    def _on_click_kill_work_chain(self, _=None):
+        if isinstance(self.work_chains_selector.value, int):
```
unkcpz, line 166: I guess this condition is to avoid click when no workchain is selected ("No process")?  If so, maybe it is more proper to disable the kill button so user can not click on it.

src/aiidalab_qe/common/process.py
```diff
@@ -150,6 +162,12 @@ def _on_click_new_work_chain(self, _=None):
...
+            self.refresh_work_chains()
+            processes = [orm.load_node(self.work_chains_selector.value)]
+            control.kill_processes(processes)
```
unkcpz, line 169:
```suggestion
            message = 'Killed through `AiiDAlab app interface`'
            control.kill_processes(processes, message=message)
```

unkcpz (changes requested):
Thanks @mikibonacci. Just two small comments.

unkcpz:
> adding the kill button to support live-killing of a workchain, just selecting it from the list and clicking on the button.

I think for this button, it should be only enabled only when the workchain is in "running" state, so the state of the button should be update when the new process is selected from the workchain selector list.

danielhollas:
Cool, I might steal this, wanted to have this feature for a while :-) Hower, my plan was to implement the button next somewhere next to the NodeTree widget of a running process. It seems a bit out of place in the Workchain selector, I would worry about accidentally clicking it. WDYT?

unkcpz:
> It seems a bit out of place in the Workchain selector, I would worry about accidentally clicking it. WDYT?

@danielhollas thanks a lot for chime in. It is a great point. My plan was to bring in this concept to the QeApp and we think about the layout afterward. 
For the misclick, I won't worry about it too much, the `window-close` icon contains the cross, so I suppose the user will notice it will cause some fatal consequence (although we can not predict too much but how fatal it can be?).  You idea of put it in the node tree is great, but I am afraid it is straightforward to implement especially when we have some issues concerned about improve the nodetree in general.

danielhollas:
Just to clarify, I did not mean to implement it directly in the Node tree widget, just as a simple standalone button somewhere next to it. You also highlight an important point, having just an icon without any explanation is very cryptic, whereas an big red scary "Kill workflow" button would be more obvious.

mikibonacci:
Dear @danielhollas  thank you for your feedback! indeed, I think that the button should be somewhere near the workchain tree. Maybe for this release we can still have the button near the other two (new and refresh), and then in the future put it in a more reasonable place?

src/aiidalab_qe/common/process.py
```diff
@@ -73,12 +75,30 @@ def __init__(self, process_label, **kwargs):
...
+            disabled=False,
+            layout=ipw.Layout(width="auto"),
+        )
+        ipw.dlink(
+            (self, "value"),
+            (self.kill_work_chains_button, "disabled"),
+            lambda workchain: orm.load_node(workchain).is_terminated
+            if isinstance(workchain, int)
+            else True,
+        )
```
unkcpz, line 92: Instead of using `dlink`, I think it is more proper to set the disable of the button in the callback function `_observe_value`. Because:
1. the `dlink` is a bit complex with the transform from uuid to the running status, which will be clear if it is a regular code block in the function.
2. In the `refresh_work_chain` method, the disable of the button is check and set, this has the same logic as `dlink` here and the refresh should trigger the callback (it is already the case I assume, if not we need to fix it).

unkcpz (changes requested):
One more minor thing to change, others are all good to me.

unkcpz:
> Just to clarify, I did not mean to implement it directly in the Node tree widget, just as a simple standalone button somewhere next to it.

I see, sorry for misunderstanding it. I guess you mean add it here, https://github.com/aiidalab/aiidalab-qe/blob/967c3f8d5a4347b69e9b2a0093e94be6bef26695/src/aiidalab_qe/app/result/__init__.py#L25-L30

I can see the advantage of putting the button here, but I also thing having the kill button with "new", "refresh" is compatible with its usage. Let's discuss this in the meeting and see what others think.

unkcpz:
Hi @mikibonacci, as we discussed during the meeting, we agree on that this button better to be placed above the node tree. Can you move it to there, thanks.

unkcpz (approved):
We had an offline pair coding and tidying up a bit on the implementation. All good from me. Thanks a lot! @mikibonacci

danielhollas:
Nice, can you post a screenshot of the final product? When thinking about
this, I couldn't figure out where is the best place to put it.

Dne čt 19. 10. 2023 17:29 uživatel Jusong Yu ***@***.***>
napsal:

PR #502, opened 2023-10-09
add entry point to structure_manager_widget

superstar54:
Continue the plugin implementation. This PR adds an entry point for step 1,  allowing the user to add a new structure importer and editor.

In principle, all the structure `importer` and `editor` are pluggable. This PR just generalizes this by adding entry point to it.

This will be very useful. 
- add structure importer specific for particular structures, e.g. surface,  adsorbate.
- add a new `editor` to edit a structure for the plugin, e.g. edit tags, and cut surface. Like the `AddingTagsEditor` by @AndresOrtegaGuerrero


AndresOrtegaGuerrero:
@superstar54 I noticed with this PR , I can reach my StructureData, it says i have 0.

AndresOrtegaGuerrero:
@superstar54 I guess we need a sample pluggin to test this PR right ? in such a way that the pluggin includes an "extra" step 1 ?

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -22,15 +34,43 @@
     ("Cobalt (hcp)", file_path / "examples" / "Co.cif"),
 ]
 
+OptimadeQueryWidget.title = "OPTIMADE"  # monkeypatch
+
 
 class StructureSelectionStep(ipw.VBox, WizardAppWidgetStep):
     """Integrated widget for the selection of structures from different sources."""
```
AndresOrtegaGuerrero, line 41: I feel the Docstring in this class should be updated to incorporate the structure of the class, and also a description of modification of the importer and editor

AndresOrtegaGuerrero:
@superstar54 For example in the example for testing if I have 
class EosOutline(OutlinePanel):
    title = "Equation of State (EOS)"


eos ={
"outline": EosOutline,
"importer": EOSStructure,
"setting": Setting,
"workchain": workchain_and_builder,
"result": Result,
}

i can just declare something like  "editor": EOSEditor, should it work ?

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -22,15 +34,43 @@
...
+            importers.append(entry_point())
+        #
+        editors = [
+            BasicCellEditor(title="Edit cell"),
+            BasicStructureEditor(title="Edit structure"),
+            AddingTagsEditor(title="Edit StructureData"),
+        ]
+        entries = get_entry_items("aiidalab_qe.properties", "editor")
+        for _, entry_point in entries.items():
+            importers.append(entry_point())
```
AndresOrtegaGuerrero, line 65: should it be here `editors.append(entry_point())`

src/aiidalab_qe/app/structure/__init__.py
superstar54: Thanks, I have updated it.

src/aiidalab_qe/app/structure/__init__.py
superstar54: Yes, I changed it. thanks!

superstar54:
Hi @AndresOrtegaGuerrero , thanks for the review. I have made the changes as you suggested.

> i can just declare something like "editor": EOSEditor, should it work ?

Yes, you can add the editor like this. I have update the `aiidalab_qe_eos`, you can test the `importer` and `editor` use it.

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -22,15 +34,47 @@
...
+        entries = get_entry_items("aiidalab_qe.properties", "importer")
+        for _, entry_point in entries.items():
+            importers.append(entry_point())
+        # add plugin specific structure editors
+        editors = [
+            BasicCellEditor(title="Edit cell"),
+            BasicStructureEditor(title="Edit structure"),
+            AddingTagsEditor(title="Edit StructureData"),
+        ]
+        entries = get_entry_items("aiidalab_qe.properties", "editor")
```
AndresOrtegaGuerrero, line 67: what about list comprehension like `importers.extend(entry_point() for entry_point in get_entry_items("aiidalab_qe.properties", "importer").values())` , this is not important just wondering :)

AndresOrtegaGuerrero:
@superstar54 Thank you again! I tried the PR it works, but it seems some of the test didnt pass

superstar54:
@AndresOrtegaGuerrero Thanks. I fixed it. All tests pass now.

AndresOrtegaGuerrero (approved):
LGTM! Check my comment on the list comprehension , in case you think is useful

src/aiidalab_qe/app/structure/__init__.py
superstar54: thanks for the suggestion. I used the list comprehension.

PR #505, opened 2023-10-13
check if code is none in workchain

### LINKED ISSUE #506: Code is None in workchain
**Describe the bug**
From Peter.
If I start a new workflow using the Pdos WorkChain and (optionally) other WorkChains, everything works fine. If instead I start one with only the Bands or Xps WorkChains (or both) the App crashes with the attached traceback.
If I select the Pdos and Bands/Xps WorkChains together,  set the dos.x and projwfc.x codes, then cancel the Pdos WorkChain, the workflow runs without issues - skipping the Pdos step.

**traceback**
```
~/apps/aiidalab-qe/src/aiidalab_qe/app/submission/__init__.py in _on_submit_button_clicked(self, _)
    355     def _on_submit_button_clicked(self, _):
    356         self.submit_button.disabled = True
--> 357         self.submit()
    358 
    359     def get_selected_codes(self):
~/apps/aiidalab-qe/src/aiidalab_qe/app/submission/__init__.py in submit(self, _)
    398         from aiida.orm.utils.serialize import serialize
    399 
--> 400         builder = self._create_builder()
    401 
    402         with self.hold_trait_notifications():
~/apps/aiidalab-qe/src/aiidalab_qe/app/submission/__init__.py in _create_builder(self)
    438         # add codes and resource info into ui_parameters
    439         self.ui_parameters.update(self.get_submission_parameters())
--> 440         builder = QeAppWorkChain.get_builder_from_protocol(
    441             structure=self.input_structure,
    442             parameters=deepcopy(self.ui_parameters),
~/apps/aiidalab-qe/src/aiidalab_qe/workflows/__init__.py in get_builder_from_protocol(cls, structure, parameters, clean_workdir, **kwargs)
    121         properties = parameters["workchain"].pop("properties", [])
    122         codes = parameters.pop("codes", {})
--> 123         codes = {key: orm.load_node(value) for key, value in codes.items()}
    124         # update pseudos
    125         for kind, uuid in parameters["advanced"]["pw"]["pseudos"].items():
~/apps/aiidalab-qe/src/aiidalab_qe/workflows/__init__.py in <dictcomp>(.0)
    121         properties = parameters["workchain"].pop("properties", [])
    122         codes = parameters.pop("codes", {})
--> 123         codes = {key: orm.load_node(value) for key, value in codes.items()}
    124         # update pseudos
    125         for kind, uuid in parameters["advanced"]["pw"]["pseudos"].items():
/opt/conda/lib/python3.9/site-packages/aiida/orm/utils/loaders.py in load_node(identifier, pk, uuid, label, sub_classes, query_with_dashes)
    204     :raise aiida.common.MultipleObjectsError: if more than one Node was found
    205     """
--> 206     return load_entity(
    207         NodeEntityLoader,
    208         identifier=identifier,
/opt/conda/lib/python3.9/site-packages/aiida/orm/utils/loaders.py in load_entity(entity_loader, identifier, pk, uuid, label, sub_classes, query_with_dashes)
     55 
     56     if inputs_provided == 0:
---> 57         raise ValueError("one of the parameters 'identifier', pk', 'uuid' or 'label' has to be specified")
     58     elif inputs_provided > 1:
     59         raise ValueError("only one of parameters 'identifier', pk', 'uuid' or 'label' has to be specified")
ValueError: one of the parameters 'identifier', pk', 'uuid' or 'label' has to be specified
```

superstar54:
fix #506 
If no code is not selected, it is `None`. So, in the workchain, we avoid loading it.


unkcpz (changes requested):
Thanks @superstar54, I guess the issue only happens if there are other code, e.g xps involved in the calculation, correct? 
The change looks all good for me, can you add a unit test to it?

superstar54:
> I guess the issue only happens if there are other code, e.g xps involved in the calculation, correct?

For all codes. e.g if not calculate `pdos`, and codes `dos` is not selected. The `dos` code is passed as `None` to the builder.

superstar54:
> The change looks all good for me, can you add a unit test to it?

Thanks for the suggestion. I have added the test.

tests/test_submit_qe_workchain.py
```diff
@@ -1,3 +1,10 @@
+def test_code_not_selected(submit_app_generator):
+    """Test if there is an error when the code is not selected."""
+    app = submit_app_generator()
+    app.submit_step.dos_code.value = None
+    app.submit_step._create_builder()
```
unkcpz, line 5: Thanks for adding the test. But I didn’t see what you were testing for. Maybe add an assert statement, what is the value of dos code in the builder? 
I also don’t understand why dos code will be None, the drop-down widget only has its value equal to None right after initialization but we already have localhost code for it as default. 

But I agree the change is needed for the case such as xps code.

PNOGillespie: > I also don’t understand why dos code will be None, the drop-down widget only has its value equal to None right after initialization but we already have localhost code for it as default.

If I may add something here (as the person that found this bug originally): I confirm that Xing's fix does work, though I agree that the nature of the issue is strange. From my own testing:
* Running Bands or Xps alone without dos or pdos set returns the error: `ValueError: one of the parameters 'identifier', pk', 'uuid' or 'label' has to be specified`. You can actually request a Pdos calculation in step 2, set the dos.x and projwfc.x codes in step 3, then un-tick the Pdos calculation in step 2 to avoid the error without running Pdos.
* Removing the line `codes = {key: orm.load_node(value) for key, value in codes.items()}` and just passing `codes` without converting the inputs (i.e. `codes = parameters["codes"]`) works without issues for Bands and Xps, but causes a different crash for Pdos. Unfortunately, I can't find the traceback for that one right now, but it seemed to be that the values for dos and pdos codes were in the form of their UUID, rather than a node like for pw.x.

Hope this helps a little in diagnosing the problem.

unkcpz: Thanks a lot for the details @PNOGillespie very helpful.

> but it seemed to be that the values for dos and pdos codes were in the form of their UUID, rather than a node like for pw.x.

I have a feeling this could be related to the source of issue. The change look all good and I should not block the PR. approved.

unkcpz (approved):

PR #507, opened 2023-10-13
fix pseudodojo in report and load_panel_value

superstar54:


AndresOrtegaGuerrero:
@superstar54 So far the bug is fixed, could you re-base the PR so I can test that the Reset GUI is also working on the Pseudos widget ?

src/aiidalab_qe/app/configuration/pseudos.py
```diff
@@ -195,7 +195,10 @@ def _update_settings_from_protocol(self, protocol):
             self.set_from_pseudo_family(pseudo_family)
 
     def set_from_pseudo_family(self, pseudo_family):
-        family, _, functional, accuracy = pseudo_family.split("/")
+        if pseudo_family.startswith("SSSP"):
+            family, _, functional, accuracy = pseudo_family.split("/")
+        elif pseudo_family.startswith("PseudoDojo"):
+            family, _, functional, _, accuracy, _ = pseudo_family.split("/")
```
unkcpz, line 201: I think here it is a good practice to use dataclass `PseudoFamily`, can you give it a try?

src/aiidalab_qe/app/result/summary_viewer.py
```diff
@@ -69,8 +69,9 @@ def generate_report_parameters(qeapp_wc):
         }
     )
     # update pseudo family information to report
-    pseudo_family = ui_parameters["advanced"]["pseudo_family"]
-    pseudo_family = PROTOCOL_PSEUDO_MAP[ui_parameters["workchain"]["protocol"]]
+    pseudo_family = ui_parameters["advanced"].get("pseudo_family")
+    if pseudo_family is None:
+        pseudo_family = PROTOCOL_PSEUDO_MAP[ui_parameters["workchain"]["protocol"]]
```
unkcpz, line 74: Is this mean the `pseudo_family` may not stored in the `ui_parameters`? Then the problem should be the `ui_parameters` not properly generated, rather than workaround the issue here.

unkcpz (changes requested):
Thanks @superstar54, I have two requests.
Can you also add the uint test to reproduce the exception and the fix? I think it won't be hard.

superstar54:
@AndresOrtegaGuerrero 
> could you re-base the PR so I can test that the Reset GUI is also working on the Pseudos widget ?

I rebased the PR, please try.

src/aiidalab_qe/app/configuration/pseudos.py
superstar54: `dataclass` is typically used to create classes that are primarily meant to store data. I think it isn't not needed here.

src/aiidalab_qe/app/result/summary_viewer.py
superstar54: I will remove the `if` here. But when I added it, I thought of another scenario: when a user uploads custom pseudos, we should report something else here. I opened an issue #511 .

AndresOrtegaGuerrero:
@superstar54 Thank you!  I tested it, and there don't appear to be any issues on my end. I have a suggestion: since the original issue was related to PseudoDojo, perhaps we could add a specific test to ensure that the PseudoDojo library isn't failing, in case another PR disrupts this logic. This might be a good addition since the current test only checks PBESol, which is the default. (Or to modify the current test to check PseudoDojo instead ? )

src/aiidalab_qe/app/configuration/pseudos.py
unkcpz: > used to create classes that are primarily meant to store data

The "family", "functional" and "accuracy" are all data in this context. 

Moreover, If you check the PEP that introduced dataclass (https://peps.python.org/pep-0557/), it didn't mention that is specific for storing data, where did you get this conclusion? 

The goal of using an advance type is to simplify the code and make it less error-prone (here, developers basically has no idea what this `pseudo_family` string looks like therefore don't know which field should be assigned to which). If you think dataclass is too much, you can try `namedtuple`.

unkcpz (changes requested):
See my comment on dataclass, I insist to make it more clear.

src/aiidalab_qe/app/configuration/pseudos.py
superstar54: > here, developers basically has no idea what this pseudo_family string looks like therefore don't know which field should be assigned to which

Could you post the code here using `dataclass` and show me how the `dataclass` could solve this? I want to see a clear example.

unkcpz: Sure, can I add a commit to the pr?

unkcpz:
@superstar54 please check the last commit, all `split('/')` are removed and replaced by the method of newly defined dataclass. I also renamed some methods and attributes to make the concept more proper.

@AndresOrtegaGuerrero could you give it a test, the code changed quite a bit maybe there are new issues I didn't notice.

tests/test_configure.py
```diff
@@ -35,6 +35,10 @@ def test_set_configuration_parameters():
     wg.set_configuration_parameters(parameters)
     new_parameters = wg.get_configuration_parameters()
     assert parameters == new_parameters
+    # test pseudodojo
+    parameters["advanced"]["pseudo_family"] = "PseudoDojo/0.4/PBEsol/SR/standard/upf"
+    wg.set_configuration_parameters(parameters)
+    new_parameters = wg.get_configuration_parameters()
```
unkcpz, line 41: I guess you missed the `assert` clause?

unkcpz (changes requested):

tests/test_configure.py
superstar54: Yes, done. thanks.

superstar54:
I had an offline discussion with @unkcpz , the reason he suggested using `dataclass` is that he found that the code segments are also used in another place. For this reason, I agree that we use it to enhance the code's reusability.

AndresOrtegaGuerrero (approved):
Thank you both for your work! LGTM

PR #510, opened 2023-10-17
Fix/remove workchain output

### LINKED ISSUE #390: Make the output of QeAppWorkchain namespace of every sub-workchain
The QeAppWorkchain now run the output of every sub-workchain to one level hierarchy output. Therefore the output ports are many and don't have clear categories for which one is output of which workchain. 
It will be more clear if every sub-workchain has its own namespace for output ports.

The current output ports are:

```
        spec.output('structure', valid_type=StructureData, required=False)
        spec.output('band_parameters', valid_type=orm.Dict, required=False)
        spec.output('band_structure', valid_type=BandsData, required=False)
        spec.output('nscf_parameters', valid_type=orm.Dict, required=False)
        spec.output('dos', valid_type=XyData, required=False)
        spec.output('projections', valid_type=Orbital, required=False)
        spec.output('projections_up', valid_type=Orbital, required=False)
        spec.output('projections_down', valid_type=Orbital, required=False)
```

better be sort of like:

```
        spec.output('final_structure', valid_type=StructureData, required=False)
        spec.output_namespace('relax', dynamic=True)
        spec.output_namespace('bands', dynamic=True)
        spec.output_namespace('pdos', dynamic=True)
```

superstar54:
Fix #390 

The electronic structure panel is still hardcoded because it needs the results from both bands and pdos plugins. When we find a good solution to support results from multiple plugin, we can remove both the code in the electronic structure result and the code in the workchain.

For the moment, we remove the code in the workchain, but still hardcoded the electronic structure result panel.

### DIFF
--- a/src/aiidalab_qe/app/result/electronic_structure.py
+++ b/src/aiidalab_qe/app/result/electronic_structure.py
@@ -18,13 +18,15 @@ def export_data(work_chain_node, group_dos_by="atom"):
 
 
 def export_pdos_data(work_chain_node, group_dos_by="atom"):
-    if "dos" in work_chain_node.outputs:
-        _, energy_dos, _ = work_chain_node.outputs.dos.get_x()
-        tdos_values = {f"{n}": v for n, v, _ in work_chain_node.outputs.dos.get_y()}
+    if "pdos" in work_chain_node.outputs:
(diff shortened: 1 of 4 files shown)


unkcpz (approved):
Super! I didn't test the change but I think we'll anyway test it the test phase before official release. Let's merge it now.

Issue #412, opened 2023-06-06
Add and activate the tot_magnetization widget when running magnetic simulation on insulator 


unkcpz:
```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine iosys (1):
     fixed occupations and lsda need tot_magnetization
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
when trying an insulator with magnetism. I think QE App should catch this and avoid QE to crash.

_Originally posted by @giovannipizzi in https://github.com/aiidalab/aiidalab-qe/issues/245#issuecomment-1579026933_

AndresOrtegaGuerrero:
@unkcpz  I guess it can be together with the starting_magnetization widget and tags, otherwise the app gives the impression you can calculate antiferromagnetic cases.

AndresOrtegaGuerrero:
@unkcpz Do we need like a new feature for tot_magnetization in qe app? because we don't have this feature

unkcpz:
I think we can add it, it is quite easy from what I understand by just adding a widget for setting it, correct?

AndresOrtegaGuerrero:
Ok i will update the widget! , though according to the manual in simulation we use `tot_magnetization` or  `starting_magnetization` so it has to be like one or the other  [tot_magnetization](https://www.quantum-espresso.org/Doc/INPUT_PW.html#tot_magnetization). However now that I read the issue, the problem is about selecting `insulator` with `magnetism` so i guess we also need to see the workchain , to check that a magnetization is defined if this combination is set.

PR #512, opened 2023-10-17
Fix tot magnetization

AndresOrtegaGuerrero:
This PR solves #412

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced.py
@@ -556,6 +556,7 @@ class MagnetizationSettings(ipw.VBox):
     def _update_widget(self, change):
         self.input_structure = change["new"]
         self.update_kinds_widget()
+        self.display_kinds()
 
     def get_magnetization(self):
         """Method to generate the dictionary with the initial magnetic moments"""


AndresOrtegaGuerrero:
@unkcpz  This is a draft, but in order for me to continue there must be a PR (in aiida-quantumespresso) to deal with the logic that if the system has` tot_magnetization` , it should remove `starting_magnetization `for PwBaseWorkChain. Likewise that the logic in PdosWorkChain and BandsWorkchain is not affected

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -176,6 +176,11 @@ def _update_input_structure(self, change):
             self.pseudo_setter.structure = change["new"]
             self._display_mesh()
 
+    @tl.observe("electronic_type")
+    def _electronic_type_changed(self, change):
+        """Input electronic_type changed, update the widget values."""
+        self.magnetization.electronic_type = change["new"]
```
unkcpz, line 182: I like this, very clear to me and I saw you made `electronic_type` a trait of `MagnetizationSettings`, exactly what I think about ;)

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -237,12 +242,10 @@ def get_panel_value(self):
...
         parameters["pw"]["parameters"]["SYSTEM"]["tot_charge"] = self.total_charge.value
-        # there are two choose, use link or parent
-        if self.spin_type == "collinear":
-            parameters[
-                "initial_magnetic_moments"
-            ] = self.magnetization.get_magnetization()
+
+        # add the kpoints_distance to the parameters
         parameters["kpoints_distance"] = self.value.get("kpoints_distance")
+
```
unkcpz, line 248: Can you write a comment here for all conditions? 
For instance:

```
# insulator and colliner: xxxx
# metal and collinear: xxx
```

Can the if..else clause below can be simplified with inverse and break return?

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -302,8 +328,11 @@ def reset(self):
             self.override.value = False
             self.smearing.reset()
             # reset the pseudo setter
-            self.pseudo_setter.structure = None
-            self.pseudo_setter._reset()
+            if self.input_structure is None:
+                self.pseudo_setter.structure = None
+                self.pseudo_setter._reset()
+            else:
+                self.pseudo_setter._reset()
```
unkcpz, line 335: Why can not call pseudo_setter.structure = None if input_structure is not `None`? If I understand correct, the reset should also do `self.structure=None` first.

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -333,45 +362,67 @@ class MagnetizationSettings(ipw.VBox):
...
+        self.magnetization_type.disabled = self.disabled
 
     def reset(self):
         self.disabled = True
-        self.kinds = None
-        self.description.value = "Define magnetization: Input structure not confirmed"
-        with self.kinds_widget_out:
-            clear_output()
-            display(self.kinds)
+        self.tot_magnetization.value = 0.0
```
unkcpz, line 416: I'd suggest making the default value defined as class attribute `_DEFAULT_TOT_MAGNETIZATION`.

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -333,45 +362,67 @@ class MagnetizationSettings(ipw.VBox):
...
+        #
+        if self.input_structure is None:
+            self.description.value = (
+                "<b>Magnetization: Input structure not confirmed</b>"
```
unkcpz, line 420: This also appears in the `__init__`, which should be a way to only define it once in a single place.

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -392,11 +443,30 @@ def create_kinds_widget(self):
...
+                    clear_output()
+
     def update_kinds_widget(self):
         self.input_structure_labels = self.input_structure.get_kind_names()
         self.kinds = self.create_kinds_widget()
-        self.description.value = "Define magnetization: "
-        self.display_kinds()
+        self.description.value = "<b>Magnetization</b>"
+
+    def _display_starting(self, value):
```
unkcpz, line 463:
```suggestion
    def _render(self, value):
```

src/aiidalab_qe/plugins/bands/workchain.py
```diff
@@ -211,6 +211,11 @@ def get_builder(codes, structure, parameters, **kwargs):
     bands.pop("relax")
     bands.pop("structure", None)
     bands.pop("clean_workdir", None)
+
+    if scf_overrides["pw"]["parameters"]["SYSTEM"].get("tot_magnetization") is not None:
+        bands.scf["pw"]["parameters"]["SYSTEM"].pop("starting_magnetization", None)
+        bands.bands["pw"]["parameters"]["SYSTEM"].pop("starting_magnetization", None)
```
unkcpz, line 217: If I understand correctly, the code here is for when `tot_magnetization` is set, the `starting_magnetization` should not be set, but this exclusive existence can be guaranteed when the builder is created, no?

unkcpz (changes requested):
Thanks a lot for trying this @AndresOrtegaGuerrero, The logic looks all good, but I think it requires some iron-up on the clarity of implementation. Can you fix the conflicting and we arrange a pair coding session to check things together?

AndresOrtegaGuerrero:
To complete this PR the following requirements should be met

- [ ] PdosWorkChain should be adapted to run calculations using tot_magnetization https://github.com/aiidateam/aiida-quantumespresso/issues/988
- [ ] the logic of the qe app should be adapted for having two fermi energies (up/down): get_pdos_data, projections, bands-widget

superstar54 (approved):
@AndresOrtegaGuerrero , thanks for the work!  LGTM!

There are two things we could improve:
- If users set `Magnetism` to `Off`, the Magnetization section could be hidden.
- When switching on `Magnetism`, the pdos plot could has an optios to plot `up` and `down` in two sides.
Could you create issues on this? so that we can do these in the future PRs.

superstar54:
Before merging the PR, two small changes are needed:

- the `tot_magnetization` should be an integer for a `fixed` occupation.
- There are two `tot_magnetization` when I load the previous job.
image

AndresOrtegaGuerrero:
@superstar54 thank you for the review, I thought i fixed the duplication i will have a look. Regarding the tot_magnetization , i kept as a Real number since in the pw.x manual is a real.

superstar54:
In the fixed occupation, if one sets a non-integer value for the `tot_magnetization,` pw will raise an error. In the smearing case, it will not.

AndresOrtegaGuerrero:
> In the fixed occupation, if one sets a non-integer value for the `tot_magnetization,` pw will raise an error. In the smearing case, it will not.

Then i guess it should be conditioned then , thank you !

PR #517, opened 2023-10-18
Fix override

AndresOrtegaGuerrero:
This PR fixes minor issues with Reset_GUI

   If you click multiple times override in advanced tab, it can delete the kinds and the input_structure
   It can also delete the pseudos and pseudos input_structure.


superstar54:
@AndresOrtegaGuerrero Thanks for the PR. LGTM.

Another issue also related to the reset is the kpoints distance. When I select `protocol` as `fast`, then go to the `advanced` panel, I click and then unselected the override checkbox. The kpoints distance is generated from the `moderate` protocol. Could you also fix this in this PR

One solution is to remove this line and also the `_default_protocol` for the `advanced` setting. What do you think?
```
self.protocol = self._default_protocol
```

superstar54:
@AndresOrtegaGuerrero There are some tests failed. Please check.

superstar54 (approved):
I fixed the tests.

PR #518, opened 2023-10-20
Add plugin: XPS

superstar54:
This PR adds a plugin for calculating X-ray Photoelectron Spectroscopy (XPS) spectra using the [`XpsWorkChain`](https://github.com/aiidateam/aiida-quantumespresso/blob/main/src/aiida_quantumespresso/workflows/xps.py) of aiida-quantumespresso package.

## Setting Panel
The Setting panel allows users to
- select the peak (element and orbital) to be calculated. The availability of peak in the panel depends on the corresponding pseudopotentials.
- adjust the parameters based on the input structure.
- 
Screenshot from 2024-02-14 13-04-05

## Results Panel
The Result panel displays the final XPS spectrum.
- broadened spectrum using a Voight profile (combined Lorenzian and Gaussian)
- the chemical shift values (differences in total energy relative to the lowest value)
- absolute binding energy
- the Offset Energies (δ)

Screenshot from 2024-02-14 13-03-10

Pseudopotentials
XPS calculations in QE require core-hole pseudopotentials. We supply a set of pseudopotentials specific to XPS calculations in the [XPS Pseudopotential Repository](https://github.com/superstar54/xps-data). This repository provides the core-hole pseudopotentials as an AiiDA archive file. There are two kind of pseudopotentials available:
- pbe
- pbesol

## WorkChain
To make the setting as simple as possible for user. The following default setting are used.
- `core_hole_treatment`. 
  - `full` for molecule
  - `xch_smear` for metal
  - `xch_fixed` for insulator
- `supercell_min_parameter` updated based on `protocol`.


superstar54:
It seems that the plugins still use the default pseudo family from pw workchain.
```
ValueError: required pseudo family `SSSP/1.2/PBEsol/efficiency` is not installed. Please use `aiida-pseudo install` toinstall it.
```

src/aiidalab_qe/plugins/xps/setting.py
```diff
@@ -0,0 +1,249 @@
...
+        """<div style="line-height: 140%; padding-top: 0px; padding-bottom: 5px">
+        You have three options:<br>
+        (1) XCH(smear): places the excited electron into the conduction band, suitable for extend system.<br>
+        (2) XCH(fixed): places the excited electron into the conduction band, suitable for extend system.<br>
+        (3) Full: remove one electron from the system, suitable for molecule. </div>"""
+    )
+
+    pseudo_title = ipw.HTML(
+        """<div style="padding-top: 0px; padding-bottom: 0px">
+        <h4>Pseudopotential</h4></div>"""
```
PNOGillespie, line 50: Just to make things easier to read:
```suggestion
        <h4>Core-Hole pseudopotential group</h4></div>"""
```

src/aiidalab_qe/plugins/xps/setting.py
```diff
@@ -0,0 +1,249 @@
...
+
+    peak_title = ipw.HTML(
+        """<div style="padding-top: 0px; padding-bottom: 0px">
+        <h4>Select peak</h4></div>"""
+    )
+    peak_help = ipw.HTML(
+        """<div style="line-height: 140%; padding-top: 6px; padding-bottom: 0px">
+        The list of peaks to be considered for analysis.
+        </div>"""
+    )
```
PNOGillespie, line 66: Would something like this make sense for readability? "Core-Hole" or "Core-Level" seems like it would be more intuitive, but if "Peak" is the correct technical term in this case then don't worry.
```suggestion
    peak_title = ipw.HTML(
        """<div style="padding-top: 0px; padding-bottom: 0px">
        <h4>Select core-level</h4></div>"""
    )
    peak_help = ipw.HTML(
        """<div style="line-height: 140%; padding-top: 6px; padding-bottom: 0px">
        The list of core-levels to be considered for analysis.
        </div>"""
    )
```

src/aiidalab_qe/plugins/xps/setting.py
```diff
@@ -0,0 +1,249 @@
...
+            self.peak_title,
+            self.peak_help,
+            ipw.HBox(
+                [self.peak_list],
+            ),
+            # self.supercell_title,
+            # self.supercell_help,
+            # ipw.HBox(
+            # [self.supercell_min_parameter],
+            # ),
```
PNOGillespie, line 158: There are quite a few commented-out sections here which should be cleaned up if possible:
```suggestion
            self.pseudo_title,
            self.pseudo_help,
            self.pseudo_group,
            self.peak_title,
            self.peak_help,
            ipw.HBox(
                [self.peak_list],
            ),
```

Does this also mean (based on the fact that `supercell_min_parameter` is not used) that the supercell size used in calculation is defined entirely by the protocol?

PNOGillespie (changes requested):
Hi @superstar54, I've had a look at the code and, generally speaking, it looks good. I've left some suggested changes here and there to improve readability and clean up commented-out blocks of code.

The only important change IMO is a selection widget for the element to display in the `Result` panel - see the [relevant comment](https://github.com/aiidalab/aiidalab-qe/pull/518#discussion_r1474614565) for discussion.

src/aiidalab_qe/plugins/xps/setting.py
superstar54: Thanks! I change it to core-level

src/aiidalab_qe/plugins/xps/result.py
```diff
@@ -0,0 +1,214 @@
...
+        g.layout.xaxis.title = "Chemical shift (eV)"
+        g.layout.xaxis.autorange = "reversed"
+        #
+        spectra = xps_spectra_broadening(
+            chemical_shifts, equivalent_sites_data, gamma=gamma.value, sigma=sigma.value
+        )
+        for element, data in spectra.items():
+            # print(element, data)
+            for site, d in data.items():
+                g.add_scatter(x=d[0], y=d[1], fill="tozeroy", name=f"{element}_{site}")
```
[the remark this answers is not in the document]
superstar54: Thanks! I added the selection dropdown for different elements, like in the XAS plugin.

src/aiidalab_qe/plugins/xps/result.py
```diff
@@ -0,0 +1,214 @@
...
+        super().__init__(node=node, **kwargs)
+
+    def _update_view(self):
+        import plotly.graph_objects as go
+
+        voigt_profile_help = ipw.HTML(
+            """<div style="line-height: 140%; padding-top: 10px; padding-bottom: 10px">
+        The Voigt function:
+        </div>"""
+        )
```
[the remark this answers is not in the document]
superstar54: I added a description for the parameters and also a link to Wikipedia.

src/aiidalab_qe/plugins/xps/setting.py
superstar54: I changed as you suggested.

src/aiidalab_qe/plugins/xps/setting.py
superstar54: Yes. This would be more user friendly

superstar54:
@PNOGillespie Thanks for your review. I made the changes as you suggested. I added the dropdown button to select the spectrum for different elements. I also added a table to show the offset energies, which are used to calibrate the calculated binding energy.

src/aiidalab_qe/plugins/xps/result.py
```diff
@@ -0,0 +1,247 @@
...
+            style={"description_width": "initial"},
+        )
+        sigma = ipw.FloatSlider(
+            value=0.3,
+            min=0.1,
+            max=1,
+            description="Gaussian profile (σ)",
+            disabled=False,
+            style={"description_width": "initial"},
+        )
```
PNOGillespie, line 118: I didn't know that Python strings could use non-ASCII characters :laughing:. While these don't raise an error at the moment, I'd suggest encoding these properly just to avoid any potential problems elsewhere:
```suggestion
        gamma = ipw.FloatSlider(
            value=0.3,
            min=0.1,
            max=1,
            description="Lorentzian profile ($\gamma$)",
            disabled=False,
            style={"description_width": "initial"},
        )
        sigma = ipw.FloatSlider(
            value=0.3,
            min=0.1,
            max=1,
            description="Gaussian profile ($\sigma$)",
            disabled=False,
            style={"description_width": "initial"},
        )
```

src/aiidalab_qe/plugins/xps/result.py
```diff
@@ -0,0 +1,247 @@
...
+                        else:
+                            g.data[i].fill = None
+                        g.data[i].name = data[i]["site"].replace("_", " ")
+
+                else:
+                    g.data = []
+                    for d in data:
+                        g.add_scatter(
+                            x=d["x"], y=d["y"], fill=fill.value, name=d["site"]
+                        )
```
PNOGillespie, line 202: This section here causes a crash because you're calling the value from the UI element rather than converting it to a type of fill method (e.g. "tozeroy"):
```suggestion
                        if fill.value:
                            fill_type = "tozeroy"
                        else:
                            fill_type = None
                        g.add_scatter(
                            x=d["x"], y=d["y"], fill=fill_type, name=d["site"]
                        )
```

src/aiidalab_qe/plugins/xps/result.py
```diff
@@ -0,0 +1,247 @@
...
+
+        ui_parameters = self.node.base.extras.get("ui_parameters", {})
+        ui_parameters = deserialize_unsafe(ui_parameters)
+        correction_energies = ui_parameters["xps"]["correction_energies"]
+        # create a table for the correction energies using ipywidgets
+        correction_energies_table = ipw.HTML(
+            """<h4>Offset Energies (δ)</h4>
+            <div>The binding energies should be corrected by these constant offsets. These offsets mainly depend on chemical element and core level, and are determined by comparing the calculated core electron binding energy to the experimental one.</div>
+            <table><tr><th>Core-level</th><th>Value (eV)</th></tr>"""
+        )
```
PNOGillespie, line 238: The first sentence should remind the user how the correction energies are used in practice:
```suggestion
        correction_energies_table = ipw.HTML(
            """<h4>Offset Energies (δ)</h4>
            <div>When comparing binding energies reported by AiiDALab-QE to experimental data, these should be corrected by the given offset listed below</div>
            <table><tr><th>Core-level</th><th>Value (eV)</th></tr>"""
        )
```

I would also recommend moving the second sentence (or maybe the relevant concept or description) to the documentation since this is not really needed here.

src/aiidalab_qe/plugins/xps/result.py
```diff
@@ -0,0 +1,247 @@
...
+        for core_level, value in correction_energies.items():
+            element = core_level.split("_")[0]
+            if element not in self.spectrum_select_options:
+                continue
+            exp = value["exp"]
+            correction_energies_table.value += (
+                f"<tr><td>{core_level}</td><td>{exp}</td></tr>"
+            )
+        return correction_energies_table
```
PNOGillespie, line 247: Just a little addition to make things clearer:
```suggestion
        for core_level, value in correction_energies.items():
            element = core_level.split("_")[0]
            if element not in self.spectrum_select_options:
                continue
            exp = value["exp"]
            if exp >= 0.0:
                sign = "+"
            elif exp < 0:
                sign = ""
            correction_energies_table.value += (
                f"<tr><td>{core_level}</td><td>{sign}{exp}</td></tr>"
            )
        return correction_energies_table
```

If the values can be positive or negative, then it's usually a good idea to give the sign explicitly even if the number is positive.

The string for the core level name should probably also have the underscore removed, since this doesn't format so well in non-monospace text. I understand this is also the way that the labels for the core levels are written in the `Setting` panel, but I wouldn't worry about changing those at this stage.

PNOGillespie (changes requested):
Hi @superstar54. Code looks good for the most part. I've added a few final changes, after which I think this PR is complete.

src/aiidalab_qe/plugins/xps/result.py
superstar54: thanks! I changed it.

src/aiidalab_qe/plugins/xps/result.py
superstar54: thanks! The sign is indeed very important.

src/aiidalab_qe/plugins/xps/result.py
superstar54: thanks! I made the changes.

superstar54:
@PNOGillespie, thanks for your review and very useful suggestions! I made all the changes as suggested.

PNOGillespie (approved):
Looks good to me, nice work @superstar54!

PR #519, opened 2023-10-20
Fix/remove deprecations

unkcpz:
Follow the https://github.com/aiidalab/aiidalab-widgets-base/pull/509, the warnings should be get rid of here as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,6 @@ concurrency:
 jobs:
 
     pre-commit:
-    # Adapted from: https://github.com/CasperWA/voila-optimade-client
 
         runs-on: ubuntu-latest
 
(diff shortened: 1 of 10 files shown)


unkcpz:
Hi @danielhollas, did you also have the following warning after you turn waring to error in pytest?

```
        # The call to .warn adds one frame, so bump the stacklevel up by one
>       warnings.warn(message, DeprecationWarning, stacklevel=stacklevel + 1)
E       DeprecationWarning: Jupyter is migrating its paths to use standard platformdirs
E       given by the platformdirs library.  To remove this warning and
E       see the appropriate new directories, set the environment variable
E       `JUPYTER_PLATFORM_DIRS=1` and then run `jupyter --paths`.
E       The use of platformdirs will be the default in `jupyter_core` v6

/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/jupyter_core/utils/__init__.py:90: DeprecationWarning
```

danielhollas:
Possibly, I've definitely seen it before. I think take a look at pyproject.toml at AWB and copy the ignore lines from there.

PR #521, opened 2023-10-20
remove blocker when updating code

### LINKED ISSUE #520: Step 3: Submission Blocker Remains if `pw.x` Code Selected Before Others
**Describe the bug**
On `main` version 23.10.0a2 (latest commit 0006ac23fbd1d0cfcaa55335102857c1fca961ff)  when selecting codes for calculation using the PdosWorkChain, the submission blocker remains stating that the dos.x and projwfc.x codes must be set even after the relevant codes are selected.

The issues only arises if the `pw` code is set first. If the `dos` or `projwfc` codes are set first, then the submission blocker clears and the WorkChain can be started with no issues

**To Reproduce**
1. Load & confirm structure, set pdos property in step 2, confirm and move to step 3
2. Select code from pw.x dropdown menu
3. Select codes from dos.x and projwfc.x dropdown menus
4. Submission blocker remains

**Expected behavior**
Submission blocker requiring codes to be set should clear once relevant codes are selected.

**Screenshots**
#### Initial Step 3 Screen:
Step_3_Page-Initial
#### Select pw.x Code:
Step_3_Page-PW_First
#### Selecting dos.x and projwfc.x Codes Does not Clear Submission Blocker
Step_3_Page-Bug_Found
#### Starting from projwfc.x and dos.x:
Step_3_Page-Pdos_First
#### Adding pw.x Code Afterwards Clears the Submission Blocker:
Step_3_Page-No_Bug_Found

**Version (if known)**
 - App version: 23.10.0a2 (commit: 0006ac23fbd1d0cfcaa55335102857c1fca961ff)

**Additional context**
This issue persists with other apps if they use a separate code and will fix in the same way.

Also, unsure if this is relevant but the App can't install `localhost` codes on my laptop. This doesn't prevent calculation work however.

superstar54:
fix #520. Remove the code blocker when updating code widgets.

### DIFF
--- a/src/aiidalab_qe/app/submission/__init__.py
+++ b/src/aiidalab_qe/app/submission/__init__.py
@@ -89,6 +89,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
         for _, entry_point in self.code_entries.items():
             for name, code in entry_point.items():
                 self.codes[name] = code
+                code.observe(self._update_state, "value")
                 self.code_children.append(self.codes[name])
         # set default codes
         self.set_selected_codes(DEFAULT_PARAMETERS["codes"])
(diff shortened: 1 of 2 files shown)


unkcpz (approved):
@superstar54 Good and thanks for adding the test!

PR #522, opened 2023-10-21
add `electronic_structure` plugin

superstar54:
This PR supports
- showing the result from multiple plugins.
- only shows the result panel when it is available.

One can use the workchain_labels to specify which plugins (outputs) are needed for this result panel. For example, the `electronic_structure` result panel needs `pdos` and `bands`.
```python
class Result(ResultPanel):
    title = "Electronic Structure"
    workchain_labels = ["bands", "pdos"]
```


tests/test_result.py
```diff
@@ -16,23 +16,6 @@ def test_workchainview(generate_qeapp_workchain):
...
-
-    from aiidalab_qe.app.result.workchain_viewer import WorkChainViewer
-
-    wkchain = generate_qeapp_workchain()
-    wkchain.node.set_exit_status(0)
-    wkchain.node.set_process_state(engine.ProcessState.FINISHED)
-    wcv = WorkChainViewer(wkchain.node)
-    assert (
-        "DOS grouped by:"
-        == wcv.result_tabs.children[2].children[0].children[0].children[0].value
```
unkcpz, line 34: I have no idea of all these children are, a detailed comment will be very useful.
Since now things are implemented as "Panel", is some of these children are panel? If so, I think `Panel` can have some way to associate children with name to make it easy to access.

src/aiidalab_qe/plugins/electronic_structure/result.py
```diff
@@ -182,3 +186,67 @@ def cmap(label: str) -> str:
...
+                            layout=ipw.Layout(
+                                justify_content="flex-start", width="120px"
+                            ),
+                        ),
+                        group_dos_by,
+                    ]
+                ),
+            ],
+            layout={"margin": "0 0 30px 30px"},
+        )
```
unkcpz, line 224: Please avoid a deep nest of widgets, if there are many, define it outside with a comment on what the box is for and where it is displayed.

unkcpz (changes requested):
Thanks @superstar54, I have two things ask to change, please have a look.

unkcpz:
BTW, there is no need to always merge the branch, if I find time to review I'll merge the main before I review. Usually, this happens before the final merge if there is no conflict.

tests/test_result.py
superstar54: The test here is too detailed, and not necessary, so I removed it. A new test that checks the length of the panel children is added.

src/aiidalab_qe/plugins/electronic_structure/result.py
superstar54: I refactored the code. Now, it's become more simple and clear.

superstar54:
@unkcpz I refactored the code and added detailed comments. It's ready for review. 

Please focus on the plugin part, which is the primary purpose of this PR.

The part of plotting the electronic structure is simply copied and pasted from the previous code, with few updates. If the code needs polished, it is better to open another PR.

src/aiidalab_qe/app/result/workchain_viewer.py
```diff
@@ -121,89 +100,39 @@ def _update_view(self):
...
-        )
-        #
-        data = export_data(self.node, group_dos_by=group_dos_by.value)
-        bands_data = data.get("bands", None)
-        dos_data = data.get("dos", None)
-        _bands_plot_view = BandsPlotWidget(
-            bands=bands_data,
-            dos=dos_data,
-        )
+                    if False not in results_ready:
```
unkcpz, line 124:
```suggestion
                    if all(results_ready):
```

src/aiidalab_qe/app/result/workchain_viewer.py
```diff
@@ -121,89 +100,39 @@ def _update_view(self):
...
+                        result._update_view()
+                        self._results_shown.add(result.identifier)
+                        # add this plugin result panel
+                        self.result_tabs.children += (result,)
+                        self.result_tabs.set_title(
+                            len(self.result_tabs.children) - 1, result.title
```
unkcpz, line 130: I guess `len(self.result_tabs.children) - 1` is the idx of the tab. Please have a variable for it with comment.

src/aiidalab_qe/plugins/bands/result.py
```diff
@@ -29,14 +29,15 @@ class Result(ResultPanel):
     """Result panel for the bands calculation."""
 
     title = "Bands"
+    workchain_labels = ["bands"]
```
unkcpz, line 32: What is this labels for? If there are more than one where will those shown?

tests/test_result.py
unkcpz: Nice, that is better!

tests/test_plugins_electronic_structure.py
```diff
@@ -0,0 +1,20 @@
...
+    wkchain.node.set_exit_status(0)
+    wkchain.node.set_process_state(engine.ProcessState.FINISHED)
+    wcv = WorkChainViewer(wkchain.node)
+    # find the tab with the identifier "electronic_structure"
+    # the built-in summary and structure tabs is not a plugin panel,
+    # thus don't have identifiers
+    tab = [
+        tab
+        for tab in wcv.result_tabs.children
+        if getattr(tab, "identifier", "") == "electronic_structure"
```
unkcpz, line 17: I assume `identifier` is a mendatory attribute of panel, so get this be: 
```suggestion
        if tab.identifier == "electronic_structure"
```

unkcpz (changes requested):
Hi @superstar54, thanks a lot for refactoring, very beautiful! Just some minor questions.

src/aiidalab_qe/plugins/bands/result.py
superstar54: One can use the workchain_labels to specify which plugins (outputs) are needed for this result panel. For example, the electronic_structure result panel needs pdos and bands.
 ```
 workchain_labels = ["bands", "pdos"]
 ```
To make this clear, I added a comment in the `ResultPanel`.

superstar54:
@unkcpz  I updated the code based on your suggestions. It's ready for another review.

unkcpz (approved):
Thanks @superstar54 ! All good for me. Ready to 🚢

PR #525, opened 2023-10-22
Correct the regex to get version tag from tag

unkcpz:


unkcpz:
The change has to be validated after the version is released, so I merge it.

PR #532, opened 2023-10-23
Pre-install app by aiidalab install

unkcpz:
The pre-installation of the app is through run `aiidalab install` from local folder, instead of using `pip install`, this mimics the installation operation from AppStore and prevents issue such as after removing `setup.py` the dependencies are not found.
I'll keep this as draft without adding `setup.py` back and after the new docker stack release with new `aiidalab/aiidalab`, the test should pass.

- The first commit should success because I add the `setup.py` back.
- The second commit is again remove `setup.py`.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,6 +17,7 @@ on:
             - .github/actions/load-image/action.yml
 
             - src/**
+            - docker/**
             - qe.ipynb
             - setup.cfg
             - pyproject.toml
@@ -40,6 +41,7 @@ on:
             - .github/actions/load-image/action.yml
 
             - src/**
+            - docker/**
             - qe.ipynb
             - setup.cfg
             - pyproject.toml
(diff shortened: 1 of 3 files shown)


unkcpz:
@danielhollas, now the aiidalab install is tested 😄

danielhollas:
@unkcpz I would suggest to actually put the `setup.py` back for now. Otherwise users on older images will not be able to install the new version (or worse, the installation will fail). 

When you do remove it (or decide to keep it removed) you should specify the minimum `aiidalab` version to prevent failure during installation.

unkcpz:
> I would suggest to actually put the setup.py back for now. Otherwise users on older images will not be able to install the new version (or worse, the installation will fail).

Ah, very true and always better before we release the stable version. This is sort of like breaking the "forward"-compatibility. I added it back, and it is ready for review.

danielhollas (approved):
LGTM! :rocket:

PR #534, opened 2023-10-25
Increase timeout of services response waiting

unkcpz:
We encountered this in AWB 2 days ago. It seems the same reason here the CI failed with timeout.


unkcpz:
This is not needed, I guess it is a GitHub action's network issue. closed.

PR #531, opened 2023-10-23
not reset steps when  select a new structure

superstar54:
Fix #527  

This bug is caused by the loop between traitlets. 
- the first loop. When selecting a new structure, it will reset all steps. This will reset the structure again.
- the second loop. When resetting the process of the submit_step, it will change the value of the workchain selector to `None`, which will reset all steps again.

PR #535, opened 2023-10-25
Adding dynamic visualization of codes

mikibonacci:
This PR enables the possibility to dynamically show codes in step 3, even if something is changed in step 2 after the confirmation. If you confirm again with different set of properties, now the shown code are updated, at variance with before. 
Moreover, codes that are not shown will not occupy blank space in the app, as instead is now in the main branch. I achieved this using display instead of visibility.


src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -308,10 +308,10 @@ def _observe_state(self, change):
         with self.hold_trait_notifications():
             self.submit_button.disabled = change["new"] != self.State.CONFIGURED
 
-    @tl.observe("previous_step_state")
+    @tl.observe("previous_step_state", "input_parameters")
```
unkcpz, line 311: Good, I think this observation on `input_paramters` is correct.

unkcpz (approved):
I agree that using `display` is more proper otherwise there will be a large blank space. 

Just a tip on git, I guess this PR is based on some merged changes, and there is a long commits list. You can try to use `git rebase main` to clean it up next time.

src/aiidalab_qe/app/submission/__init__.py
superstar54: Hi @mikibonacci , could you please clarify why this is needed?

mikibonacci: Hi @superstar54, sure. So if you select properties in step 2 and then confirm, the corresponding codes will be shown in step 3. However, if you go back to step 2, select new properties (or deselect), the codes are not updated correspondingly. I guess this is due to the fact that the "previous_step_state" is not really changed, but the "input_parameters" yes. 

By explaining this, I noticed that probably it is better to separate the _update_state and _update_code_visibility.
I will put _update_code_visibility in a _observe_input_parameters function, if you agree. This is more clean and allow to leave only one argument in the observe. Do you also agree @unkcpz ?

superstar54:
Thanks for the improvement. @mikibonacci . I have one question.

> , even if something is changed in step 2 after the confirmation.

In my opinion, this feature is not needed. After the confirmation, the user should not change the setting anymore. Otherwise, it breaks the design of the Wizard UI, and makes the logic less clear.

mikibonacci:
I have one case in mind: what if, in step 3, I understand that I am forgetting a property and I go back to add it? probably we should let the user to go back and change it (in the same way as, if we change the structure, the following parameters are updated).

superstar54:
In this case, the user needs to reset the step2, which is not supported yet. See my comment here. https://github.com/aiidalab/aiidalab-qe/pull/531#issuecomment-1778586537, which I copied it here:
> There is a third option, which I think is the best, the reset button to reset the current step and its subsequent steps. However, we need to update the aiidalab-widget-base.

src/aiidalab_qe/app/submission/__init__.py
unkcpz: I agree and I find it is quite useful. Before submitting, I would imagine I still want to change anything in any previous steps.

I think the best design will be if step2 confirmed and moved to step3, user go back to step2 and change things. The confirm button will be activated again. When the button clicked it move to step3 and keep things that don't need to updated there. For the things that will be updated, raise a waring message that it is depended on the previous step and since it is changed "Please check and reset XXX before submit. ".

unkcpz: I know it is not easy and requires more work to do, but we personally use this flow quite a lot to go back and change things. I think it is a thing we must fix before the release.

mikibonacci:
The resetting of step 2 in this case would be before the submission of the job. After that, I think only reset should allow new modifications.

unkcpz:
Joke side, @superstar54 welcome to the hell. This is what happens very often when real users start to use it, we encounter it every time after a real live test session. More and more use cases raise up to break the app with "unpredict" operation flow.

superstar54:
> Yes, but this is what was supported in the previous version. Users can go back and change things and confirm again.

In the previous version, this is only partially supported. When the user changes `relax` and properties, it resets the state. But if the user changes other settings, e.g. kpoint, it will not reset the step. This is very confusing for the user.
We need a clear and simple way to tell users how to reset a step. We may need to provide a button similar to `un-confirm` to allow the user to un-confirm the step and continue editing it.

unkcpz:
Can we arrange a quick meeting today? Before that everyone write the best solution they think about. We decide how the flow should be. Also pining @AndresOrtegaGuerrero for join.

mikibonacci:
fine for me

superstar54:
I just ran a test. There is a bug in the confirmation button of step 2. The `confirm` button is not disabled after the confirmation. It should be disabled, just like in step 1.
Thus, one can not change the property and confirm it again.

unkcpz:
Hi @mikibonacci, next time to merge things, please to the "spaush and merge" with a clear commit message write up, thanks.

PR #537, opened 2023-10-26
Detects the unsaved changes in the already confirmed steps.

superstar54:
In the discussion with @unkcpz  and @mikibonacci , we proposed a solution to handle the situation where the user edits the already confirmed steps.
This PR detects the unsaved changes in the already confirmed steps. If there are any unsaved changes, it will block the submit step and show the blocker messages there.

The technique details:
- check the unsaved changes when the user leaves the step. This is achieved by observing the `selected_index` of the `_wizard_app_widget`, and comparing the related values (`confirmed_structure` in the step1, and `configuration_parameters` in step 2).
- The `_submission_blockers` is added in the app and linked to the `_app_submission_blockers` in the submit step.


src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -25,6 +25,7 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     configuration_parameters = tl.Dict()
 
     def __init__(self, **kwargs):
+        self.is_confirmed = False
```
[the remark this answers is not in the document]
unkcpz: And why don't we name it `self.saved`?

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -61,6 +61,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     previous_step_state = tl.UseEnum(WizardAppWidgetStep.State)
     input_parameters = tl.Dict()
     _submission_blockers = tl.List(tl.Unicode())
+    _app_submission_blockers = tl.List(tl.Unicode())
```
unkcpz, line 64: Please just use `_submission_blocker`.

src/aiidalab_qe/app/main.py
```diff
@@ -85,6 +94,30 @@ def __init__(self, qe_auto_setup=True):
...
+        return self._wizard_app_widget.steps
+
+    # Check unconfirmed change in the step when leaving the step.
+    def _observe_selected_index(self, change):
+        with self.submit_step.hold_sync():
+            if change["old"] is None:
+                return
+            # check if the step is confirmed before
+            # if not, we don't need to check the unconfirmed changes
+            previous_step = self.steps[change["old"]][1]
```
unkcpz, line 108: "previous_step" sound more like step2 for step3, if I understand correctly, this is for the step where the old index point to.
Maybe name it "step_old_idx" and "step_new_idx".

unkcpz (changes requested):
Thanks a lot! Some requests and I find some small issues in the behavior, here is how to reproduce:
1. select structure
2. confirm the configuration
3. now it is in step3, go back to step2 and change, without save.
4. not go back to step2 but to step1 or step4 and then go to step 2 to confirm.
5. it leads me to step3 but this blocker is still there.

src/aiidalab_qe/app/configuration/__init__.py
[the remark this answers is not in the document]
superstar54: I used `saved` before, but since the button is `confirm`, and the text will be strange with both `saved` and `confirm` in the code, but mean the same thing.

src/aiidalab_qe/app/main.py
```diff
@@ -18,13 +19,16 @@
 class App(ipw.VBox):
     """The main widget that combines all the application steps together."""
 
+    _submission_blockers = tl.List(tl.Unicode())
```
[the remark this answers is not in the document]
superstar54: Here needs to be a trait, because we need to link it to the `_app_submission_blocker` in the submit_step.

src/aiidalab_qe/app/configuration/__init__.py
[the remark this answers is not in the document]
superstar54: > It is an attribute that be called by other widgets, I think it will be better to make it a trait.

I think `trait` is not needed here, it is static data and we don't need to synchronize it with others.

superstar54:
Thanks @unkcpz for reviewing the code. 

> Some requests and I find some small issues in the behavior, here is how to reproduce:

This is fixed.

For our comment on `trailt`. I kind of disagree. Please give a more clear description of why you think the change is needed.

superstar54:
After a discussion with @unkcpz , we decided to make the following changes:
- only when entering the submit step, do we update the blocker messages
- select a new structure will partially reset the structure step, thus enabling the `confirm` button

We also agreed that in the submit step, updating the blocker_message inside the `_update_date` method is not a good way. I tried to separate it from the `_update_date` method, but I found that several other codes, e.g. `qe_setup_status` and `sssp_installation_status` also depend on this, and some logic and more codes need to update to make this happens,  thus better to do this in another PR in the future. Anyway, for this PR, the current code works without a problem.

@unkcpz Please have another check.

unkcpz:
Emm... no need to make complex changes actually. You can have a look at https://github.com/aiidalab/aiidalab-qe/pull/539/commits/5a775ea4fa6f82d495182ec65486f8d33817aba2 how the logic can be improved for the submission blocker message by separate the internal blocker and external blocker.
Can you adopt the changes from https://github.com/aiidalab/aiidalab-qe/pull/539 to this PR and fix the unit test?

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -61,6 +61,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     previous_step_state = tl.UseEnum(WizardAppWidgetStep.State)
     input_parameters = tl.Dict()
     _submission_blockers = tl.List(tl.Unicode())
+    _outside_submission_blockers = tl.Dict()
```
unkcpz, line 64: In fact, my "picky" review starts from here, I smell a not-good implementation because the object with the same functionality has a different type, it will subtlely cause the problem and make future maintenance not easy.

src/aiidalab_qe/app/main.py
```diff
@@ -85,14 +86,35 @@ def __init__(self, qe_auto_setup=True):
...
         with self.structure_step.hold_sync():
             if (
                 self.structure_step.confirmed_structure is not None
                 and self.structure_step.confirmed_structure != change["new"]
             ):
-                self._wizard_app_widget.reset()
+                self.structure_step.confirmed_structure = None
+
+    def _observe_selected_index(self, change):
+        """Check unsaved change in the step when leaving the step."""
```
unkcpz, line 103: I did some iron-up for this method in https://github.com/aiidalab/aiidalab-qe/pull/539, can you adopt it as well?

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -138,10 +139,12 @@ def __init__(self, qe_auto_setup=True, **kwargs):
             ]
         )
 
-    @tl.observe("_submission_blockers")
+    @tl.observe("internal_submission_blockers", "external_submission_blockers")
     def _observe_submission_blockers(self, change):
-        if change["new"]:
-            fmt_list = "\n".join((f"<li>{item}</li>" for item in sorted(change["new"])))
+        """Observe the submission blockers and update the message area."""
+        blockers = self.internal_submission_blockers + self.external_submission_blockers
```
unkcpz, line 145: Is that possible that blockers are `None` then you cannot use `+`. 
The `change` is not used, please use `_` in as argument.

unkcpz (changes requested):
Thanks @superstar54, some more requests. Should be all fine then, I'll give it a through test then.

src/aiidalab_qe/app/main.py
superstar54: Thanks. I adopted it.

src/aiidalab_qe/app/submission/__init__.py
superstar54: Thanks for the suggestion. I change it to `_change`. The blockers are always `list`, I think we don't need to check `None` here.

src/aiidalab_qe/app/main.py
```diff
@@ -85,14 +86,37 @@ def __init__(self, qe_auto_setup=True):
...
+        with self.submit_step.hold_sync():
```
unkcpz, line 104: I think this is not needed, only at very end the `submit_step` is updated.

src/aiidalab_qe/app/submission/__init__.py
unkcpz: If the variable is not used usually we use `_` in widget callback.

unkcpz (approved):
Thanks! Looks good.

PR #545, opened 2023-11-02
Feature/result multiple plugins

unkcpz:
Rebase PR #522 to main, supports
- showing the result from multiple plugins.
- only shows the result panel when it is available.

One can use the workchain_labels to specify which plugins (outputs) are needed for this result panel. For example, the `electronic_structure` result panel needs `pdos` and `bands`.
```python
class Result(ResultPanel):
    title = "Electronic Structure"
    workchain_labels = ["bands", "pdos"]
```


unkcpz:
Rebased to main, please have a look and test @superstar54.

superstar54 (approved):
I did a quick test and also compared it with #522. LGTM!

Issue #528, opened 2023-10-22
Some layout issues that deteriorate UX


unkcpz:
The tabs are out of box in some panels:

<img width="589" alt="image" src="https://github.com/aiidalab/aiidalab-qe/assets/8831685/2b1f747f-abaa-4ed2-9336-c6c10bf1f2ee">

- The "Band structure settings" do not show properly in the tab. 
- "Pdos Settings" -> "PDOS settings"

In result panel:

<img width="734" alt="image" src="https://github.com/aiidalab/aiidalab-qe/assets/8831685/60ddb836-562e-453a-aa1b-d811845232a4">

- The "Electronic structure (n/a)" is overbox.
- "** not available." -> "The calculation of ** is still running, please wait and the result will show after the calculation is finished."

@superstar54 could you have a look at this?

superstar54:
The result panel is fixed in #522 , which only shows the result panel when it is available.

PR #555, opened 2023-11-09
fix layout of the setting tab title for bands and pdos

superstar54:
forward to #528 

Screenshot from 2023-11-09 11-20-18


unkcpz (commented):
Do we want to call it "Workflow settings" for the first tab? In order to align with the "Advanced Settings"?

unkcpz (approved):
Thanks! @superstar54 

We discussed offline and agreed on changing the tab name from "Workflow" to "Basic settigs".

edan-bainglass:
Why not "Workflow settings" as Jason suggested?

PR #556, opened 2023-11-09
use the function itself instead of the decorated calcfunction

### LINKED ISSUE #488: HTML kpoints mesh
We should re-consider not using the `create_kpoints_from_distance` an instead use one that is not a calc_function
because everytime we change the kpoints_distance it creates an aiida_node 

Screenshot 2023-09-25 at 15 20 24

superstar54:
fix #488 .
To avoid creating an aiida node every time we change the kpoints_distance, we use the function itself instead of the decorated calcfunction.


AndresOrtegaGuerrero (commented):
Nice i didnt know we could use it like that!

superstar54:
AiiDA keeps the original function in the `process_class`, so that it can call this function later in the `run` method.

AndresOrtegaGuerrero (approved):
LGTM!

PR #557, opened 2023-11-10
Add more QE codes with the `calcjob` entry point

superstar54:
In case that a plugin requires codes other than `pw`, `dos` and `projwfc`. This PR adds all QE codes with the `calcjob` entry point in the `aiida-quantumespresso`.


src/aiidalab_qe/common/setup_codes.py
```diff
@@ -25,7 +25,22 @@
     ".conda", "envs", f"quantum-espresso-{QE_VERSION}"
 )
 
-CODE_NAMES = ("pw", "projwfc", "dos")
+CODE_NAMES = (
```
unkcpz, line 28: Can you add a comment that these are codes having calcjobs in aiida-quantemespresso? There are more executable in QE but should at the moment not in here as we discussed.

superstar54: Done. Thanks for the suggestion.

unkcpz (approved):
Looks all good, thanks!

PR #559, opened 2023-11-20
Register plugin node viewer

superstar54:
This PR allows the result panel (as an AiiDA node viewer) for the plugin to show when the user selects the plugin's workchain process in the node tree. 

Blocked by [aiidalab/aiidalab-widgets-base/issues/540](https://github.com/aiidalab/aiidalab-widgets-base/issues/540)

Here is an example: I selected the `Pdos` plugin process, which will trigger the plugin's result panel.

Screenshot from 2023-11-20 14-47-00

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = aiidalab_qe
-version = 23.10.0b2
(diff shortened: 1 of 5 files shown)


superstar54:

PR #560, opened 2023-11-20
Store already added node viewer and load it when needed

superstar54:
In aiidalab-widgets-base/viewer.py, it creates a new `viewer` every time we select a new AiiDA node, and the viewer is not saved in the app, so we can not reuse it later. This has two drawbacks:
- The performance is back. user needs to wait every time when selecting a new node.
- One can not call the viewer later, e.g., for `reset`.

This PR stores the already added viewer and reuses it later.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = aiidalab_qe
-version = 23.10.0b2
+version = 23.11.0rc0
 description = Package for the AiiDAlab QE app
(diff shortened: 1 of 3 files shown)


unkcpz:
@superstar54 this is ready for review? Can you update the PR.

For the performance, can you give it a benchmark?

superstar54:
Will create another PR from my forked repo.

PR #561, opened 2023-11-20
Show result panel immediately when available. 

superstar54:
This PR supports showing the result panel immediately when the plugin's result is available.  The plugin result panel reads the result from the plugin's workchain process directly.

Still need to fix the failed workchain test.


superstar54:

PR #564, opened 2023-11-20
revert some codes setup to accelarate container loading

unkcpz:


unkcpz:
Hi @superstar54, I find every code setup takes 1-2 seconds, so I propose to only keep those that will be added in the foreseeable future.

superstar54:
Hi @unkcpz thanks for sharing the information. I am a little bit surprised that it takes such a long time to set up a code. I think most of the time is used to set up the `verdi` command and profile. The Python code to set up a code should not take so long.

I tried to combine the setup of all codes as one and test the time, and it is fast. Will make a PR soon.

unkcpz:
Close for the moment, it will either be replaced by #565 or just become less problematic after we move to aiida-core 2.5.0 with performance improvement.

PR #565, opened 2023-11-21
Improve setup code performance

superstar54:
Reported by @unkcpz in #564  that, it takes 1-2 sec to set up a code, and we have 14 codes, so around 20 sec in total. Which is not good, and it takes too long to load the container.
We can remove some of the codes, and only keep the ones that will used in the present plugin, as suggested in #564 . There are still 6 codes, which need around 10 secs. 

I looked into the issue and found that most of the time is used to set up the verdi command, profile and database. Thus it's would be good to combine all code set up as one.

Since the code setup is run in a different thread, thus, using the `InstalledCode` directly will fail when storing the code. This PR uses the `python -c` command to run in the `subprocess`. 
The total time to set up all 14 codes now reduced to 2-3 secs.

### DIFF
--- a/src/aiidalab_qe/common/setup_codes.py
+++ b/src/aiidalab_qe/common/setup_codes.py
@@ -79,40 +79,59 @@ def codes_are_setup():
     return all(_code_is_setup(code_name) for code_name in CODE_NAMES)
 
 
-def _setup_code(code_name, computer_name="localhost"):
+def _generate_header_to_setup_code():
+    """Generate the header string to setup a code for a given computer."""
+    header_code = """
+from aiida.orm.nodes.data.code.installed import InstalledCode
+from aiida.orm import load_computer
+from aiida import load_profile
+load_profile()
+
+"""
+    return header_code
+
+
+def _generate_string_to_setup_code(code_name, computer_name="localhost"):
+    """Generate the python string to setup a code for a given computer."""
     try:
         load_code(f"{code_name}-{QE_VERSION}@localhost")
+        return ""
     except NotExistent:
-        run(
-            [
-                "verdi",
-                "code",
-                "create",
-                "core.code.installed",
-                "--non-interactive",
-                "--label",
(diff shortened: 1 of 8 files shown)


src/aiidalab_qe/common/setup_codes.py
```diff
@@ -79,40 +79,59 @@ def codes_are_setup():
...
+        return python_code
 
 
 def setup_codes():
+    python_code = _generate_header_to_setup_code()
     for code_name in CODE_NAMES:
-        _setup_code(code_name)
+        python_code += _generate_string_to_setup_code(code_name)
+    try:
+        run(["python", "-c", python_code], capture_output=True, check=True)
```
unkcpz, line 132: I just realized this change might be a bit flaky because it is not very common way we setup codes in aiida. 
It looks still like a workaround. The ideal solution will be calling the API to setup the codes instead of creating a script and run it with `subprocess.run`. 

On the other hand, we already know that the performance issue comes from every time calling `verdi` it takes time to import the modules which improved in `v2.5.0` and will be released soon. 

Let's hold the solution of this PR or maybe you try using calling API directly?

unkcpz, line 132: I just realized this change might be a bit flaky because it is not very common way we setup codes in aiida. 
It looks still like a workaround. The ideal solution will be calling the API to setup the codes instead of creating a script and run it with `subprocess.run`. 

On the other hand, we already know that the performance issue comes from every time calling `verdi` it takes time to import the modules which improved in `v2.5.0` and will be released soon. 

Let's hold the solution of this PR or maybe you try using calling API directly?

unkcpz (changes requested):
@superstar54 thanks! Please see my comment above, I propose to hold this PR at the moment to see if there are better solutions and see if `aida-core==2.5.0` will fix the performance issue.

superstar54:

PR #566, opened 2023-11-21
Enhanced computational resource widget with resource setup

superstar54:
QeApp allows a plugin to add a new computational code in the submission step. As the number of codes grows, the old `Resouces` and `Parallelization` sections are not proper anymore because the users may want to set different `resources` (e.g., nodes, cpus) for different codes.
There are also such discussions before, e.g., see this [comment](https://github.com/aiidalab/aiidalab-qe/pull/487#:~:text=We%20had%20such%20a%20discussion%20before%2C%20about%20adding%20different%20resources%20for%20pw%2C%20dos%2C%20etc.%20You%20can%20mention%20this%20in%20the%20AiiDAlab%20meeting%2C%20and%20good%20if%20you%20have%20time%20to%20work%20out%20a%20prototype.)

This PR creates a new `QEappComputationalResourcesWidget` from one of the `aiidalab-widigets-base`. The new widget supports setting the resource for the selected code.

Besides, some settings are special to a code, for example, the parallelization levels (pools, images, nk, etc.) in the pw.x code. Thus, this PR creates a new `PWscfWidget` for the `pw.x`, which is inherited from the `QEappComputationalResourcesWidget`.

Here is the screenshot:
Screenshot from 2023-11-21 16-55-12


superstar54:
@AndresOrtegaGuerrero @mikibonacci , It would be great if you could test this PR in your plugins.

@mbercx , do you think we still need to add the `number of k-pools` here? Does the new QE version support setting the parallelization automatically?

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero @mikibonacci , It would be great if you could test this PR in your plugins.
> 
> @mbercx , do you think we still need to add the `number of k-pools` here? Does the new QE version support setting the parallelization automatically?

I think the k-pools should be somehow controlled, for example if you want to run the Berry phase code in pw.x you cant use pools. What it could be nice is to have a dummy run. , or a method that given the structure and the input parameters, it can tell you how many k-point you will have, like that we can provide the user a suggestion for the pools to use

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -400,53 +344,49 @@ def _create_builder(self) -> ProcessBuilderNamespace:
...
-                    }
-                    # Continue to the next item to avoid overriding the resources in the
-                    # recursive `update_builder` call.
-                    continue
-                if k == "resources":
-                    buildy["resources"] = resources
-                else:
-                    self._update_builder(v, max_mpi_per_pool)
+    def _update_builder(self, builder, codes):
+        """Update the resources and parallelization of the ``relax`` builder."""
```
AndresOrtegaGuerrero, line 360: This update  should also affect the other plugins right ?

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,141 @@ def _select_periodicity(self, _=None):
...
+                ),
+            ]
+        )
+
+    def reset(self):
+        """Reset the parallelization settings."""
+        self.npool.value = 1
+
+
+class PWscfWidget(ComputationalResourcesWidget):
```
AndresOrtegaGuerrero, line 726: Maybe add docstring in this one

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,141 @@ def _select_periodicity(self, _=None):
         new_structure.set_pbc(periodicity_options[self.periodicity.value])
         self.structure = None
         self.structure = deepcopy(new_structure)
+
+
+class ComputationalResourcesWidget(AiiDACodeWidget):
+    nodes = traitlets.Int(default_value=1)
```
AndresOrtegaGuerrero, line 624: should it be docstring for the nodes and cpus traitlets ?

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -31,12 +32,35 @@ def check_codes(pw_code, dos_code, projwfc_code):
...
+    builder.projwfc.settings = orm.Dict(dict={"cmdline": ["-nk", str(npool)]})
+
+
 def get_builder(codes, structure, parameters, **kwargs):
     from copy import deepcopy
 
-    pw_code = codes.get("pw")
-    dos_code = codes.get("dos")
-    projwfc_code = codes.get("projwfc")
+    pw_code = codes.get("pw")["code"]
```
AndresOrtegaGuerrero, line 61: Would this affect how you call the codes in the workchain of external plugin?

src/aiidalab_qe/app/submission/__init__.py
superstar54: If other plugins use the same `pw` code as the `relax` workchain, yes.

src/aiidalab_qe/plugins/pdos/workchain.py
superstar54: Yes, I have updated the doc.

src/aiidalab_qe/common/widgets.py
superstar54: thanks. I added the docstring.

mbercx:
> do you think we still need to add the number of k-pools here? Does the new QE version support setting the parallelization automatically?

New QE versions (I forgot from which one, but def v7.2) indeed support setting k-pools automatically.

> I think the k-pools should be somehow controlled, for example if you want to run the Berry phase code in pw.x you cant use pools.

Hmm, I'm not familiar with this use case, but give running on of these calculations a try _without_ specifying k-points and see if the header mentions k-point parallelization. I hope the automated k-pool configuration won't be used in this case then. ^^

> What it could be nice is to have a dummy run. , or a method that given the structure and the input parameters, it can tell you how many k-point you will have, like that we can provide the user a suggestion for the pools to use

I haven't tried this, but figuring out the number of k-points can be a bit tricky since you need to determine the symmetries _exactly_ like Quantum ESPRESSO does. A dummy run is useful in some cases (e.g. we do this to figure out the number of q-points in the [`PhParallelizeQpointsWorkChain`](https://github.com/aiidateam/aiida-quantumespresso-ph/blob/3e99a1ef776882e1a543727b0b816add386f20d2/src/aiida_quantumespresso_ph/workflows/ph/parallelize_qpoints.py#L38)), but I'm not sure it's worth it for `pw.x` k-points because typically you have so many it doesn't matter that much.

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,143 @@ def _select_periodicity(self, _=None):
...
+
+class QEAppComputationalResourcesWidget(AiiDACodeWidget):
+    nodes = traitlets.Int(default_value=1)
+    cpus = traitlets.Int(default_value=1)
+
+    def __init__(self, max_num_nodes=1000, **kwargs):
+        """Widget for the selection of compute resources.
+        max_num_nodes: maximum number of nodes allowed.
+        """
+        self.num_nodes = ipw.BoundedIntText(
```
mikibonacci, line 631: In principle, same code for different step may require different resources. 
In my opinion this can be achieved by defining multiple QEAppComputationalResourcesWidget, e.g. one for relax, one for bands (where we may want to add the -nband...).

So we may have PwRelaxWidget, PwBandsWidget... do you think it is the right way to do it? In this way multiple codes which share the same executable (same aiida node) may be used differently by different plugins. I am thinking about for example a case in which we start from a primitive cell, then we compute something for a big supercell. Using the same resources for the simulations would be not efficient

AndresOrtegaGuerrero: @superstar54 @mikibonacci , there could be many examples , like in the vibroscopy plugin, the PhononWorkChain can use npools, but the Dielectric cannot , or for example in the case of the one i am developing for spin orbit coupling, i might need more resources than the ones use for PwRelax (with out SOC) , though it might create a lot of instances in the step 3

src/aiidalab_qe/common/widgets.py
```diff
@@ -17,7 +17,8 @@
 import traitlets
 from aiida.orm import CalcJobNode
 from aiida.orm import Data as orm_Data
-from aiida.orm import load_node
+from aiida.orm import load_code, load_node
+from aiidalab_widgets_base import ComputationalResourcesWidget as AiiDACodeWidget
```
unkcpz, line 21: The name `AiiDACodeWidget` easily get confused with `AiiDACodeSetup` in AWB: https://github.com/aiidalab/aiidalab-widgets-base/blob/aba0bdb3dc7a7e1af268145a0fd643194e681523/aiidalab_widgets_base/computational_resources.py#L1086 which is for detailed code setup specificly. 

I didn't see the point of renaming the class.

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,143 @@ def _select_periodicity(self, _=None):
...
+            value=1, step=1, min=1, description="CPUs", width="10%"
+        )
+        super().__init__(**kwargs)
+        # add nodes and cpus into the children of the widget
+        self.children[0].children += (
+            self.num_nodes,
+            self.num_cpus,
+        )
+
+    @traitlets.observe("value")
```
unkcpz, line 644: I know this `value` comes from `ComputationalResourceWidget` but the interface is not clearly exposed. 
Instead of inherit from `ComputationalResourceWidget`, this case fit the composite object quite well. Please consider using `ComputationalResourceWidget` as sub-widget in this widget you defined.

unkcpz (changes requested):
If you don't override the methods of a class to create a new class, but simply extend it with more methods, please don't use class inherit. 
If I understand correctly, there is no modification of `ComputationalResourceWidget`, but add two more widgets on top of it. So why not using composite widget design? Am I miss something?

AndresOrtegaGuerrero:
> > do you think we still need to add the number of k-pools here? Does the new QE version support setting the parallelization automatically?
> 
> New QE versions (I forgot from which one, but def v7.2) indeed support setting k-pools automatically.
> 
> > I think the k-pools should be somehow controlled, for example if you want to run the Berry phase code in pw.x you cant use pools.
> 
> Hmm, I'm not familiar with this use case, but give running on of these calculations a try _without_ specifying k-points and see if the header mentions k-point parallelization. I hope the automated k-pool configuration won't be used in this case then. ^^
> 
> > What it could be nice is to have a dummy run. , or a method that given the structure and the input parameters, it can tell you how many k-point you will have, like that we can provide the user a suggestion for the pools to use
> 
> I haven't tried this, but figuring out the number of k-points can be a bit tricky since you need to determine the symmetries _exactly_ like Quantum ESPRESSO does. A dummy run is useful in some cases (e.g. we do this to figure out the number of q-points in the [`PhParallelizeQpointsWorkChain`](https://github.com/aiidateam/aiida-quantumespresso-ph/blob/3e99a1ef776882e1a543727b0b816add386f20d2/src/aiida_quantumespresso_ph/workflows/ph/parallelize_qpoints.py#L38)), but I'm not sure it's worth it for `pw.x` k-points because typically you have so many it doesn't matter that much.

I think the number of k-points ("having so many") can be debatable when using a system with a big unit cell

mbercx:
> I think the number of k-points ("having so many") can be debatable when using a system with a big unit cell

Sure, that's true. But anyways QE should now be able to figure out how to parallellize over the k-points itself, so adding a dummy run to any of the `aiida-quantumespresso` work chains is currently not on the agenda.

AndresOrtegaGuerrero:
> > I think the number of k-points ("having so many") can be debatable when using a system with a big unit cell
> 
> Sure, that's true. But anyways QE should now be able to figure out how to parallellize over the k-points itself, so adding a dummy run to any of the `aiida-quantumespresso` work chains is currently not on the agenda.

Even for the pools? QE can now do this?

mbercx:
> Even for the pools? QE can now do this?

Yup ^^. Give it a try with QE v7.2

unkcpz:
Update on the PR, @mikibonacci has a use case require this feature. So we need to try to get it merged ASAP.
@superstar54 and I had a in person quick chat, we agree on combine the widgets instead of using class inherit might be better solution.

AndresOrtegaGuerrero:
@unkcpz This PR is also important for #577 since something the dos.x and projwfn.x require at least 2 nodes when using DFT+U (if a system is big)

unkcpz:
Thanks @AndresOrtegaGuerrero for the remind. 

@superstar54, can you update the PR and address the comments that not answered? It is also worth to check the new default QE feature for the npool setting as mentioned in by @AndresOrtegaGuerrero above. We can have a chat this week to see what should be in this PR and what should do after to not block it to be merged.

superstar54:
Thanks all for the review. I made changes based on your suggestions:

1. Use composite widget design for `QEAppComputationalResourcesWidget`. It now use `ComputationalResourcesWidget` as a sub-widget.
2. Add an override box for `parallelization` so that advanced users can set the `number of k-pools` for `pw`; otherwise, it will be determined by QE if not set. 


> same code for different step may require different resources.. ...So we may have PwRelaxWidget, PwBandsWidget... do you think it is the right way to do it?

For the moment, I suggest not adding this because our users may be confused, otherwise, there will be a lot of codes. Users may not be confused about which codes are for which. I think they don't need to understand the steps of the workchain. Users can set the max (cpus) resource anyway for all steps.

Do you think we also need to add this `parallelization` (`npool`) for the `projwfc`? For the moment, I disabled this options.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -5,24 +5,21 @@
...
 from aiida.engine import ProcessBuilderNamespace, submit
-from aiidalab_widgets_base import ComputationalResourcesWidget, WizardAppWidgetStep
+from aiidalab_widgets_base import WizardAppWidgetStep
 from IPython.display import display
 
 from aiidalab_qe.app.parameters import DEFAULT_PARAMETERS
 from aiidalab_qe.app.utils import get_entry_items
 from aiidalab_qe.common.setup_codes import QESetupWidget
 from aiidalab_qe.common.setup_pseudos import PseudosInstallWidget
+from aiidalab_qe.common.widgets import PWscfWidget, QEAppComputationalResourcesWidget
```
unkcpz, line 20: The name makes more sense if we call:
- `PWscfWidget` -> `PwCodeResourceSetupWidget`
- `QEAppComputationalResourcesWidget` -> `ComputationalResourceSetupWidget` (Mind with the "s", I think it is good to use singular format.)

The naming in AWB for `ComputationalResourcesWidget` might not ideal, and we can probably change it to `ComputationalResourceSelctionWidget`.

unkcpz (changes requested):
Me and @superstar54 discussed in person on the PR, and now I get most of the idea of the changes. 
Another small milestone to the fully plugin design of the QeApp, thanks a lot for the great work @superstar54 

I only have a small request on the naming convention and probably also add a small test for the widget that when the checkbox ticked you test the k-point override is shown. 

@mikibonacci, can you give these change a test for the code setup you need?

AndresOrtegaGuerrero:
@superstar54  I think this PR should include an option for `--ntasks-per-node` , when using QE 7.2 with gpu, this parameters becomes quite important , I have been experiencing this issue. So please provide this extra option for the `pw.x`

AndresOrtegaGuerrero:
@superstar54 Currently in the app, `dos.x `and `projwfn.x ` , use the same number of resources assigned to `pw.x` , as well as the `npools`

superstar54:
> Currently in the app, dos.x and projwfn.x  , use the same number of resources assigned to pw.x , as well as the npools

What's would be the problem, if we add `--ntasks-per-node` and `--cpus-per-task`?

AndresOrtegaGuerrero:
> > Currently in the app, dos.x and projwfn.x  , use the same number of resources assigned to pw.x , as well as the npools
> 
> What's would be the problem, if we add `--ntasks-per-node` and `--cpus-per-task`?

There is none, at the moment these options arent available, or at least you dont have control over it 
For example imagine in one job you would like this input 
#SBATCH --job-name="qe-gpu"
#SBATCH --nodes=6
#SBATCH --ntasks-per-node=2
#SBATCH --cpus-per-task=6
#SBATCH --constraint=gpu

but maybe in another one you would like 

#SBATCH --job-name="qe-gpu"
#SBATCH --nodes=6
#SBATCH --ntasks-per-node=3
#SBATCH --cpus-per-task=4
#SBATCH --constraint=gpu

superstar54:
Hi @AndresOrtegaGuerrero , I added the support for `--ntasks-per-node` and `--cpus-per-task`, please try and review the PR. Here is a demo:

https://github.com/aiidalab/aiidalab-qe/assets/11457659/e23e045f-fad1-4278-9b7c-666d98b5cecf

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,256 @@ def _select_periodicity(self, _=None):
...
+            self.npool.layout.display = "block"
+        else:
+            self.npool.layout.display = "none"
+
+    def reset(self):
+        """Reset the parallelization settings."""
+        self.npool.value = 1
+
+
+class PWscfWidget(QEAppComputationalResourcesWidget):
```
AndresOrtegaGuerrero, line 835: One question Xing, if i have an external plugin, can i set an independent pw.x for it ? or the pw.x is for all the plugins?

src/aiidalab_qe/app/submission/__init__.py
superstar54: Thank you for the suggestion. I've renamed it to `PwCodeResourceSetupWidget`. Regarding the name `ComputationalResourceSetupWidget`, I'll wait for the changes in AWB to be finalized before making updates here.

superstar54:
> I only have a small request on the naming convention and probably also add a small test for the widget that when the checkbox ticked you test the k-point override is shown.

Hi @unkcpz , thanks for the review. I changed the name as you suggested and added a test for the new widget.

src/aiidalab_qe/common/widgets.py
superstar54: yes!

AndresOrtegaGuerrero (approved):
LGTM! , just be sure to communicate to the external plugins, so we make the change for the release . Great work !

unkcpz (commented):
Thanks @superstar54. I think I am late for the party.

Just add two small comments.

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -31,12 +32,40 @@ def check_codes(pw_code, dos_code, projwfc_code):
...
+    builder.nscf.pw.parallelization = orm.Dict(dict=codes["pw"]["parallelization"])
+    builder.dos.metadata.options.resources = {
+        "num_machines": codes.get("dos")["nodes"],
+        "num_mpiprocs_per_machine": codes.get("dos")["ntasks_per_node"],
+        "num_cores_per_mpiproc": codes.get("dos")["cpus_per_task"],
+    }
+    builder.projwfc.metadata.options.resources = {
+        "num_machines": codes.get("projwfc")["nodes"],
+        "num_mpiprocs_per_machine": codes.get("projwfc")["ntasks_per_node"],
+        "num_cores_per_mpiproc": codes.get("projwfc")["cpus_per_task"],
```
unkcpz, line 56: This part seems a bit verbose and codes are repeating, consider to simplify it a bit?
- `builder.scf.pw.metadata.options.resources` five dots attributes call looks frightening and I think it makes plugin developer easy to make mistakes.
- For override the resources, maybe there already functions to do this in aiida-quantumespresso? @mbercx  Maybe you know it?

Somehow my previous comment on this was eaten by GitHub. I do it again.

src/aiidalab_qe/common/widgets.py
```diff
@@ -620,6 +621,259 @@ def _select_periodicity(self, _=None):
...
+        """Set the parameters."""
+        self.code_selection.value = parameters["code"]
+        if "nodes" in parameters:
+            self.num_nodes.value = parameters["nodes"]
+        if "cpus" in parameters:
+            self.num_cpus.value = parameters["cpus"]
+        if "ntasks_per_node" in parameters:
+            self.resource_detail.ntasks_per_node.value = parameters["ntasks_per_node"]
+        if "cpus_per_task" in parameters:
+            self.resource_detail.cpus_per_task.value = parameters["cpus_per_task"]
```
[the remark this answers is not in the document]
superstar54: Thanks for your comment. It's a common way to use the `parameters.setter` interface, and then in another function `set_parameters` implement the real thing. In this pattern, the derived classes can easily modify or extend the setting logic by overriding a dedicated method, without needing to alter the property's interface. As you can see in the `PwCodeResourceSetupWidget` class.

src/aiidalab_qe/plugins/pdos/workchain.py
superstar54: Thanks for the suggestion. Indeed, one can use the `options` to set resources in the `get_builder_from_protocol`. and I agree that this should be the recommended way. I opened an [issue](https://github.com/aiidalab/aiidalab-qe/issues/682) on this.

src/aiidalab_qe/common/widgets.py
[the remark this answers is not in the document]
unkcpz: Where you find it "common"? I am fine with it, just seems the `get_` and `set_` are used and the property maybe redundant.

PR #567, opened 2023-11-22
Use AiiDA Monitor

superstar54:
This is an attempt to use AiiDA Monitor inside AiiDAlab. So the user can monitor the real-time job status (energy, trajectory). It's a good use case for the new AiiDA feature: Monitor.

### Technique details:
- A monitor is added to the pw calcjob of the relax workchain. It reads the SCF energies and the trajectory from the `aida.out` file, then saves the date in the `node.base.extras`.
- In the AiiDAlab-QE result panel, open a new `thread` to read data from the `node.base.extras` and update the widiget.

Here is the screencast

https://github.com/aiidalab/aiidalab-qe/assets/11457659/46d05b7f-4303-42b0-a8a1-803ac525a08d


superstar54:

PR #568, opened 2023-11-22
Removing hardcoded values of NSCF

AndresOrtegaGuerrero:
This PR removes the values from the kpoints_distance from the pdos plugin
This PR also fixes the behavior of create_kpoints_from_distance creating a new aiida node


superstar54 (approved):
LGTM!

PR #570, opened 2023-11-22
Confirm structure and configuration steps when loading previous result

### LINKED ISSUE #569: Go to the submit step will reset previous steps
**Describe the bug**
I loaded a previous calculation, and then went to the submit step. Then the status of the previous steps is reset. And show 

**Expected behavior**
Should not be reset in this case.

**Additional context**
I believe that this is introduced by #537

superstar54:
fix #569


unkcpz (approved):
Thanks! @superstar54 
The changes make sense, I didn't give it a test myself but I trust you did it locally.

PR #572, opened 2023-11-29
DOC: new docker stack update and more

unkcpz:


unkcpz:
I merged this for the moment, and we'll anyway give it as the final go of all text in the documentation.

PR #576, opened 2023-12-05
providing a subsection in developer guide docs, about automatic code installation

mikibonacci:


unkcpz:
Cool thanks @mikibonacci
It looks good to me, but maybe we can wait for https://github.com/aiidalab/aiidalab-qe/pull/565, @superstar54 did some changes on using AiiDA python API instead of calling `verdi` to speed it up. The performance is not a problem for phonopy installation, but I think it is better to have a consistent method to use for new code setups.

mikibonacci:
yep, fine for me to wait :) 

thanks

mikibonacci:
What I should add (from aiidalab meeting discussions):

- [ ] actually, a entry point in aiidalab_qe.console_scripts, ore something similar. Anyway, a standardized way for all plugins (and document it of course);
- [ ] the script should be something that it can be triggered also from the GUI (functionality not yet implemented in the GUI). So, forward-compatibility;

docs/source/development/plugin.rst
```diff
@@ -354,11 +354,89 @@ Here is the example of the built-in `pdos` plugins with codes `dos.x` and `projw
     }
 
 For the moment, the app does not support set up the newly added codes automatically.
-Thus, the user needs to set up the codes manually.
+Thus, the user needs to set up the codes manually. However, the developer can provide console
+scripts to make user life easier. Please read to following section for some instructions on how
+to do it.
```
unkcpz, line 359: We try to break lines for a full sentence to make it easy to review and for future change. Do need to worry about a long line (if it is super long, then probably means it can be cut to short sentences for readability).

docs/source/development/plugin.rst
```diff
@@ -354,11 +354,89 @@ Here is the example of the built-in `pdos` plugins with codes `dos.x` and `projw
...
+    phonopy_install = "phonopy_install:install_phonopy"
+
+When the user, after the package installation, will execute the command
+
+.. code-block:: console
+
+      phonopy_install
+
+this will invoke the function `install_phonopy`, contained in the phonopy_install.py file in the
+root directory of the package.
```
unkcpz, line 380:
```suggestion
this will invoke the function `install_phonopy`, contained in the `phonopy_install.py` file in the root directory of the package.
```

unkcpz (changes requested):
After a second look at https://github.com/aiidalab/aiidalab-qe/pull/565, I think it requires more discussion on it. Therefore I personally don't want to block this PR and propose to go with calling `verdi` command to set the code which we test over many releases and should be robust. 
The performance issue is caused from `verdi` command and will be improved a lot after `aiida-core` v2.5.0 which will soon release. 

@mikibonacci thanks for adding this, I added some minor requests, if you can give it a look I'll do a local test with the method you write here.

PR #577, opened 2023-12-11
Feature DFT+U 

AndresOrtegaGuerrero:
This PR add the feature to the Qe App to run simulations with DFT+U on-site

image

The feature include a HubbardWidget, where the hubbard manifold is only one orbital, 
for example for Co will the the 3d , of for Si it will be the 3p 
(This can modified so the widget display the valence orbital shell , for example Co 3d and 4s) (Or if you prefer the the previous occupied shell like 3p for Co, we can discuss about this :D ) , 

For a second PR:
I would like to have an independent tab for displaying DFT+U results (of interest) like the occupation matrix, and the occupation per atom and spin. I was wondering if this should be implemented as a plugin (Like in electronic_structure plugin) but i think the results should be available independently of the workchain run (bands, pdos, or an ext plugin)


AndresOrtegaGuerrero:
Before i continue to the final steps of the PR (like the independent tab) I was wondering if you @superstar54 and @unkcpz could give me your input on this PR design :D

AndresOrtegaGuerrero:
I have decided that the tab to display the occupation should be better in a second PR, otherwise this one will be too big

unkcpz:
@AndresOrtegaGuerrero thanks! I'll give it a look.

> I have decided that the tab to display the occupation should be better in a second PR

I agree, and since it is for the next release. I think even better steps are using your bandsplot widget to replace the current one first.

unkcpz:
Very cool! I didn't get to too much detail yet. But the first thing I notice is this create widget from an elements pattern which appeared in the magnetization setting and pseudopotential setting. I think the third time it appears to push me to provide a commend interface for the design. I'll try to work on that first.

superstar54 (changes requested):
Thanks for the great work. Two quick comments:
- In the workflow, you need to pass the `hubbard_structure` to the plugin's workchain.
- The GUI is not set correctly when loading a finished job.

AndresOrtegaGuerrero:
@superstar54 @unkcpz  Thank you , I will work on the changes, I did notice something `aiida-quantumespresso` plugin PwBandsWorkChain is incompatible with HubbardStructure data, and also HubbardStructureData doesnt support 1D, 2D periodicity. I am waiting for this PR (https://github.com/aiidateam/aiida-quantumespresso/pull/998) to get approved. Let me know further changes

AndresOrtegaGuerrero:
For this PR to be evaluated we need the following PRs to be merged

- [ ] #566 
- [ ] https://github.com/aiidateam/aiida-quantumespresso/pull/998

tests/configuration/test_advanced.py
```diff
@@ -151,3 +151,72 @@ def test_advanced_kpoints_mesh():
...
+        ["Co", "Co", (0, 0, 0)],
+        ["O", "O", (0, 0, 3.6020728736387)],
+        ["O", "O", (0, 0, 10.201532881212)],
+        ["Li", "Li", (0, 0, 6.9018028772754)],
+    ]
+    structure = orm.StructureData(cell=cell)
+
+    for site in sites:
+        structure.append_atom(position=site[2], symbols=site[0], name=site[1])
+
```
superstar54, line 178: Since this structure may be useful to test other properties, could you move this to `conftest.py` and create a fixture for it.

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,409 @@ def _select_periodicity(self, _=None):
...
+        Returns:
+            list: A list of labels in the format "{kind} - {manifold}".
+        """
+        kind_list = self.input_structure.get_kind_names()
+        hubbard_manifold_list = [
+            self._get_hubbard_manifold(Element(x.symbol))
+            for x in self.input_structure.kinds
+        ]
+        result = [
+            f"{kind} - {manifold}"
```
superstar54, line 728: There is possible risk. If a kind has a name "Co - up", then the label will be "Co - up - 3d", this will raise an error in the Workchain when it tries to get the kind and manifold back.

src/aiidalab_qe/common/widgets.py
```diff
@@ -617,3 +618,409 @@ def _select_periodicity(self, _=None):
...
+class HubbardWidget(ipw.VBox):
+    """Widget for setting up Hubbard parameters."""
+
+    def __init__(self, input_structure=None):
+        self.input_structure = input_structure
+        self.eigenvalues_help = ipw.HTML(
+            value="For transition metals and lanthanoids, the starting eigenvalues can be defined (Magnetic calculation).",
+            layout=ipw.Layout(width="auto"),
+        )
+        self.hubbard = ipw.Checkbox(
```
superstar54, line 632: suggest to change the name, e.g.
```suggestion
        self.activate_hubbard = ipw.Checkbox(
```

superstar54 (changes requested):
Thanks for the great work! There are a few comments, and then I think it is ready to go.
As for the DFT+U+V part, I think we can implement it in a future PR. For the moment, let's merge this PR first.

src/aiidalab_qe/common/widgets.py
AndresOrtegaGuerrero: but usually we dont allow names as Co-up, we allow tags like Co1, or Co2 ,

AndresOrtegaGuerrero: But i guess this should be control @ the structure step then , dont you think ?

tests/configuration/test_advanced.py
AndresOrtegaGuerrero: thank you , i will add it in` conftest.py`

AndresOrtegaGuerrero:
@superstar54 Before merging, we need to request a new release of aiida-quantumespresso plugin, and bump the version here. So it doesn't fail when considering 1D, 2D systems for bands and so on.

src/aiidalab_qe/common/widgets.py
superstar54: Yes, the tags are controlled in the structure step. However, the problem here is the `{kind} - {manifold}` could result in something like `Co-up - 3d` (e.g. if user import data from database, and the kind name is Co-up already), thus it will raise an error in the WorkChain, where you have one line of code 
```
kind, orbital = key.split(" - ")
```
I suggest you split by the last `-` to solve this problem, or you can use a dict with the kind and orbital saved explicitly.

AndresOrtegaGuerrero: @superstar54 so for something like this , `kind, orbital = key.rsplit("-", 1)`?

superstar54: Yes

AndresOrtegaGuerrero: Done, let me know

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -121,8 +122,24 @@ def get_builder_from_protocol(
...
         builder = cls.get_builder()
-        # Set the structure.
-        builder.structure = structure
+        # Set a HubbardStructureData if hubbard_parameters is specified
+        hubbard_dict = parameters["advanced"].pop("hubbard_parameters", None)
+        if hubbard_dict is not None:
+            hubbard_parameters = hubbard_dict["hubbard_u"]
+            hubbard_structure = HubbardStructureData.from_structure(structure)
+            for key, value in hubbard_parameters.items():
+                kind, orbital = key.rsplit("-", 1)
```
superstar54, line 131: There are spaces when creating the label.
```suggestion
                kind, orbital = key.rsplit(" - ", 1)
```

AndresOrtegaGuerrero: Done

superstar54 (approved):
LGTM!

PR #578, opened 2023-12-12
Bump SSSP to 1.3

unkcpz:
- [ ] ~The test fail because the default value will read from `aiida-quantumespresso` where the pseudo version in the protocol not updated yet.~

### DIFF
--- a/docs/source/development/architecture.rst
+++ b/docs/source/development/architecture.rst
@@ -46,7 +46,7 @@ The dictionary has the following structure:
                 },
                 "pseudos": {"Si": "eaef3352-2b0e-4205-b404-e6565a88aec8"},
             },
-            "pseudo_family": "SSSP/1.2/PBEsol/efficiency",
+            "pseudo_family": "SSSP/1.3/PBEsol/efficiency",
             "kpoints_distance": 0.5,
         },
         "bands": {"kpath_2d": "hexagonal"},
--- a/docs/source/development/plugin.rst
+++ b/docs/source/development/plugin.rst
@@ -206,7 +206,7 @@ The `parameters` passed to the `get_builder` function has the following structur
                 },
                 "pseudos": {"Si": "eaef3352-2b0e-4205-b404-e6565a88aec8"},
             },
-            "pseudo_family": "SSSP/1.2/PBEsol/efficiency",
+            "pseudo_family": "SSSP/1.3/PBEsol/efficiency",
             "kpoints_distance": 0.5,
         },
         "bands": {"kpath_2d": "hexagonal"},
--- a/post_install
+++ b/post_install
@@ -2,9 +2,10 @@
 
 # This script is executed after the app installation completes
 # triggered by the post install hook implemented in https://github.com/aiidalab/aiidalab/pull/295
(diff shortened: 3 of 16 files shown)


unkcpz:
As discussed in AiiDA meeting, I move on without the protocol change in aiida-quantumespresso. But it means we need to explicitly override the pseudo_family when creating a builder. 
@superstar54 When I make these changes, I see two things we can improve after we back from Christmas break, 1) the unit tests and fixtures must be refactored, there are too many duplicates and quite a mess. 2) In aiida-quantumespresso, the `ProtocolMixin` is used and there are protocols defined by yaml files for every workchains. We probably use the same way rather than override it in the code, however, I see this against the plugin where the `get_builder` is used and the builder is created out of air. 

Anyhow, I don't want to block the this SSSP bump which Nicola asks for the QeApp explicitly.

unkcpz:
There still couple of tests failed and I have no idea how to fix those, I guess it requires to haveing a proper `parameters` passed to every workchain creator fixture. @superstar54 Can you have a look?

superstar54:
> There still couple of tests failed and I have no idea how to fix those, I guess it requires to haveing a proper `parameters` passed to every workchain creator fixture. @superstar54 Can you have a look?

Hi @unkcpz which one? I saw all the tests passed.

superstar54:
> 1) the unit tests and fixtures must be refactored, there are too many duplicates and quite a mess. 2) In aiida-quantumespresso, the ProtocolMixin is used and there are protocols defined by yaml files for every workchains. We probably use the same way rather than override it in the code, however, I see this against the plugin where the get_builder is used and the builder is created out of air.

1) Indeed, we need to update the tests and fixtures. Initially, when I wrote the fixture for the workchain and plugin-related tests, the concept of the plugin was still in its early stages. Now, with the plugin architecture having stabilized, we could update the test with a better mind.

2) Using ProtocolMixin, one defines all the parameters in the YAML file. When creating the builder, it fetches these parameters based on the `protocol` . However, QEapp has a different way. In QEapp, one can still define the protocol and related parameters in the YAML file. But when the user selects the `protocol` from the GUI, the GUI will fetch the corresponding parameter immediately and show the values to the user. At this stage, the user can modify the GUI to override the parameters. In the final submission step, we read all the parameters from the GUI (the protocol value and the user override values) and use these values to create the builder, so we don't need to read the YAML file in the `get_builder` function again.
Unless we want to define some other protocol-related parameters, that don't have their corresponding value in the GUI, we can read the YAML in the `get_builder` function. By the way, the `get_builder` function is quite flexible. The plugin developer can define the YAML and load it into the `get_builder. If this is a common use case, one can improve the `get_builder` function to standardize it.

   @mbercx  knows better `ProtocolMixin`, feel free to comment.

unkcpz:
> Hi @unkcpz which one? I saw all the tests passed. 

Ha, nice. Some failed in my local environment, I guess I messed up with the environment somehow. Can you give this branch a test?

unkcpz:
Thanks for the very detailed comment, I’ll convert it to an issue for future improvements. For the PR itself, can you give it a look? I want to make another release and build the quantum mobile again on Monday.

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

unkcpz:
I merged this anyway so as not to block the release of Quantum Mobile, in principle I should not merge it without approve.

unkcpz:
Following tests are all good:
- Install the app in a `full-stack` image, the pseudo library and qe are all installed.
- Start from the `23.10.0rc0` where the SSSP version is v1.2, update the app from appstore the pseudo libraries of SSSP 1.3 installed automatically and the app shows the progress bar to block the submission step until the pseudo families installed correctly.
- Start from a bare qe image the SSSP 1.3 is installed and ready to launch a calculation.

Issue #533, opened 2023-10-24
Improve the clarity of unit test fixtures 


unkcpz:
The unit test fixtures are a bit messy at the moment, the performance is not good as well. Some fixtures to create app object and step objects seems redundant. We need to set the app as a session with a clear tear-down process to reset, instead creat a new object and setting parameters from the beginning. 

@superstar54 let's try to set a pair coding session between two of us to tidy things up.

unkcpz:
I think this issue become a high priority after the stable release and more qeapp plugins start shaping. 
Can I assign this to you @superstar54? Since most nice unit tests are brought to the code base with the plugin design.

unkcpz:
Hi @superstar54, can this be assigned to you? I think the plugins would benifit from having testing. Clear fixtures can help developers to writing sustainable plugins.

superstar54:
Here are the catalogs for the test:

- Setting GUI
  - set
  - reset
  - load previous results

- Result GUI
  - mock workchain output


- Workchain
  - main app
  - plugin related

PR #580, opened 2023-12-13
Add Feature: XAS Plugin

PNOGillespie:
# Overview
This PR adds a plugin for calculating X-Ray Absorption (XAS) spectra using the `XspectraCrystalWorkChain` of AiiDA-QE. Currently this is implemented for K-edge calculations of neutral systems. Calculations of systems with hubbard parameters and magnetic configurations can be made possible, though would require a backend change to AiiDA-QE (see https://github.com/aiidateam/aiida-quantumespresso/pull/969 for relevant PR)
XAS_Plugin_Select

## Setting Panel
The `Setting` panel for the plugin allows users to simply select which element to calculate XAS for, as well as tuning the supercell size to be used in the calculation. This panel also provides the option to change the type of core-hole approximation between FCH (removing the excited electron with tot_charge = 1) and XCH (allowing the excited electron to occupy the conduction band. Since it is still an ongoing subject of debate in the literature, we allow the user to change this value - though recommended values for each element are provided, which will also occupy the default value in the dropdown box.
XAS_Plugin_Setting_Panel
Note that only elements for which appropriate pseudopotentials are available will be presented in the panel. See later section for notes about pseudopotentials.
### Overview
The final XAS spectrum is displayed in the `Result` panel. A dropdown box selects which element to display spectra for. The widget will display both the complete powder XAS for the input structure and the contributions from each symmetrically non-equivalent site. Note that all spectra are normalised with respect to area under the peak by default. Spectra from different sites are scaled in intensity according to the ratio of site multiplicity (number of symmetrically-equivalent sites) to total multiplicity (number of sites for the element).
XAS_Plugin_Result_Panel

### Broadening Tools
Broadening effects present in experimental XAS data greatly affect the appearance of the final spectrum (see [Bunau & Calandra (2013)](http://dx.doi.org/10.1103/PhysRevB.87.205105)) and so post-procesing of spectra to include these effects (which cannot be computed ab-initio with DFT) is often used in published works using DFT-derived spectra. We therefore provide a set of broadening tools for the final spectrum. The tools apply one of two Lorentzian broadening schemes (toggled by a checkbox widget):
1. Constant broadening applied to the entire spectrum.
2. Variable-energy broadening using a starting value ($\Gamma_{hole}$), maximum value ($\Gamma_{max}$, and a mid-point energy ($E_{Center}$).

The 1st scheme uses just $\Gamma_{hole}$ (which refers to the natural linewidth of the core level) to define the Lorentzian parameter. The 2nd scheme uses each parameter to define an s-curve function with an inflection point at $E_{Center}$. The broadening scheme is applied to all spectra on the plot simultaneously. Note that when the spectrum is downloaded, the plotted data on the Plotly widget (including broadening) is what is written to the CSV file - not the raw data. To view the raw data, the user can simply set $\Gamma_{hole}$ to 0.
## Pseudopotentials
Calculations of XAS using QE require a pseudopotential to define the electron configuration of the final state of the excitation for the absorbing atom. In addition, a radial coordinate plot of the initial state (e.g. the 1s state of the absorbing atom in a K-edge calculation) is needed by XSpectra. To provide both of these, including a relevant pseudopotential for the ground-state of the absorbing element (and from which we generate the initial state wavefunction) a small pseudopotential library is provided with the plugin. This library is taken from another GitHub repository (https://github.com/PNOGillespie/Core_Level_Spectra_Pseudos) which is a small collection of pseudos which we have used already in XAS calculations. This repo has an archive file for each XC functional (currently only PBE) and a "Pseudo Table of Contents" (`pseudo_toc.yaml`) file to record the filenames and directories for each pseudo and associated core wavefunction file. The plugin uses this YAML file to import files, check for missing files, load the correct nodes for calculation, and set the default core-hole approximation. The only requirement for maintainers is just to copy the YAML file from `Core_Level_Spectra_Pseudos` into the plugin directory in order to update the local pseudo archive (which is done automatically on startup if the archive file is missing or there are pseudos missing which should be there).


src/aiidalab_qe/plugins/xas/pseudo_toc.yaml
```diff
@@ -0,0 +1,26 @@
...
+xas_xch_elements: [C, Li]
+pseudos:
+    pbe:
+        gipaw_pseudos:
+            C: C.pbe-n-kjgipaw_psl.1.0.0.UPF
+            Cu: Cu.pbe-n-van_gipaw.UPF
+            F: F.pbe-gipaw_kj_no_hole.UPF
+            Li: Li.pbe-s-rrkjus-gipaw.UPF
+            O: O.pbe-n-kjpaw_gipaw.UPF
+            Si: Si.pbe-van_gipaw.UPF
```
unkcpz, line 11: I don't want to slow down the progress, is that possible to integrate this to `aiida-pseudo`?

src/aiidalab_qe/plugins/xas/result.py
```diff
@@ -0,0 +1,539 @@
...
+from scipy.interpolate import make_interp_spline
+
+from aiidalab_qe.common.panel import ResultPanel
+
+
+class SpectrumDownloadButton(ipw.Button):
+    """Download button with dynamic content
+    The content is generated using a callback when the button is clicked.
+    Modified from responses to https://stackoverflow.com/questions/61708701/how-to-download-a-file-using-ipywidget-button#62641240
+    """
```
unkcpz, line 20: We have the widget to use in https://github.com/aiidalab/aiidalab-widgets-base/blob/master/aiidalab_widgets_base/export.py

unkcpz (changes requested):
Really cool! Welcome XAS to join the QeApp family. 
I give a broad look with two comments, @superstar54 is capable of giving it a detailed look.

src/aiidalab_qe/plugins/xas/pseudo_toc.yaml
PNOGillespie: So, I will say for this that the solution being used here is not necessarily intended as a permanent solution to providing pseudopotentials for these types of calculations. In principle, integrating some sort of family in `aiida-pseudo` for core-hole pseudopotentials is the ideal long-term objective here - since then the procedure for assigning pseudos obviously follows that of other plugins.

The two issues I see with integrating the current library with `aiida-pseudo` are:
1. Some sort of new `Class` or other new feature will need to be made for `aiida-pseudo` in order to work with these types of pseudopotentials: for instance to ensure the correct core wavefunction data file is matched to the correct GIPAW pseudo, also that the core-hole and GIPAW pseudos are matched correctly.
2. The current library is a stub which covers only some elements, therefore potentially not enough to make it worth pursuing with just what's there at the moment.

I would be interested in your thoughts on this in any case @unkcpz.

src/aiidalab_qe/plugins/xas/result.py
PNOGillespie: I can try it using the AWB version and see how that goes. In principle, it looks like it should be fine since the file for the output spectra is created outside of the widget, but I will let you know if there are any issues with using it here.

unkcpz: Sorry, @PNOGillespie I think what you did are correct, the `export` widget is for exporting the aiida archive, we don't yet have the generic widget for export the file. 
Can you try to it to `aiidalab-qe/src/aiidalab_qe/common/`, @AndresOrtegaGuerrero also implement a similar widget.

PNOGillespie: Sure @unkcpz, I will work on using the `export` widget with the plugin, though I'm currently only a couple of days from vacation, so maybe in January at this rate :laughing:

I had intended it to show explicitly (for the user) that it would download a CSV file - also that the filename be related to the element selected to be displayed (e.g. O_K-edge_XAS.csv for spectra of oxygen). Does the widget allow for the filename/filetype to be specified? @AndresOrtegaGuerrero 

I don't mind working on changes to `export` in order to make this happen, if that's ok with everyone.

AndresOrtegaGuerrero: @PNOGillespie  i think @unkcpz refers to the BandsWidget PR #581 , you can adapt the export (code) to your needs, like the filename name

src/aiidalab_qe/plugins/xas/workchain.py
```diff
@@ -0,0 +1,219 @@
...
+        # TODO: We will need to merge the changes in AiiDA-QE PR#969 in order
+        # to better handle magnetic and Hubbard data. For now, we can probably
+        # leave it as it is.
+        initial_magnetic_moments=parameters["advanced"]["initial_magnetic_moments"],
+        overrides=overrides,
+        **kwargs,
+    )
+    builder.pop("relax")
+    builder.pop("clean_workdir", None)
+    builder.spglib_settings = spglib_settings
```
superstar54, line 202: It seems the get_builder_from_protocol does not support `structure_preparation_settings` as input. so we need add it here:
```suggestion
    builder.spglib_settings = spglib_settings
    builder.structure_preparation_settings = structure_preparation_settings
```

src/aiidalab_qe/plugins/xas/workchain.py
```diff
@@ -0,0 +1,219 @@
...
+    # there is a bug in aiida-quantumespresso xps, that one can not set the kpoints
+    # this is fxied in a PR, but we need to wait for the next release.
+    # we set a large kpoints_distance value to set the kpoints to 1x1x1
+    if is_molecule_input:
+        # kpoints = KpointsData()
+        # kpoints.set_kpoints_mesh([1, 1, 1])
+        # parameters["advanced"]["kpoints"] = kpoints
+        # builder.ch_scf.kpoints_distance = Float(5)
+        pass
```
superstar54, line 211: Please remove these lines.
Does this kpoint problem also exist in XAS workflow?

superstar54 (changes requested):
@PNOGillespie thanks for the work!

I did some test calculations. Here are some comments on the workchain part.

src/aiidalab_qe/plugins/xas/workchain.py
```diff
@@ -0,0 +1,219 @@
...
+    # before thinking about a workflow.
+    is_molecule_input = (
+        True if xas_parameters.get("structure_type") == "molecule" else False
+    )
+
+    # core_hole_treatment = xas_parameters["core_hole_treatment"]
+    # core_hole_treatments = {element: core_hole_treatment for element in elements_list}
+
+    structure_preparation_settings = {
+        # "supercell_min_parameter": Float(supercell_min_parameter_map[protocol]),
```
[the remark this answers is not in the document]
PNOGillespie: Essentially, the setting is not used because it doesn't seem to be particularly useful in the vast majority of cases. The reasons for this are:

1. The default value of 8 angstrom is sufficiently converged for every system we've ever worked on, regardless of type.
2. Increasing the value for XAS has very little effect on the quality of the final spectrum compared to the cost increase of the `pw.x` calculation.
3. Reducing the value in order to save time doesn't work as one may think:
      - The cost of the `xspectra.x` calculation doesn't depend much on the system size (as the Lanczos procedure specifically avoids a costly calculation of the full manifold of unoccupied states).
      - `xspectra.x` performs the Lanczos calculation for one K-point at a time and can't reduce K-points by symmetry, so systems with more K-points always take longer than those with fewer. Thus reducing the cell size while maintaining the same K-point density actually *increases* the calculation time for the same system.

Ultimately, while one saves time on the `pw.x` calculation by reducing the system size, this saving is lost because the `xspectra.x` calculation takes so much longer.

I am in favour of removing this and the other `structure_preparation_settings` parts. I think these were commented out originally just in case they became useful later.

src/aiidalab_qe/plugins/xas/workchain.py
PNOGillespie: At the moment, these can be temporarily removed or commented out since most are simply not useful at the moment (see my other comment here about `supercell_min_parameter` in particular).

The `standardize_structure` option will eventually be needed for magnetic systems, since this is to ensure that the magnetic ordering isn't changed by standardizing the structure, as well as `use_element_types` from the [PR](https://github.com/aiidateam/aiida-quantumespresso/pull/969) that I opened for AiiDA-QE), but they will only work with the changes in that other PR.

src/aiidalab_qe/plugins/xas/workchain.py
PNOGillespie: > Please remove these lines.

Will do.

> Does this kpoint problem also exist in XAS workflow?

I've not tested the behaviour, but this is because we're only considering periodic systems for XAS at this point and thus using a K-points distance value obviously makes more sense than an explicit mesh.

I can try testing this in my own time, which may be useful for an XAS workflow for molecules, but it's not worth pursuing at the moment.

src/aiidalab_qe/plugins/xas/result.py
PNOGillespie: Hi @unkcpz & @AndresOrtegaGuerrero, I've run some tests with the `DownloadButton` widget and compared it to the `SpectrumDownloadButton` widget I made for this plugin. I can get all the same functionality out of `DownloadButton` as I did for my own one, though the issue I notice is that the Plotly viewer runs noticeably slower when working with the `DownloadButton`.

The reason, I'm guessing, is that the `DownloadButton` needs the file's contents (`payload`) to be a `bytes` object, so each time the spectrum is re-drawn (e.g. if the spectrum broadening or element selection is changed) it needs to encode a new `bytes` object, which takes extra time (probably even more for spectra composed of multiple different absorbing atoms). The `SpectrumDownloadButton` doesn't have this issue because encoding is done during `__on_click` and so the input can just be a `str` object.

I do like the idea of a generalised `DownloadButton` widget (particularly as there's only really one way to do make this work anyway) but I think it needs some encoding methods implemented to enhance the performance.

src/aiidalab_qe/plugins/xas/workchain.py
PNOGillespie: Lines have been removed in the latest commit (https://github.com/aiidalab/aiidalab-qe/pull/580/commits/9a9f16269185b2a136b7914d095c6d41959dc9f1)

src/aiidalab_qe/plugins/xas/workchain.py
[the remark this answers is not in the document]
PNOGillespie: The parameter has now been enabled in the latest commit (https://github.com/aiidalab/aiidalab-qe/pull/580/commits/9a9f16269185b2a136b7914d095c6d41959dc9f1).

superstar54 (changes requested):
Hi @PNOGillespie, I see that you checked and downloaded the pseudo in the `workchain.py` part. I suggest moving them to the `setting.py.` Because the pseudo information is used in the setting panel. Of course, you can do this in another PR.

The PR will be good to me when you make the changes.

For the download button, let Andres finish his, and we can discuss which one to go with. You can update it in the future PR.

src/aiidalab_qe/plugins/xas/setting.py
```diff
@@ -0,0 +1,244 @@
...
+            "core_hole_treatments": core_hole_treatments,
+            "elements_list": elements_list,
+            "structure_type": self.structure_type.value,
+            "gipaw_pseudo": self.gipaw_pseudos,
+            "ch_pseudo": self.core_hole_pseudos,
+            "core_wfc_data": self.core_wfc_data,
+        }
+        return parameters
+
+    def set_panel_value(self, input_dict):
```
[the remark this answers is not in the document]
PNOGillespie: Ok, this did confuse me a little on review, but `element_and_ch_treatment_options` is indeed not set, however it also doesn't do anything. The `core_hole_treatments` and `elements_list` however *are* set correctly - I tested it just now - but that's defined by the code in `_update_element_selection_panel()` (lines 166+).

The short answer to this is that we can just delete all of the code in lines 141-159. I can discuss this in more detail separately if you would like.

src/aiidalab_qe/plugins/xas/setting.py
```diff
@@ -0,0 +1,244 @@
...
+    identifier = "xas"
+    input_structure = tl.Instance(StructureData, allow_none=True)
+    protocol = tl.Unicode(allow_none=True)
+
+    element_selection_title = ipw.HTML(
+        """<div style="padding-top: 0px; padding-bottom: 0px">
+        <h4>Element and Core-Hole Treatment Setting.</h4></div>"""
+    )
+
+    # TODO: The element selection should lock the "Confirm" button if no elements have been
```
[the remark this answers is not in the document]
PNOGillespie: I'll open one once this PR is concluded, no problem.

src/aiidalab_qe/plugins/xas/setting.py
```diff
@@ -0,0 +1,244 @@
...
+        )
+
+        element_and_ch_treatment_options = {}
+        for element in elements_to_select:
+            if element in xch_elements:
+                element_and_ch_treatment_options[element] = "xch_smear"
+            else:
+                element_and_ch_treatment_options[element] = "full"
+
+        self.element_and_ch_treatment_options = element_and_ch_treatment_options
```
PNOGillespie, line 160: Hi @superstar54. Does this seem like a reasonable way to address your [previous comment](https://github.com/aiidalab/aiidalab-qe/pull/580/files/3f6acbceb603cc282e1333674e3b4b7ddb0f84a2#r1465091243)?

I suggest this because the logic to define the initial values for the `element_and_ch_treatment` panel has already been defined in `_update_element_select_panel()`. This works fine with my own testing and sets the right behaviour for `set_panel_value()`.

```suggestion
        self._update_element_select_panel()
```

superstar54:
The codecov decreased because there is no test on the workchain and its result. We will come to this when we resolve #533.

superstar54 (approved):

superstar54 (approved):

unkcpz:
> The codecov decreased because there is no test on the workchain and its result. We will come to this when we resolve #533. 

 Thanks for mentioning this. I'd suggest we iron the test first.

unkcpz:
> We leave the pseudopotential setup and the download button to the feature PR.

As we talked over the phone, please open issues for these isuses that not addressed. And we should try to do it as soon as possible, otherwise they will never happened.

unkcpz:
Also a mild reminder that try not do the hush merge next time. If you have DDL then we need to plan things ahead and make things happened under the plan. The integration test is failed so please try to fix it by this week @superstar54 @PNOGillespie

unkcpz:
In fact there is a crucial break in the PR, the integration test shows:

```
FileNotFoundError                         Traceback (most recent call last)
/tmp/ipykernel_1713/2189467020.py in <cell line: 17>()
     15 from jinja2 import Environment
     16 
---> 17 from aiidalab_qe.app import App, static
     18 from aiidalab_qe.version import __version__

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/app/__init__.py in <module>
      1 """Package for the AiiDAlab QE app."""
      2 
----> 3 from .main import App
      4 
      5 __all__ = [

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/app/main.py in <module>
     12 from aiidalab_qe.app.result import ViewQeAppWorkChainStatusAndResultsStep
     13 from aiidalab_qe.app.structure import StructureSelectionStep
---> 14 from aiidalab_qe.app.submission import SubmitQeAppWorkChainStep
     15 from aiidalab_qe.common import QeAppWorkChainSelector
     16 

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/app/submission/__init__.py in <module>
     20 from aiidalab_qe.common.setup_codes import QESetupWidget
     21 from aiidalab_qe.common.setup_pseudos import PseudosInstallWidget
---> 22 from aiidalab_qe.workflows import QeAppWorkChain
     23 
     24 from .resource import ParallelizationSettings, ResourceSelectionWidget

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/workflows/__init__.py in <module>
     40 
     41 
---> 42 plugin_entries = get_entry_items("aiidalab_qe.properties", "workchain")
     43 
     44 

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/workflows/__init__.py in get_entry_items(entry_point_name, item_name)
     32 # load entry point items
     33 def get_entry_items(entry_point_name, item_name="workchain"):
---> 34     entries = get_entries(entry_point_name)
     35     return {
     36         name: entry_point.get(item_name)

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/workflows/__init__.py in get_entries(entry_point_name)
     25     entries = {}
     26     for entry_point in entry_points().get(entry_point_name, []):
---> 27         entries[entry_point.name] = entry_point.load()
     28 
     29     return entries

/opt/conda/lib/python3.9/site-packages/importlib_metadata/__init__.py in load(self)
    205         """
    206         match = self.pattern.match(self.value)
--> 207         module = import_module(match.group('module'))
    208         attrs = filter(None, (match.group('attr') or '').split('.'))
    209         return functools.reduce(getattr, attrs, module)

/opt/conda/lib/python3.9/importlib/__init__.py in import_module(name, package)
    125                 break
    126             level += 1
--> 127     return _bootstrap._gcd_import(name[level:], package, level)
    128 
    129 

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/plugins/xas/__init__.py in <module>
      8 
      9 from .result import Result
---> 10 from .setting import Setting
     11 from .workchain import workchain_and_builder
     12 

/opt/conda/lib/python3.9/site-packages/aiidalab_qe/plugins/xas/setting.py in <module>
     17 from aiidalab_qe.plugins import xas as xas_folder
     18 
---> 19 PSEUDO_TOC = yaml.safe_load(resources.read_text(xas_folder, "pseudo_toc.yaml"))
     20 pseudo_data_dict = PSEUDO_TOC["pseudos"]
     21 xch_elements = PSEUDO_TOC["xas_xch_elements"]

/opt/conda/lib/python3.9/importlib/resources.py in read_text(package, resource, encoding, errors)
    137     bytes.decode().
    138     """
--> 139     with open_text(package, resource, encoding, errors) as fp:
    140         return fp.read()
    141 

/opt/conda/lib/python3.9/importlib/resources.py in open_text(package, resource, encoding, errors)
    119     """Return a file-like object opened for text reading of the resource."""
    120     return TextIOWrapper(
--> 121         open_binary(package, resource), encoding=encoding, errors=errors)
    122 
    123 

/opt/conda/lib/python3.9/importlib/resources.py in open_binary(package, resource)
     89     reader = _get_resource_reader(package)
     90     if reader is not None:
---> 91         return reader.open_resource(resource)
     92     absolute_package_path = os.path.abspath(
     93         package.__spec__.origin or 'non-existent file')

/opt/conda/lib/python3.9/importlib/_bootstrap_external.py in open_resource(self, resource)

FileNotFoundError: [Errno 2] No such file or directory: '/opt/conda/lib/python3.9/site-packages/aiidalab_qe/plugins/xas/pseudo_toc.yaml'
```

unkcpz:
You can reproduce it by start the container `docker run --rm -it -p 8888:8888 aiidalab/qe:edge` and go to the app to see the error.

superstar54:
@unkcpz Sorry for the bug. It should be fixed in #609 .

unkcpz:
No worries, I add the integration test to the required test. It is quite robust and can give useful information, so we need to put an eye on it next time.

Issue #552, opened 2023-11-09
Pdos and Bands widget shrinks the plot when having unit cells with more than 2 atoms


AndresOrtegaGuerrero:
![Screenshot 2023-11-09 at 10 51 28](https://github.com/aiidalab/aiidalab-qe/assets/34098967/595b6de2-5a86-4401-8df2-b18b710b38d3)

AndresOrtegaGuerrero:
This is an example of a unitcell with 18 atoms ,

PR #581, opened 2023-12-13
New BandsPdosWidget

AndresOrtegaGuerrero:
This PR introduces a new bands_pdos_widget using plotly, the widget include options of drag, pan and toogle , save a snapshot , and download the data.

The formating of the bands/pdos has been moved from the result plugin to the common/bandspdoswidget.py

It allows different selection of atoms to consider, or different representations ( We can modified or create new according to your suggestions)

For spin polarized calculations it represents bands_up in red, and bands_down in blue. For non polarized the line is in black color.

The new `_projections_curated_options` allows to curated SOC pdos. 

The legend has a wheel fixing #552

### DIFF
--- a/src/aiidalab_qe/common/bandpdoswidget.py
+++ b/src/aiidalab_qe/common/bandpdoswidget.py
@@ -1,0 +1,862 @@
+import base64
+import json
+
+import ipywidgets as ipw
+import numpy as np
+import plotly.graph_objects as go
+from aiida.orm import ProjectionData
+from aiidalab_widgets_base.utils import string_range_to_list, StatusHTML
+from IPython.display import clear_output, display
+from plotly.subplots import make_subplots
+import re
+
+
+class BandPdosPlotly:
+    SETTINGS = {
+        "axis_linecolor": "#111111",
+        "bands_linecolor": "#111111",
+        "bands_up_linecolor": "rgba(205, 0, 0, 0.4)",  # Red Opacitiy 40%
+        "bands_down_linecolor": "rgba(72,118,255, 0.4)",  # Blue Opacitiy 40%
+        "combined_plot_height": 600,
+        "combined_plot_width": 900,
+        "combined_column_widths": [0.7, 0.3],
+        "bands_plot_height": 600,
+        "bands_plot_width": 850,
+        "pdos_plot_height": 600,
+        "pdos_plot_width": 850,
+        "vertical_linecolor": "#111111",
+        "horizontal_linecolor": "#111111",
+        "vertical_range_bands": [-10, 10],
+        "horizontal_range_pdos": [-10, 10],
+    }
+
+    def __init__(self, bands_data=None, pdos_data=None):
+        self.bands_data = bands_data
+        self.pdos_data = pdos_data
+        self.fermi_energy = self._get_fermi_energy()
+
+        # Plotly Axis
+        # Plotly settings
+        self._bands_xaxis = self._band_xaxis()
+        self._bands_yaxis = self._band_yaxis()
+        self._dos_xaxis = self._dos_xaxis()
+        self._dos_yaxis = self._dos_yaxis()
+
+    def _get_fermi_energy(self):
+        fermi_energy = (
+            self.pdos_data["fermi_energy"]
+            if self.pdos_data
+            else self.bands_data["fermi_energy"]
+        )
+        return fermi_energy
+
+    def _band_xaxis(self):
+        """Function to return the xaxis for the bands plot."""
+
+        if not self.bands_data:
+            return None
+        paths = self.bands_data.get("paths")
+        slider_bands = go.layout.xaxis.Rangeslider(
+            thickness=0.08,
+            range=[0, paths[-1]["x"][-1]],
+        )
+        bandxaxis = go.layout.XAxis(
+            title="k-points",
+            range=[0, paths[-1]["x"][-1]],
+            showgrid=True,
+            showline=True,
+            tickmode="array",
+            rangeslider=slider_bands,
+            fixedrange=False,
+            tickvals=self.bands_data["pathlabels"][1],  # ,self.band_labels[1],
+            ticktext=self.bands_data["pathlabels"][0],  # self.band_labels[0],
+            showticklabels=True,
+            linecolor=self.SETTINGS["axis_linecolor"],
+            mirror=True,
+            linewidth=2,
+            type="linear",
+        )
+
+        return bandxaxis
+
+    def _band_yaxis(self):
+        """Function to return the yaxis for the bands plot."""
+
+        if not self.bands_data:
+            return None
+
+        bandyaxis = go.layout.YAxis(
+            title=dict(text="Electronic Bands (eV)", standoff=1),
+            side="left",
+            showgrid=True,
+            showline=True,
+            zeroline=True,
+            range=self.SETTINGS["vertical_range_bands"],
+            fixedrange=False,
+            automargin=True,
+            ticks="inside",
+            linewidth=2,
+            linecolor=self.SETTINGS["axis_linecolor"],
+            tickwidth=2,
+            zerolinewidth=2,
+        )
+
+        return bandyaxis
+
+    def _dos_xaxis(self):
+        """Function to return the xaxis for the dos plot."""
+
+        if not self.pdos_data:
+            return None
+
+        if self.bands_data:
+            dosxaxis = go.layout.XAxis(
+                title="Density of states",
+                side="bottom",
+                showgrid=True,
+                showline=True,
+                linecolor=self.SETTINGS["axis_linecolor"],
+                mirror="ticks",
+                ticks="inside",
+                linewidth=2,
+                tickwidth=2,
+                automargin=True,
+            )
+
+        else:
+            dosxaxis = go.layout.XAxis(
+                title="Density of states (eV)",
+                showgrid=True,
+                showline=True,
+                linecolor=self.SETTINGS["axis_linecolor"],
+                mirror="ticks",
+                ticks="inside",
+                linewidth=2,
+                tickwidth=2,
+                range=self.SETTINGS["horizontal_range_pdos"],
+            )
+
+        return dosxaxis
+
+    def _dos_yaxis(self):
+        """Function to return the yaxis for the dos plot."""
+
+        if not self.pdos_data:
+            return None
+
+        if self.bands_data:
+            dosyaxis = go.layout.YAxis(
+                # title= {"text":"Density of states (eV)", "standoff": 1},
+                showgrid=True,
+                showline=True,
+                side="right",
+                mirror="ticks",
+                ticks="inside",
+                linewidth=2,
+                tickwidth=2,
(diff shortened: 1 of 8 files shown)


AndresOrtegaGuerrero:
Please have a look , test it let me know any further changes and/or developments.

AndresOrtegaGuerrero:
I think in a second PR, i can add different test for the new widget

unkcpz:
Pinning @mikibonacci, he mentioned he has a use case that needs some adaption on the bandsplotwidget which will be replaced by this PR, so he can give a quick review and a test.

I didn't start looking at the code, but give it a try on the qeapp, from the outcome it is exactly what we needed, thanks for the great work @AndresOrtegaGuerrero.

unkcpz (changes requested):
@AndresOrtegaGuerrero I give it a first go. 

I notice that the data validation is missing. The `BandPdosWidget` accept AiiDA node as input which make the widget not generic and vulnerable to the future change of `BandData`. 
In the original `bandsplotwidget`, we add a middle layer of `dict` type between  AiiDA node and input for the widget. The object pass to the widget then can be validated by the schema. 
I'd suggest adding this design back to the new widget. For the data validation, you can try to use pydantic.

Besides, the unit tests for the newly added widget would be very useful.

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero I give it a first go.
> 
> I notice that the data validation is missing. The `BandPdosWidget` accept AiiDA node as input which make the widget not generic and vulnerable to the future change of `BandData`. In the original `bandsplotwidget`, we add a middle layer of `dict` type between AiiDA node and input for the widget. The object pass to the widget then can be validated by the schema. I'd suggest adding this design back to the new widget. For the data validation, you can try to use pydantic.
> 
> Besides, the unit tests for the newly added widget would be very useful.

@unkcpz  Thank you , you mean that the widget doesnt call `get_pdos_data` , `exports_bands_data` from the widget using the node, but instead that the output from these functions is the input of the widget ?

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+        )
+        self.selected_atoms = ipw.Text(
+            description="Select atoms:",
+            value="",
+            style={"description_width": "initial"},
+        )
+        self.wrong_syntax = ipw.HTML(
+            value="""<i class="fa fa-times" style="color:red;font-size:2em;" ></i> wrong syntax""",
+            layout={"visibility": "hidden"},
+        )
```
unkcpz, line 90: Consider to using `StatusHTML`? Which is designed for pop the warning/error message.
The message display widget like this usually used inside the widget only, so you can make it `self._wrong_syntax` or `self._message`

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+    - fermi_energy: Fermi energy value.
+    - bands_data: Band structure data.
+    - band_labels: Labels and values for band structure plotting.
+    - bands_xaxis: X-axis settings for band structure plot.
+    - bands_yaxis: Y-axis settings for band structure plot.
+    - dos_xaxis: X-axis settings for PDOS plot.
+    - dos_yaxis: Y-axis settings for PDOS plot.
+    - bandsplot_widget: Plotly widget for band structure and PDOS plot.
+    - bands_widget: Output widget to display the bandsplot widget.
+    - pdos_options_out: Output widget to clear specific widgets.
```
unkcpz, line 40: I think most of these attributes don't need to be exposed.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+            description="Select atoms:",
+            value="",
+            style={"description_width": "initial"},
+        )
+        self.wrong_syntax = ipw.HTML(
+            value="""<i class="fa fa-times" style="color:red;font-size:2em;" ></i> wrong syntax""",
+            layout={"visibility": "hidden"},
+        )
+        self.update_plot_button = ipw.Button(
+            description="Update Plot",
```
unkcpz, line 92: The `update_plot_button` seems redundant to me, when users make changes to the widget, after the change, the plot is supposed to be updated automatically.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+        )
+        self.download_button = ipw.Button(
+            description="Download Data",
+            icon="download",
+            button_style="primary",
+            disabled=False,
+            layout=ipw.Layout(visibility="hidden"),
+        )
+
+        self.export_dir = Path.cwd().joinpath("exports")
```
unkcpz, line 105: this is not used anywhere else. 
```suggestion
```

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+            else:
+                my_clean_labels.append(i)
+
+        labels = [label[0] for label in my_clean_labels]
+        labels_values = [label[1] for label in my_clean_labels]
+        return [labels, labels_values]
+
+    def _band_xaxis(self):
+        """Function to return the xaxis for the bands plot."""
+        import plotly.graph_objects as go
```
unkcpz, line 256: Any reason to put the import for each function? If not, please move it to the top.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+        return fermi_energy
+
+    def _get_bands_data(self):
+        if not self.bands:
+            return None
+
+        bands = export_bands_data(self.bands, self.fermi_energy)
+        return bands
+
+    def _bands_labeling(self, bands):
```
unkcpz, line 220:
```suggestion
    def _get_bands_labeling(self, bands):
```
this is a method that return the label from the input data, better to call with `_get` as you did for other methods.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+        """Function to return two lists containing the labels and values (kpoint) for plotting.
+        params:
+        - bands: dictionary from export_bands_data function
+        output: label (list of str), label_values (list of float)
+        """
+        if not bands:
+            return None
+
+        my_paths = bands[0].get("paths")
+        my_labels = []
```
unkcpz, line 230:
```suggestion
        paths = bands[0].get("paths")
        labels = []
```
"my"? The paths and labels have their own free will, they belong to nobody ;P

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+            if label_a not in my_labels:
+                my_labels.append(label_a)
+            if label_b not in my_labels:
+                my_labels.append(label_b)
+
+        my_clean_labels = []  # Format
+        for i in my_labels:
+            if my_clean_labels:
+                if i not in my_clean_labels:
+                    if my_clean_labels[-1][-1] == i[1]:
```
unkcpz, line 243:
```suggestion
            if my_clean_labels and (i not in my_clean_labels) and (my_clean_labels[-1][-1] == i[1]):
```
I guess this is the same? If not can you try to simplify it, too many layers of nest indicate things that can be polished.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+        # Plotly widget
+        self.bandsplot_widget = self._bandsplot_widget()
+
+        # Output widget to display the bandsplot widget
+        self.bands_widget = ipw.Output()
+
+        # Output widget to clear the specific widgets
+        self.pdos_options_out = ipw.Output()
+
+        def download_data(_=None):
```
unkcpz, line 128: Can you move this function out?

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+            layout=ipw.Layout(visibility="hidden"),
+        )
+
+        self.export_dir = Path.cwd().joinpath("exports")
+
+        # Information for the plot
+        self.dos_data = self._get_dos_data()
+        self.fermi_energy = self._get_fermi_energy()
+        self.bands_data = self._get_bands_data()
+        self.band_labels = self._bands_labeling(self.bands_data)
```
unkcpz, line 111: I didn't see the reason why these attributes need to initialize here. When the widget initialized, there is no plot, so these attributes are not needed.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -0,0 +1,833 @@
...
+                ticks="inside",
+                linewidth=2,
+                tickwidth=2,
+                linecolor=self.SETTINGS["axis_linecolor"],
+                zerolinewidth=2,
+            )
+
+        return dosyaxis
+
+    def _bandsplot_widget(self):
```
unkcpz, line 381: The method name is `bandsplot` but it also deal with pdos, better to be called as `bandpdosplot`.
Moreover, this function return a plotly plot objcet, so I'd call it `get_bandpdos_plot`. 

The data flow is more clean to be like: data -> plotly plot object -> ipyoutput widget, can you check the PR and try to apply this flow?

unkcpz (changes requested):
Thanks for the changes @AndresOrtegaGuerrero. I have some request upon method names and especially the flow of data/objects.

This is a widget for dealing with the bands/pdos data, convert the data to the format that plotly can plot a plot object. Then generate the plotly plot object and render it to the ipyoutput. 

I'd suggest at least separate methods for plotly plot object creating to a class `BandPdosPlot`. The class getting a formatted bands/pdos data and knowing how to create a plotly plot object. Then for the `BandPdosWidget`, it dealing with 1) parsing the data to the formatted data and pass to `BandPdosPlot` 2) dealing the logic with the QeApp, such as initialize the widget and hide the widget.

src/aiidalab_qe/common/bandpdoswidget.py
AndresOrtegaGuerrero: @unkcpz  This could be problematic, considering that you also have the choice to select atoms , is better to control when to update the plot , once you are done defining the atoms you are selecting.

src/aiidalab_qe/common/bandpdoswidget.py
AndresOrtegaGuerrero: When i do it , it doesnt work i will check if i can move it out

unkcpz:
Hi @AndresOrtegaGuerrero, it is all in good shape. Before merge, I want to ask for a final test on using the new widget for the phonon bands plot. Can you or @mikibonacci give a test on using the new widget for it? I think it is a good proof that the API exposed by the new widget is generic enough for most of use cases.

unkcpz:
Since this replace the `widget_bandsplot`, it should be able to remove the dependency from `setup.cfg`.

PR #598, opened 2024-01-23
removing force_parity from create_kpoints_from_distance

AndresOrtegaGuerrero:
This allows the app to follow the same logic of aiida-quantumespresso plugin, therefore displaying the final kpoint mesh as in the calculation

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced.py
@@ -328,7 +328,7 @@ class AdvancedSettings(Panel):
             mesh = create_kpoints_from_distance.process_class._func(
                 self.input_structure,
                 orm.Float(self.kpoints_distance.value),
-                orm.Bool(True),
+                orm.Bool(False),
             )
             self.mesh_grid.value = "Mesh " + str(mesh.get_kpoints_mesh()[0])
         else:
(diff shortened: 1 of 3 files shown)


unkcpz:
> This allows the app to follow the same logic of aiida-quantumespresso plugin

Hi @AndresOrtegaGuerrero, what you mean of "same logic"? I look at the function in `aiida-quantumespresso`, but it has no default value defined for `force_parity`. https://github.com/aiidateam/aiida-quantumespresso/blob/c44faadc7c11f7118290cf1b7affe49313bd4467/src/aiida_quantumespresso/calculations/functions/create_kpoints_from_distance.py#L7

AndresOrtegaGuerrero:
@unkcpz in the base workchain, https://github.com/aiidateam/aiida-quantumespresso/blob/main/src/aiida_quantumespresso/workflows/pw/base.py#L7 

thats why the value we display is not necessarily the one is being used in the calculations since it should be force_parity false

unkcpz:
Ah, I see, it is https://github.com/aiidateam/aiida-quantumespresso/blob/b517686df8cd33765ef12cbf430620ea8e94f137/src/aiida_quantumespresso/workflows/pw/base.py#L278 😉

unkcpz (approved):
Thanks!

PR #593, opened 2024-01-17
Use ruff as code formatter

unkcpz:
`aiida-core` from `2.5.0` start to move to using ruff as code formatter, it is much faster.

PR #599, opened 2024-01-24
Remove the load_profile warning by move it to front

unkcpz:
The problem is introduced by https://github.com/aiidalab/aiidalab-qe/pull/593, where I move the load_profile to after the import of AWB.


AndresOrtegaGuerrero (approved):
i tested and it removed the warning message, LGTM ! , thank you !

PR #608, opened 2024-01-26
Update acknowledgement

superstar54:
- Add Acknowledgement to BIG-MAP and EU in both README and docs
- move logos to the `docs/_static/logos`, so that both README and docs can access


unkcpz (approved):
LGTM, thanks!

PR #614, opened 2024-02-01
Add Feature: Documentation Page for XAS Plugin

PNOGillespie:
# Overview

This PR adds a complete documentation set for using the XAS plugin of AiiDALab-QE, covering both calculation setup and post-processing of the final spectra using the Plotly-derived widget provided with the plugin. The instructions should adequately cover all features of the XAS App using a self-contained, robust, and well-tested example.

This will be followed up by a video demonstration of this example, as required for BIG-MAP project key deliverables, to accompany the documentation.

Note that the documentation example uses Li<sub>2</sub>CO<sub>3</sub> as the structure, which has been added to the "From Examples" list in `src/aiidalab_qe/app/static/`, so there shouldn't be any issues in the future with missing links to important files.


PNOGillespie:
Hi @superstar54, I reformatted the source code as you asked so that each sentence is on a separate line. Obviously this wasn't possible for lists, because the syntax doesn't let you do that, but hopefully that makes review easier.

superstar54 (changes requested):
Nice docs! Very clear and rigorous. I only has one small request, please see the comment.

docs/source/howto/xas.rst
```diff
@@ -0,0 +1,190 @@
...
+
+.. note::
+    The XAS plugin feature is available in the official AiiDALab-QE App as of pre-release `v24.04.0a1` and upwards.
+    See the relevant :doc:`How-To </howto/upgrade_uninstall>` guide on upgrading and remember to tick `include prereleases` when searching for the latest version in the App Manager.
+
+    If installing `via` the terminal, simply give the version number using the correct `PEP 440 version specifier`_\ :
+
+    .. code-block:: console
+
+       $ aiidalab install quantum-espresso>=v24.04.0a1
```
superstar54, line 26: There seems to be a bug in the `aiidalab install`.  This command does not work as expected. I would suggest removing this part. We try to avoid the user using the command line if they can do it in the GUI.

superstar54:
> Just out of curiosity (and also in case this needs to be documented properly), is there already an issue page for the problem with `aiidalab install`?

One needs to include `"` or `'`
```diff
- aiidalab install quantum-espresso>=v24.04.0a1
+ aiidalab install "quantum-espresso>=v24.04.0a1"
```

superstar54 (approved):

PR #619, opened 2024-02-09
Fix Hexagonal 2D path order

AndresOrtegaGuerrero:
The Hexagonal path should be 
Gamma-M-K-Gamma, 

This fixes the issue and also 
M = 0.5 0.0 0.0


superstar54 (changes requested):
@AndresOrtegaGuerrero, Can you add a reference (link or paper) to show where all these paths come from?
Can you also double-check the `K`?

AndresOrtegaGuerrero:
@superstar54  the citation is already in the code , in line 107 of the same file , `https://pubs.acs.org/doi/10.1021/acs.jpclett.2c02972`

superstar54 (approved):
LGTM. Thanks!

PR #620, opened 2024-02-09
Adding Van del Waals Correction widget (Adv Setting)

AndresOrtegaGuerrero:
This PR adds the feature to include Van der Waals Correction to the QE App.


superstar54 (approved):
Great! Thanks for the work.

One tiny question: I am not familiar with the `vdw_corr`. I see in the pw doc, there are other options: 'mdb_vdw' and 'xdm'. Any reason why not include them?

PR #622, opened 2024-02-14
Examples: Add Lithium Carbonate to List of Examples

PNOGillespie:
Fixes an oversight from https://github.com/aiidalab/aiidalab-qe/pull/614 where the lithium carbonate example was not added to `Examples` in `__init__.py` and thus couldn't be selected as a structure.


PNOGillespie:
Hi @superstar54, many apologies for missing this in the PR. I think I had it set correctly when I was building the docs in the docker container, then switched development to the local machine and forgot to include this change.

I tested it just now, also to confirm that the structure loads correctly, and everything seems fine.

superstar54 (approved):

superstar54:
No worry!

PR #624, opened 2024-02-21
Add `fat bands` to the `BandPdosWidget`.

t-reents:
Hi @unkcpz @AndresOrtegaGuerrero 

As discussed, I worked on the support for the "fat bands" plots and I managed to merge this into the existing `BandPdosWidget`. While working on this, I also started to refactor some of the existing methods, see the commit message for more details, mainly to split things up in a more compact way.

I have some further ideas, but I already wanted to share this version, to get further input. One thing that would be useful to address in my opinion, are different linestyles for the spin-up and spin-down projected bands, as the underlying red and blue linecolour is not  always easily visible.

Important to note, the `Project bands` checkbox is only used for this first draft. This would be changed to something more properly formatted in the final version. 

One additional question @AndresOrtegaGuerrero: is it really necessary to have these nested loops to iterate each segment separately? At the moment, I don't really see the reason, please correct me if I'm wrong. As mentioned in the commit message, since this version relies on `BandsData._get_bandplot_data`, one could also plot the bands in a single trace (need to check how much this would improve performance).
https://github.com/aiidalab/aiidalab-qe/blob/f64607c601f0f468902050e167b5617d415a37db/src/aiidalab_qe/common/bandpdoswidget.py#L262-L266

The current implementation works for non-spin-polarized and spin-polarized calculations. Moreover, the projected bands can be used with and without the PDOS (and all kind of groupings are supported as well).
image

The orbital/atom contributions in the legend are grouped if the PDOS is plotted as well, so that the chosen contributions can be hidden in both subplots at the same time. 
image


src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -463,6 +502,7 @@ def __init__(self, bands=None, pdos=None, **kwargs):
                 self.dos_plot_group,
                 ipw.HBox([self.selected_atoms, self._wrong_syntax]),
                 self.update_plot_button,
+                self.project_bands_box,
```
AndresOrtegaGuerrero, line 505: I would place this check box above the update_button

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -564,6 +619,109 @@ def _clear_output_and_display(self, widget=None):
...
 
+def _prepare_projections_to_plot(bands_data, projections, bands_width):
+    """Prepare the projected bands to be plotted.
+
+    This function concatenates the different bands in a single list and adds a np.nan as a separator.
+    This way, all the bands are plotted in a single trace. To use the fill option `toself`,
+    a band needs to be concatenated with its mirror image, first.
+    """
+    projected_bands = []
+    for spin in [0, 1]:
```
AndresOrtegaGuerrero, line 630: maybe try to use the naming from aiidalab_qe , as collinear

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -639,17 +797,39 @@ def get_pdos_data(pdos, group_tag, plot_tag, selected_atoms):
...
 
-def _projections_curated_options(
-    projections: ProjectionData,
+def _get_grouping__key(
     group_tag,
     plot_tag,
-    selected_atoms,
-    spin_type="none",
-    line_style="solid",
+    atom_position,
```
AndresOrtegaGuerrero, line 803: what is this atom_position ? is this meant to be the selected_atoms ? if so keep the original naming so it is easy to understand that these are the list of the selected atoms to project

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -676,141 +856,123 @@ def _projections_curated_options(
...
+    """Extract and curate the projections.
+
+    This function can be used to extract the PDOS or the projections data.
+    """
+    _proj_pdos = {}
+    list_positions = []
+
+    # Constants for spin types
+    SPIN_LABELS = {"up": "(↑)", "down": "(↓)", "none": ""}
+    SPIN_PDOS_FACTORS = {"up": 1, "down": -1, "none": 1}
```
AndresOrtegaGuerrero, line 903: this is not clear to me the naming , in particular , the "none": 1 , instead of using numbers could you try other labels so is easy to read and maintaing ?

t-reents: Those aren't labels, the factors are used to multiply the PDOS with this factor, i.e. leaving spin-up and non-polarized unchanged and converting spin-down to negative values. 
https://github.com/aiidalab/aiidalab-qe/blob/99469d6ec193192d74e1f8b5f9a8748427c3eba7/src/aiidalab_qe/common/bandpdoswidget.py#L957

This way, the if-statements can be avoided and a similar logic as in the case of the labels is applied:
https://github.com/aiidalab/aiidalab-qe/blob/f64607c601f0f468902050e167b5617d415a37db/src/aiidalab_qe/common/bandpdoswidget.py#L780-L785

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -397,6 +431,7 @@ class BandPdosWidget(ipw.VBox):
         Select the style of plotting the projected density of states.
         </div>"""
     )
+    projected_bands_width = 0.5
```
AndresOrtegaGuerrero, line 434: Maybe add this to the SETTINGS dictionary

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -443,6 +478,10 @@ def __init__(self, bands=None, pdos=None, **kwargs):
             disabled=False,
             layout=ipw.Layout(visibility="hidden"),
         )
+        self.project_bands_box = ipw.Checkbox(
+            value=False,
+            description="Project bands",
```
AndresOrtegaGuerrero, line 483: Maybe something more clear fr the user , change the description so the user understand that is going to add like the projection as a Fat band , or a little more specific for the user

src/aiidalab_qe/common/bandpdoswidget.py
AndresOrtegaGuerrero: also try to see if this logic is consistent for SOC scenario ?

t-reents:
Thanks for your comments @AndresOrtegaGuerrero. I tested all kind of selections and they seem to work. I was also thinking about a SOC calculation but don't have experience with those calculations and no data right now. Do you have a WorkChain to share as an archive, so that I can use it for the development?

Yes, I'll provide a "one trace" solution for the bands tomorrow, and I'll keep in mind that each band/k-point can still be selected.

>If the plot is only bands with projection then the widgets for displaying the projection should be displayed if there is a projection and hidden if not

Exactly, I haven't done this yet, but the idea was that the grouping options are alway displayed in case PDOS or band projections are available

src/aiidalab_qe/common/bandpdoswidget.py
t-reents: Makes sense!

src/aiidalab_qe/common/bandpdoswidget.py
AndresOrtegaGuerrero: I see  would it make sence to change SPIN_PDOS_FACTORS , as MULTIPLICATION_FACTORS ? or SIGN_MULT_FACTOR ? or a name that is easy to understand that is a multiplication factor ?

AndresOrtegaGuerrero:
For SOC , you can install this plugin to your aiidalab_qe (https://github.com/AndresOrtegaGuerrero/aiidalab-qe-soc/tree/main) , and run a calculation , however for the projections , we need to do a modification , you can check in (https://github.com/AndresOrtegaGuerrero/aiidalab-qe-soc/blob/main/aiidalab_qe_soc/socworkchain.py) , the special settings are in (https://github.com/AndresOrtegaGuerrero/aiidalab-qe-soc/blob/main/aiidalab_qe_soc/workchain.py) , one single thing is to replace the BandsWorkChain for the one you use for the projection

src/aiidalab_qe/common/bandpdoswidget.py
t-reents: Not sure, if I fully get your point here. The `0` and `1` are the indicators for the different spin-channels that are used in the `BandsData. _get_bandplot_data` method. Therefore, I didn't use something like "up" and "down". 

But good point, I'll check how these list would look like in case of SOC.

src/aiidalab_qe/common/bandpdoswidget.py
t-reents: This was just to have a first starting point, to check that everything works.

Not sure, how you think about it (I'm also not a 100% sure, what I think about it, as the layout might get too complicated with too many options), but one could also add a dropdown/slider to choose the maximum linewidth, since the perfect value might differ for different applications?

t-reents:
> For SOC , you can install this plugin to your aiidalab_qe (https://github.com/AndresOrtegaGuerrero/aiidalab-qe-soc/tree/main) , and run a calculation , however for the projections , we need to do a modification , you can check in (https://github.com/AndresOrtegaGuerrero/aiidalab-qe-soc/blob/main/aiidalab_qe_soc/socworkchain.py) , the special settings are in (https://github.com/AndresOrtegaGuerrero/aiidalab-qe-soc/blob/main/aiidalab_qe_soc/workchain.py) , one single thing is to replace the BandsWorkChain for the one you use for the projection

Thanks, I'll give it a try. Probably I'll just do it via AiiDA-core as I'm used to it

AndresOrtegaGuerrero:
@t-reents Currently i am also working with a modification to the widget here #512 , the new feature is to allow the app to use tot_magnetization. When you use `tot_magnetization` , you now have two fermi energies in the BandsWorkChain and also in PdosWorkChain. I would suggest you run your workchain adding tot_magnetization , and verify your logic when the data doesnt have "fermi_energy".

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -385,6 +418,7 @@ class BandPdosWidget(ipw.VBox):
     - selected_atoms: Text widget to select specific atoms for PDOS plotting.
     - update_plot_button: Button widget to update the plot.
     - download_button: Button widget to download the data.
```
AndresOrtegaGuerrero, line 420: Does the download button method include the information of the project_bands for downloading ?

src/aiidalab_qe/common/bandpdoswidget.py
t-reents: It's actually the original naming. I just moved your inner function to a private function at the module level.

This is the original naming convention:
https://github.com/aiidalab/aiidalab-qe/blob/f64607c601f0f468902050e167b5617d415a37db/src/aiidalab_qe/common/bandpdoswidget.py#L682-L686

src/aiidalab_qe/common/bandpdoswidget.py
t-reents: Would need to cross check if something is broken, but the `projected_bands` are transformed into a list and stored in the `dict`, so they should be downloaded as well.

t-reents: I fixed it in the latest commit, works again.

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -103,217 +118,154 @@ def _band_yaxis(self):
...
-    def _get_bandspdos_plot(self):
-        """Function to return the bands plot widget."""
-        conditions = {
-            (True, False): self._create_bands_only_plot,
-            (False, True): self._create_dos_only_plot,
-            (True, True): self._create_combined_plot,
+        axis_settings = {
+            "showgrid": True,
+            "showline": True,
+            "side": "right" if self.plot_type == "combined" else "left",
```
AndresOrtegaGuerrero, line 157: nice!

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -103,217 +118,154 @@ def _band_yaxis(self):
...
-
-        self._add_band_traces(fig, paths, "bands_only")
+        fig = self._create_fig()
+        if self.bands_data:
+            self._add_band_traces(fig)
+
+            band_labels = self.bands_data.get("pathlabels")
+            for lab in band_labels[1]:
+                fig.add_vline(
+                    x=lab, line=dict(color=self.SETTINGS["vertical_linecolor"], width=1)
```
AndresOrtegaGuerrero, line 178: maybe replace lab, for label , so it is clear and easy to maintain

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -103,217 +118,154 @@ def _band_yaxis(self):
...
 
-    def _create_bands_only_plot(self):
+    def _get_bandspdos_plot(self):
         """Function to return the bands plot widget."""
 
-        fig = go.Figure()
-        paths = self.bands_data.get("paths")
-
-        self._add_band_traces(fig, paths, "bands_only")
+        fig = self._create_fig()
```
AndresOrtegaGuerrero, line 171: I noticed that you always create a subplot ? why ?

src/aiidalab_qe/common/bandpdoswidget.py
AndresOrtegaGuerrero: ah my bad, i though this was `_projections_curated_options` i see your point, one thing `_get_grouping__key` change it to `_get_grouping_key` :)

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -112,303 +127,241 @@ def _band_yaxis(self):
...
+
+    def _add_band_traces(self, fig):
+        """Generate the band traces and add them to the figure."""
+        colors = {
+            (True, 0): self.SETTINGS["bands_up_linecolor"],
+            (True, 1): self.SETTINGS["bands_down_linecolor"],
+            (False, 0): self.SETTINGS["bands_linecolor"],
+        }
+        fermi_energy_mapping = {
+            (False, 0): self.fermi_energy.get("fermi_energy_up", None),
```
AndresOrtegaGuerrero, line 245: what happens when there is only one fermi_energy?

t-reents: The key would be `(True, 0)` or `(True, 1)` in that case but those are not present in the mapping. Therefore, the value for `fermi_energy` would be returned as expected:

https://github.com/aiidalab/aiidalab-qe/blob/25edba65da0e321c93020e4ca315a57565f00b1b/src/aiidalab_qe/common/bandpdoswidget.py#L267-L270

src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -523,6 +490,7 @@ def __init__(self, bands=None, pdos=None, **kwargs):
                 self.dos_atoms_group,
                 self.dos_plot_group,
                 ipw.HBox([self.selected_atoms, self._wrong_syntax]),
+                self.project_bands_box,
```
AndresOrtegaGuerrero, line 493: We should a logic that if the calculations doesnt have project_bands, then the box should vi invisible, maybe add a logic with ipw.Output() , for display or clear.

AndresOrtegaGuerrero: Also if you add the box here , this box will only be visible for the pdos, and this is not correct since it should appear if we have bands_data

AndresOrtegaGuerrero: We need to modify the logic here , since the full pdos box should appear for bands with fat bands projections

t-reents: Done

AndresOrtegaGuerrero (approved):
LGTM!, We did different tests and check the logic and test the qe app! It also improves the performance

AndresOrtegaGuerrero:
This PR solved #626

PR #625, opened 2024-02-27
fix pw setting for molecule in XPS calculation

superstar54:
For molecules, we use `gamma_only` and `assume_isolated`. This should be fixed in the aiida-quantumespresso XpsWorkchain, but we don't want to wait for the new PR and release. Thus, we fix it here directly.

I also adjust the `supercell_min_parameter_map,` so it uses the default 8.0 when the user selects the `moderate` protocol.


PNOGillespie (approved):
Looks fine to me and a quick test on my machine works with no issues. Good work @superstar54.

PR #631, opened 2024-03-05
Run one SCF calculation for all plugins

superstar54:
Currently, plugins are independent, meaning one plugin will not know what other plugins are running. Thus, when selecting both Bands and DOS, the app will run two completely independent SCFs in parallel.

This PR adds one `SCF` step in the main workflow before running any plugins. On the plugin side, I added a `requires_scf` and `from_ctx` keys. If any plugin asks a SCF calculation, the main workchain will do that, and save the scf folder as `ctx.scf_folder`, so that the plugins can use it.

However, this PR is blocked by aiida-quantumespresso, because the `scf` namespace is required in the `PwBandsWorkChain`. In principle, one can remove this requirement so that it can use `scf` calculation from other calculations.
Hi @mbercx , am I correct on this?

### DIFF
--- a/src/aiidalab_qe/plugins/bands/workchain.py
+++ b/src/aiidalab_qe/plugins/bands/workchain.py
@@ -215,6 +215,8 @@ def get_builder(codes, structure, parameters, **kwargs):
 
     # pop the inputs that are excluded from the expose_inputs
     bands.pop("relax")
+    # run the scf workchain from the app's workchain
+    bands.pop("scf")
     bands.pop("structure", None)
     bands.pop("clean_workdir", None)
     return bands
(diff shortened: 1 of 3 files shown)


AndresOrtegaGuerrero:
depending on a plugin the settings might be different, perhaps just try to link bands and pdos , but not all plugins.

mikibonacci:
Hi @superstar54, as we discussed, maybe part of the solution can be to exploit the caching mechanism.
The main advantage is that we can trigger only a previous scf which is really the same. In this way, no errors in connecting things will be done (ex. scf for conventional cell, but then you want the bands for primitive).

However for now all the plugins are submitted at the same time, so no time to create the caching record.

superstar54:
run a scf if no property is selected and no relax

PR #642, opened 2024-03-20
add an adjustable intensity factor for XPS spectra

### LINKED ISSUE #627: Support scaling the intensities of the computed XPS spectra
Add a UI element to scale the intensities of the computed XPS traces so as to make it easier to compare the profile of (in particular) the total XPS to the experimental data.

superstar54:
As suggested by @PNOGillespie  in this [comment](https://github.com/aiidalab/aiidalab-qe/pull/438#pullrequestreview-1909066843). fix #627 

Now, the user can scale the calculated spectra and compare them with experimental data. Here is an example for the PA molecule, green color is the experimental data.

Screenshot from 2024-03-20 13-53-17


PNOGillespie (changes requested):
Hi @superstar54. Though I agree that this is a necessary feature, I think that the code should scale the intensity of the experimental data rather than the simulated data.

This is for the simple reason that, as the intensities of the computed spectra are modified whenever `intensity` is changed, the `Result` panel has to re-draw the entire plot from scratch. Scaling the intensity of the experimental data instead means that only one `trace` in the plot object needs to be updated, which is considerably faster.

I have a potential way to do this, but the changes are too extensive to easily show as suggested changes, so I will open a PR to your feature branch shortly.

superstar54:
As discussed with @PNOGillespie , we think it's better to provide an initial scale value for the user.

superstar54:
@PNOGillespie Now, the app will calculate an initial guess for the intensity factor, by aligning the max value of the total spectra with the max value of the experimental data. Please review.

PNOGillespie (approved):
Looks good to me! I think the last change particularly does a lot of good for usability. Good work @superstar54.

PR #643, opened 2024-03-20
add aiida-core-version in ci and action

### LINKED ISSUE #638: Add test for latest and oldest support aiida-core version
The same idea was implemented to aiidalab-widgets-base in https://github.com/aiidalab/aiidalab-widgets-base/pull/558, it can be adopted here to make the support of migration from pydantic v1 to pydantic v2 smooth and robust.

superstar54:
It seems #638 is an urgent issue that needs to be fixed.

This PR tries to fix #638, and

- add `aiida-core-version` in the CI test.

### DIFF
--- a/.github/actions/integration-tests/action.yml
+++ b/.github/actions/integration-tests/action.yml
@@ -11,13 +11,13 @@ inputs:
         description: GitHub Actions Runner image
         required: true
         type: string
+    aiidaCoreVersion: # Define aiida-core-version as an input
+        description: aiida-core version
+        required: true
+        type: string
 
 runs:
     using: composite
-
-    strategy:
-        matrix:
-            aiida-core-version: [2.3, 2.4.3]          # test on the latest and the oldest supported version
 
     steps:
         - name: Set jupyter token env
@@ -27,7 +27,7 @@ runs:
         - name: Run pytest to test image is working
           run: pytest tests_integration/test_image.py
           env:
-              TAG: aiida-${{ matrix.aiida-core-version }}
+              TAG: ${{ inputs.aiidaCoreVersion }}
           shell: bash
 
         # The Firefox and its engine geckodrive need do be installed manually to run
@@ -50,14 +50,14 @@ runs:
           if : ${{ inputs.architecture == 'amd64' }}
           run: pytest --driver Firefox tests_integration/test_app.py
           env:
-              TAG: aiida-${{ matrix.aiida-core-version }}
+              TAG: ${{ inputs.aiidaCoreVersion }}
           shell: bash
 
         - name: Run pytest for Chrome
           if : ${{ inputs.architecture == 'amd64' }}
           run: pytest --driver Chrome tests_integration/test_app.py
           env:
-              TAG: aiida-${{ matrix.aiida-core-version }}
+              TAG: ${{ inputs.aiidaCoreVersion }}
           shell: bash
 
         - name: Upload screenshots as artifacts
--- a/.github/actions/integration-tests/action.yml
+++ b/.github/actions/integration-tests/action.yml
@@ -15,13 +15,19 @@ inputs:
 runs:
     using: composite
 
+    strategy:
+        matrix:
+            aiida-core-version: [2.3, 2.4.3]          # test on the latest and the oldest supported version
+
     steps:
         - name: Set jupyter token env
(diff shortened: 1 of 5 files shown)


superstar54:
In aiidalab_qe, it requires the `aiida-quantumespresso~=4.4.0`, which needs aiida-core>=2.3. Thus I set the `aiida-core-version: [2.3, 2.4.3]`. This limits the power of the test. 
Hi @danielhollas @unkcpz , since you are more experienced with the package dependence. Any idea how to improve?

.github/actions/integration-tests/action.yml
```diff
@@ -15,13 +15,19 @@ inputs:
...
     steps:
         - name: Set jupyter token env
           run: echo "JUPYTER_TOKEN=$(openssl rand -hex 32)" >> $GITHUB_ENV
           shell: bash
 
         - name: Run pytest to test image is working
-          run: TAG=newly-baked pytest tests_integration/test_image.py
+          run: pytest tests_integration/test_image.py
+          env:
+              TAG: aiida-${{ matrix.aiida-core-version }}
```
danielhollas, line 30: Hmm, I don't think you want to change the tag here, since the purpose was to test the freshly-built QEApp image? I think it's fine to keep the integration tests as they were before, and only modify unit tests in `workflows/ci.yml`

danielhollas (changes requested):
> Thus I set the aiida-core-version: [2.3, 2.4.3]. This limits the power of the test.

That's fine for this PR, since we will soon bump the latest version to `2.5.0`.

.github/actions/integration-tests/action.yml
superstar54: the `integration-test` is an important test in the app. I think @unkcpz also wants to run it for a different aiida-core version. 
I can change it to 
```
aiida-core-version: [2.3, 2.4.3, "newly-baked"] 
```

superstar54:
Any idea why these three tests didn't run?

Screenshot from 2024-03-20 21-53-52

unkcpz:
Thanks @superstar54 

> Thus I set the aiida-core-version: [2.3, 2.4.3]. This limits the power of the test.

The oldest version set to `2.3` makes sense as you mentioned it compatible with the aiida-quantumespresso version we pinned. But can you try to set the lastest version to 2.5.0 and bump the required aiida-quantumespresso version? The initial idea was to see the tests pass for aiida-quantumespresso 4.4 and fail for 4.5 and manifest the problem of data-regression test.

.github/actions/integration-tests/action.yml
danielhollas: Okay, then I think 2.3 and newly-baked is enough, since the latest aiida core should be part of newly-baked.

danielhollas:
Maybe let's get this merged first and deal with the 2.5 version in a separate PR.

.github/workflows/docker-build-test-upload.yml
```diff
@@ -23,6 +23,9 @@ on:
 jobs:
     build-test-upload:
         runs-on: ${{ inputs.runsOn }}
+        strategy:
+            matrix:
+                aiida-core-version: [2.3, newly-baked]
```
danielhollas, line 28: Hmm, I really don't think you should add new build here, that doesn't make sense to me. Afaik the qeapp image should always be build on top of the lastest aiidalab docker stack.

CC @unkcpz

danielhollas (changes requested):
Erm, the I don't think the integration test changes are going to the right direction. The integration test is meant to test the newly build image, and it doesn't make sense to build an extra image with old aiida version just for this test. Let's stick with the changes to unit tests for now.

danielhollas:
For a simple integration test with an older image, you could just pull the existing image, e.g. aiidalab/full-stack:aiida-2.3.3, and test that you can install QeApp into it.

We have a test like that in aiidalab-docker-stack repo.
https://github.com/aiidalab/aiidalab-docker-stack/blob/main/tests%2Ftest-full-stack.py#L36

superstar54:
@unkcpz I tried passing the BASE_IMAGE variable directly through the command line when invoking docker buildx bake, thus allows us to bake the image using a different `aiida-core-version`. But the test failed, and I don't have time to investigate for the moment.

As suggested by @danielhollas , it is fine that this PR only adds the CI test. So I reverted the change on the `build-test-upload`.

danielhollas (approved):

PR #644, opened 2024-03-21
Regression test for the parameters generated by the app

superstar54:
The previous test used the final builder generated by the workchain, which is like an end-to-end test. However, the builder depends on the version of the aiida-quantumespresso, which we want to avoid for the moment.  In this PR, we only test the ui_parameters generated by the app. This ui_parameter is the most important parameter, and it contains all the information on the jobs. We can use it to recreate the whole qeapp instance. The test here guarantees that the ui_parameters are the same as the reference data.


danielhollas (approved):
Cool! I didn't realize you have these tests, might steal them for my app as well. :blush: 

I didn't check all the details but the overall approach and motivation makes sense to me.

unkcpz:
Emmm.... I got the idea to remove the builder generate part, but instead of removing it entirely it should be some where else. 
The initial implementation of data-regression test test two things, 1) The parameters are correctly passed from UI to widget 2) the parameters get by widgets can generate to the desired builder. 
In this PR you remove the test for the second part. It actually useful to spot things like the parameters are correctly overwritten, I remember it used to failed for some PR of @AndresOrtegaGuerrero and help us to find the issue of parameters passing.

Since this PR is merged, I'd suggest to bring the workflow test back, since the idea is borrow from aiida-quantumespresso to test the workchain. We implement the workchain here as well and having some logic inside, so make no sense to me to just remove it entirely.

superstar54:
> Since this PR is merged, I'd suggest to bring the workflow test back, since the idea is borrow from aiida-quantumespresso to test the workchain. We implement the workchain here as well and having some logic inside, so make no sense to me to just remove it entirely.

I agree that the test on the WorkChain part is also important. I opened an issue on this, #645. I think it's meaningless to move it somewhere else and disable the test. The test itself is not a good design and needs to be rewritten completely.

PR #647, opened 2024-03-26
Bump version v24.04.0a2 -> v24.04.0a3

superstar54:
When specifying `--minor` with bumper update, it tries to increment the minor version part of the project's current version. However, `v24.04.0a3` is not a valid next version based on the current versioning scheme.

## Manually Specify the Next Version
Since the automatic minor update didn't work as intended, I used the --set-version flag to manually specify the next alpha version. This approach gives direct control over the version naming:

```console
bumpver update --set-version v24.04.0a3 --dry
```


superstar54:
@danielhollas I followed the wiki, but there is an error. Could you check the description of this PR. If you agree, I will update the wiki.

danielhollas:
Yes, in the wiki the minor version bump was just an example. Here you want to bump the alpha version, not the minor version. bumbver has a special flag for that, I don't remember from the top of my head, please see bumpver update -h. Feel free to update the wiki with more examples.

danielhollas:
@superstar54 I think you'll have to recreate the branch to get rid of the extra merge commit. You cannot squash/rebase here otherwise the tag would get lost.

superstar54:
> I think you'll have to recreate the branch to get rid of the extra merge commit. You cannot squash/rebase here otherwise the tag would get lost.

Do you mean to merge the branch to the main directly?

danielhollas:
Yes, you will need to merge by creating a merge commit, otherwise you will loose the tagged commit.

PR #649, opened 2024-03-26
update docstrings and documentation consistent

### LINKED ISSUE #496: List and discuss the name for the plugin design
Thanks to the great work contributed by @superstar54, the plugin design allows new property developers can focus on the functionality and widgets needed for the property itself without knowledge of how other part of QeApp works. 
It decouples the settings of parameters of different calculations by introducing the "Panel" as the base widget class which has some common functionality to interact with other parts of the code.
There are some terminology and methods used but not intuitively straightforward to know what is for, such as "identifier", "get_panel_value" etc.
@superstar54, can you make a list of these names and we can have a discussion with @giovannipizzi? Once we make the stable release at the end of this month, it become a new API for developers and will be hard to change after.

superstar54:
fix #496


superstar54:
Hi @PNOGillespie, thanks for the valuable feedback. Could you please check this PR and see if it makes the text clearer to the user?

docs/source/development/plugin.rst
```diff
@@ -98,14 +98,15 @@ Beside, you need to override the the following methods:
             }
 
         def set_panel_value(self, input_dict):
-            """Set a dictionary with the input parameters for the plugin."""
+            """Set the value of the widgets in the panel from the input dictionary.
+            This method is called when the user wants to reload the panel from the previous calculation.
+            Or reset the panel to the default values."""
```
PNOGillespie, line 103:
```suggestion
            This method is called when the user wants to reload the panel from the previous calculation,
            or reset the panel to the default values."""
```

PNOGillespie (changes requested):
Hi @superstar54, this seems reasonable to me. I've suggested a small correction to the first docstring, but overall this does address the point I raised in the Issue page - nice work.

superstar54:
Thanks @PNOGillespie , I adapted your suggestion.

PR #655, opened 2024-04-02
Add HP code and LiCoO2 example

superstar54:


superstar54:
Already added in #650

PR #646, opened 2024-03-25
Add plugin list page

superstar54:
We foresee more plugins coming, and they may be developed by other developers and in other repositories. To help users easily find and manage the plugins, this PR adds a new plugin list page.

## Plugin registry
User can add their plugin entry point into the `plugins.yaml` file in this repository.

## Home page
On the home page, I added a new **Preference** section on the right side of the logo. For the moment, there is only the `plugin list` item, but we can add more user preference items in the future.

## Plugin list page
The UI has the following features:
- **Accordion Display for Plugins**: Each plugin is listed within an accordion-style component. This design keeps the interface clean and organized, allowing users to quickly scan through available plugins. The accordion's title bar displays essential information:
  - Plugin Name: Clearly indicates the name of the plugin for easy identification.
  - Installation Status: An icon indicator shows whether the plugin is currently installed.
- **Expandable Details**: Users can interact with the accordion to expand it, revealing more detailed information about each plugin.
- **Install button and remove button.**

Despite the `AppStore`'s extensive features, I opted against using it due to its complexity and lack of fit for this specific application.

## Demo
https://github.com/aiidalab/aiidalab-qe/assets/11457659/e1c37567-e1e6-486a-8083-906da8f5e851

## Future improvement
After this PR, open issues on
- [ ] version control, add `Update` button.
- [ ] add search bar

PR #657, opened 2024-04-03
Add job list page

### LINKED ISSUE #515: Add search page
In the last AiiDAlab meeting, we discussed the possible improvement for selecting (searching) the workchain. One good option would be the search page introduced by @cpignedoli .
In the case of aiidalab-qe, some special designs are needed:
- a way to show the properties calculated by the plugins.
- the search page itself could be a plugin, and we need to expose the entry point for the logo page of the QE.

Some other useful features:
- search by keywords (label, group, etc)

superstar54:
To help users easily search and manage their jobs, this PR adds a new job list page. This is the first step to fix #515 .

## Home page
On the home page, I added a new **Utils** section on the right side of the logo. Currently, there is only the `job list` item, but we can add more items in the future, e.g., 'plugin list' in #646 .

## Job list page
- show a table of the processes: pk, ctime, state, relaxed, properties
- **delete button and inspect button.**

## Demo

https://github.com/aiidalab/aiidalab-qe/assets/11457659/d3ddfce2-d32f-46fc-a7a1-0d1bcc98874e

## Future improvement
After this PR, open issues on
- [ ] add search bar


AndresOrtegaGuerrero:
@superstar54  Nice, I will test it and let you know, I was wondering. If an user has a WorkChain , maybe lets say a bands , but the structure as an input is coming from another WorkChain, and by deleting the first node it deletes automatically all the linked nodes. Is this avoided?

superstar54:
> If an user has a WorkChain , maybe lets say a bands , but the structure as an input is coming from another WorkChain, and by deleting the first node it deletes automatically all the linked nodes. Is this avoided?

Good question. I will look into it and provide more information to the user when deleting nodes.

AndresOrtegaGuerrero:
@superstar54 I would suggest you add a section explaning that by deleting a node (QeWorkChain) all the associated and link nodes will be eliminated , as well as other calculations that might be linked with this node.

delete.ipynb
```diff
@@ -0,0 +1,64 @@
...
+    "    load_profile()\n",
+    "    ## delete the group\n",
+    "    _, was_deleted = delete_nodes([pk], dry_run=False)\n",
+    "\n",
+    "    if was_deleted:\n",
+    "\n",
+    "        print(f'Node {pk} deleted.')\n",
+    "\n",
+    "# Ask for confirmation\n",
+    "if input(f'Do you really want to delete node {pk}? [y/N] ').strip().lower() == 'y':\n",
```
AndresOrtegaGuerrero, line 36: Also display the number nodes associated, so the person know what nodes or workchains are being deleted

superstar54:
> I really like the job list, is easy to search through jobs. I could be nice if there is a choice to add some filters like, filter by state : Finished[0] , or properties. I guess in this page the search app from edan could be useful.

Thanks, I added basic search filters. We can use @edan-bainglass 's query builder when it's ready in the future.

Screenshot from 2024-04-08 17-04-14

superstar54:
from @cpignedoli and @giovannipizzi 
- Add label/description for search.
- allow edit the label/description.

superstar54:
Not delete a job, but hide it.

delete.ipynb
superstar54: Yes, I display the number of nodes to be deleted.

superstar54:
Hi @AndresOrtegaGuerrero , I made the changes as you suggested. Here is the screenshot of the page when user deleting a node. Could you please have another look?

Screenshot from 2024-04-24 07-39-26

delete.ipynb
```diff
@@ -0,0 +1,114 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# AiiDAlab QauntumESPRESSO App\n",
+    "\n",
+    "<font color=\"red\"><b>Caution!</b></font> Deleting this job will also remove <b>all associated nodes</b>, including any related calculations and their respective results. This action is <b>irreversible</b>.\n"
```
AndresOrtegaGuerrero, line 9: `<font color="red"><b>Caution!</b></font> Deleting this job will also remove <b>all associated nodes</b>, including any workflows that share a common input structure, as well as various calculations and their respective results. <b>This action is irreversible.</b>`

delete.ipynb
```diff
@@ -0,0 +1,114 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# AiiDAlab QauntumESPRESSO App\n",
```
AndresOrtegaGuerrero, line 7: AiiDAlab QuantumESPRESSO App

job_list.ipynb
```diff
@@ -0,0 +1,57 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# AiiDAlab QauntumESPRESSO App\n",
```
AndresOrtegaGuerrero, line 7: # AiiDAlab QuantumESPRESSO App\n

AndresOrtegaGuerrero:
Hey Xing , it looks good to me , just check the few typos. I was wondering something, is it possible to check if the selected node to deete has other QeAppWorkChain associated ? and display this ? if not we can leave it for a future PR

delete.ipynb
superstar54: Hi Andres, thanks for the comment. 
> including any workflows that share a common input structure
It will not delete the workflows that share the input structure. It only deletes its descendants. 
- If QEApp jobs A and B use the same input structure, if A is deleted, B will not.
- If QEApp job A has an output structure and job B uses it as input, then if A is deleted, B will also be deleted. I added a check here, if this happened, user can not delete the job.

superstar54:
Hi @AndresOrtegaGuerrero , thanks for the review and suggestion. I have implemented the changes as you suggested.

> I was wondering something, is it possible to check if the selected node to deete has other QeAppWorkChain associated ? and display this ? if not we can leave it for a future PR


Yes, it's possible. I have added a check to see if there is a QEApp job using the outputs of the nodes that need to be deleted. If there is, the deletion will be blocked. Only after the user deletes the related QEApp job can the root one be deleted.

Screenshot from 2024-04-24 12-06-28

AndresOrtegaGuerrero (approved):
LGTM!

PR #663, opened 2024-04-04
add reminder text when users selecting property

### LINKED ISSUE #662: Tell people to check tabs when they activate the property checkboxes
Some plugins have setting panels, which will pop up when the user selects the property. However, people may not notice them. Tell people to check tabs when they activate some checkboxes like bands.

superstar54:
fix #662 

Here is a demo:

https://github.com/aiidalab/aiidalab-qe/assets/11457659/68b8190c-4f4d-4ff8-8561-6f606311360c


src/aiidalab_qe/app/configuration/workflow.py
```diff
@@ -94,9 +94,24 @@ def __init__(self, **kwargs):
...
+            if name in setting_entries:
+                reminder_text = ipw.HTML()
+            self.property_children.append(
+                ipw.HBox([self.properties[name], reminder_text])
+            )
+
+            # observer change to update the reminder text
+            def update_reminder_text(change, reminder_text=reminder_text):
+                if change["new"]:
+                    reminder_text.value = """(There is a setting panel available for this property on the top.)"""
```
AndresOrtegaGuerrero, line 109: What about like "Use the set panel for $property on top"

PR #667, opened 2024-04-08
assgin clean_workdir to plugin's builder

### LINKED ISSUE #666: `clean_workdir` failed for plugins
**Describe the bug**
If a plugin has `clean_workdir` be `True`, even user didn't select the `clean_workdir` box, the plugin still clean the workdir.

superstar54:
fix #666 

If a plugin has `clean_workdir` default to True, even if the user didn't select the `clean_workdir` box, the plugin still cleans the workdir. This PR assigns the `clean_workdir` to the plugin's builder to make sure the `clean_workdir` is consistent.

### DIFF
--- a/src/aiidalab_qe/workflows/__init__.py
+++ b/src/aiidalab_qe/workflows/__init__.py
@@ -158,20 +158,20 @@ class QeAppWorkChain(WorkChain):
         if properties is None:
             properties = []
         builder.properties = orm.List(list=properties)
+        # clean workdir
+        clean_workdir = orm.Bool(parameters["advanced"]["clean_workdir"])
+        builder.clean_workdir = clean_workdir
         # add plugin workchain
         for name, entry_point in plugin_entries.items():
             if name in properties:
                 plugin_builder = entry_point["get_builder"](
                     codes, structure, copy.deepcopy(parameters), **kwargs
                 )
+                # assume all builder has a input "clean_workdir"
+                plugin_builder.clean_workdir = clean_workdir
                 setattr(builder, name, plugin_builder)
             else:
                 builder.pop(name, None)
-
-        # XXX (unkcpz) I smell not proper design here since I have to look at
-        # configuration step to know what show be set here.
-        clean_workdir = parameters["advanced"]["clean_workdir"]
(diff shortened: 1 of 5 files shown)


src/aiidalab_qe/workflows/__init__.py
```diff
@@ -158,21 +158,21 @@ def get_builder_from_protocol(
...
+        # clean workdir
+        clean_workdir = orm.Bool(parameters["advanced"]["clean_workdir"])
+        builder.clean_workdir = clean_workdir
         # add plugin workchain
         for name, entry_point in plugin_entries.items():
             if name in properties:
                 plugin_builder = entry_point["get_builder"](
                     codes, structure, copy.deepcopy(parameters), **kwargs
                 )
+                # assume all builder has a input "clean_workdir"
```
AndresOrtegaGuerrero, line 170: Do we need to add something to the documentation for avoid a ext plugin to cause a fail?

superstar54: Thanks for pointing out this. Indeed, there is a risk here if the workchain dose not have a `clean_workdir` input. I just modified it to check if the plugin has a clean_workdir input.

AndresOrtegaGuerrero (approved):
LGMT!

AndresOrtegaGuerrero:
@superstar54 before you merge, do we have a test, to check if the builder set the clean_workdir ? in the workchains ?

superstar54:
> do we have a test, to check if the builder set the clean_workdir ? in the workchains ?

For all the plugins we have (bands, pdos, xas, xps etc), I think the default `clean_workdir` is False, so we can not make a test using them. But I test this using the `aiida-quantumespresso-hp`, which has a `clean_workdir` default to be True, and it works.

AndresOrtegaGuerrero:
> > do we have a test, to check if the builder set the clean_workdir ? in the workchains ?
> 
> For all the plugins we have (bands, pdos, xas, xps etc), I think the default `clean_workdir` is False, so we can not make a test using them. But I test this using the `aiida-quantumespresso-hp`, which has a `clean_workdir` default to be True, and it works.

Is it possible to just check the generated builder , to check when set clean_workdir is true ? is not then is ok , you can merge.

PR #669, opened 2024-04-10
check if the code is installed and usable

### LINKED ISSUE #664: Sharing Results from aiidalab_qe between users
I tried to share one of my aiidalab_qe results with a collaborator that has aiidalab_qe app for them to visualize the results.

Though using `verdi archive create my-calculations.aiida --nodes pk` i can share the QeWorkChain node, the collaborator indeed can have access in their aiida, when using the aiidalab_qe , there seems to be issues when loading the workchain. 
We should create test ok these kinds to improve the logic to facilitate sharing nodes

superstar54:
fix #664 

If code is imported from another user, it is not usable and thus will not be treated as an option in the `ComputationalResourcesWidget`.

### DIFF
--- a/src/aiidalab_qe/app/submission/__init__.py
+++ b/src/aiidalab_qe/app/submission/__init__.py
@@ -347,7 +347,11 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
 
         with self.hold_trait_notifications():
             for name, code in self.codes.items():
-                code.value = _get_code_uuid(codes.get(name))
+                # check if the code is installed and usable
+                # note: if code is imported from another user, it is not usable and thus will not be
+                # treated as an option in the ComputationalResourcesWidget.
+                if _get_code_uuid(codes.get(name)) in code.code_select_dropdown.options:
+                    code.value = _get_code_uuid(codes.get(name))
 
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -347,7 +347,11 @@ def _get_code_uuid(code):
 
         with self.hold_trait_notifications():
             for name, code in self.codes.items():
-                code.value = _get_code_uuid(codes.get(name))
+                # check if the code is installed and usable
+                # note: if code is imported from another user, it is not usable and thus will not be
+                # treated as an option in the ComputationalResourcesWidget.
+                if _get_code_uuid(codes.get(name)) in code.code_select_dropdown.options:
+                    code.value = _get_code_uuid(codes.get(name))
```
AndresOrtegaGuerrero, line 354: Is there any issue if code.value is not defined ?

superstar54: Thanks for the review. The code is a `ComputationalResourcesWidget`, which always has a `value` attribute.

AndresOrtegaGuerrero (approved):
LGTM! , I did test with another account and it works!

PR #683, opened 2024-04-24
Fix reminder_text

superstar54:
Only observe to update the `reminder_text` if there is a setting panel.

### DIFF
--- a/src/aiidalab_qe/app/configuration/workflow.py
+++ b/src/aiidalab_qe/app/configuration/workflow.py
@@ -89,7 +89,7 @@ class WorkChainSettings(Panel):
             value="moderate",
         )
         self.properties = {}
-        self.reminder_texts = {}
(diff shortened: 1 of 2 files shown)


src/aiidalab_qe/app/configuration/workflow.py
```diff
@@ -89,6 +89,7 @@ def __init__(self, **kwargs):
             value="moderate",
         )
         self.properties = {}
+        self.reminder_texts = {}
```
AndresOrtegaGuerrero, line 92: what about `reminder_info` ? what do you think ?

superstar54: Sure, it works for me. I changed it to `reminder_info`.

AndresOrtegaGuerrero (approved):
LGTM!

PR #688, opened 2024-04-25
Add bumpver pre-commit

danielhollas:
This is the last piece in the release workflow.

 The newly added `bumpver_pre_commit.sh` script is configured to run automatically by bumpver
 before it creates the release commit.
 We check for common mistakes, such as making a release commit
 in a wrong branch, or trying to push to a wrong remote.
 
 For now, only two checks are implemented:

 1. Check that the current branch name matches either `release/*` or `support/*`

 2. Check that the remote `origin` is pointing to the origin repository,
    and not a fork. Note however that this assumes that `origin` is the default remote
    where new branches are pushed. If the user configured a different default remote,
    this check will not save them in the current implementation.

## Future work:
  - make sure the main branch is up-to-date with origin/main
  - make sure the HEAD commit was branched off of main branch,
    although this rule should only apply to `release/*` branches, not `support/*` branches
  - check what is the default push branch 

 It would be great if at least some of these checks were handled by bumpver itself:

 Restricting releases from branch: https://github.com/mbarkhau/bumpver/issues/198
 Restricting releases to specified remote: https://github.com/mbarkhau/bumpver/issues/234

### DIFF
--- a/bumpver_pre_commit.sh
+++ b/bumpver_pre_commit.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# This script is configured to run automatically by bumpver
+# before it creates the release commit.
+# We check for common mistakes, such as making a release commit
+# in a wrong branch, or trying to push to a wrong remote.
+#
+# For now, only two checks are implemented:
+#
+# 1. Check that the current branch matches either release/* or support/*
+#
+# 2. Check that the remote 'origin' is pointing to the origin repository,
+#    and not a fork. Note however that this assumes that origin is the default remote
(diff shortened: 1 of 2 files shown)


superstar54 (approved):
Hi @danielhollas I looked it this morning. I am not very familiar with this. But LGTM!

danielhollas:
@superstar54 thanks for taking a look. Feel free to merge so you can test it while doing releases. 
I think you will want to recreate the rc2 release PR #701 anyway to pick up the most recent commits on main, so this might be a good opportunity for a test.

PR #695, opened 2024-04-28
use python API to setup code

superstar54:


superstar54:
This raises an error related to the thread, and I don't find a good solution for the moment.
```
Object '<DbUser at 0x7f08ab67a280>' is already attached to session '4' (this is '2')
```
Replaed this with #706 .

danielhollas:
I think the issue here might be that the User instanced is cached, so reloading the profile is not enough. You probably need to bust the cache manually somehow.

superstar54:
It seems that aside-core does not support creating a new session in a different thread.

danielhollas:
Maybe a question for @sphuber

unkcpz:
> It seems that aside-core does not support creating a new session in a different thread.

Exactly, we had a long discussion and struggling that we have to replace the direct use of aiida Node as trait for widgets to using UUID because of the similar problem. 

@superstar54 good if you can check this PR https://github.com/aiidalab/aiidalab-widgets-base/pull/543. The `_get_code` is the function that calls the setup_code if I remember correctly. I think the error "Object '<DbUser at 0x7f08ab67a280>' is already attached" is releted.

PR #699, opened 2024-04-29
Feature: Converge Thresholds 

AndresOrtegaGuerrero:
When doing a simulation with a system containing a higher number of atom , the convergence thresholds are quite lose 
image


src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -242,6 +294,21 @@ def _update_settings_from_protocol(self, protocol):
 
         self.kpoints_distance.value = parameters["kpoints_distance"]
 
+        num_atoms = 1
+        if self.input_structure is not None:
+            ase_structure = self.input_structure.get_ase()
+            num_atoms = ase_structure.get_global_number_of_atoms()
```
superstar54, line 300:
```suggestion
        num_atoms = len(self.input_structure.sites) if self.input_structure else 1
```

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -357,7 +441,7 @@ def set_metallic_magnetization(self, parameters):
 
     def set_panel_value(self, parameters):
         """Set the panel value from the given parameters."""
-
+        # protocol_parameters = PwBaseWorkChain.get_protocol_inputs(protocol)
```
superstar54, line 444: delete this line?

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -164,6 +164,51 @@ def __init__(self, default_protocol=None, **kwargs):
             (self.magnetization, "disabled"),
             lambda override: not override,
         )
+
+        # Convergence Threshold settings
+        self.scf_conv_thr = ipw.BoundedFloatText(
+            min=1e-15,
+            max=1.0,
+            step=0.000001,
```
superstar54, line 172:
```suggestion
            step=1e-10,,
```

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -164,6 +164,51 @@ def __init__(self, default_protocol=None, **kwargs):
...
+        self.scf_conv_thr.observe(self._callback_value_set, "value")
+        ipw.dlink(
+            (self.override, "value"),
+            (self.scf_conv_thr, "disabled"),
+            lambda override: not override,
+        )
+        self.forc_conv_thr = ipw.BoundedFloatText(
+            min=1e-15,
+            max=1.0,
+            step=0.00001,
```
superstar54, line 186:
```suggestion
            step=0.0001,
```

superstar54 (approved):
Hi @AndresOrtegaGuerrero , thanks for the work! I made a few small suggestions.

- The `step` of `scf_conv_thr` etc., is better to be similar to the default value.

I already approved the PR. Please make the changes and merge the PR. There is no need for me to review it again.

AndresOrtegaGuerrero:
> Hi @AndresOrtegaGuerrero , thanks for the work! I made a few small suggestions.
> 
> * The `step` of `scf_conv_thr` etc., is better to be similar to the default value.
> 
> I already approved the PR. Please make the changes and merge the PR. There is no need for me to review it again.

Hey could you explain this part `The `step` of `scf_conv_thr` etc., is better to be similar to the default value.`

AndresOrtegaGuerrero:
> > The user may want to change the value by mouse; in the case of SCF conv., when I click the up arrow, I assume the user wants to increase from 8e-10 to 9e-10, but it increases too much. As shown here:
> > qeapp-conv.mp4
> 
> I did the changes you suggested and it helps on this, however the issue is that the default values depend on the structure selected, since aiida-quantumespresso plugin set this based on the number of atoms of the structure , so for big systems the step of the widget will affect this differently ... I will try if i can modify the step after setting the structure

AndresOrtegaGuerrero:
> > however the issue is that the default values depend on the structure
> 
> Good point! Anyway, we provide a small value so the user can control it. Otherwise, it's impossible for the user to control it using a mouse.

I have an idea, just let me try something

superstar54 (approved):
LGTM!

PR #702, opened 2024-04-29
fix structures

superstar54:
Feedback from @cpignedoli : As a general note: let’s use experimental lattice parameters.
- [x] Silicon: rename to “bulk silicon”, change cell to be properly oriented (standardize with spglib)
- [x] Diamond structure is wrong (it’s graphene). Replace with 2-atom-cell diamond
- [x] Check SiO2 which cristobalite is (@Carlo) and put correct name (e.g. “β Quartz (SiO2)”)
- [x] Check GaAs, lattice parameter seems to be wrong cell as well
- [x] Gold: switch to primitive cell, also here use experimental lattice constant
- [x] Check Cobalt, use symmetrized/standardized primitive cell (and check lattice parameter). 
    - [ ] Provide guidance to the user on how to obtain the correct magnetic solution. This is beyond this PR.

For elements, using experimental lattice constants from:

 http://lampx.tugraz.at/~hadley/ss1/crystalstructure/structures/diamond/diamond.php
 https://periodictable.com/Properties/A/LatticeConstants.html

For SiO2 and GaAs, I downloaded the data from Materials Cloud MC3D and modified the lattice constants using experimental lattice constants.


superstar54:
Hi @cpignedoli, for our next release of QEApp, it's important to fix the structures as you suggested. Since you are probably busy, and we are trying to release the new version ASAP, so I made a quick fix for the structures, and I asked @AndresOrtegaGuerrero  to review this PR. I hope you don't mind.

AndresOrtegaGuerrero:
Is it ok that the atoms arent inside the cell? 

image

AndresOrtegaGuerrero:
When you select the Gold, the camera is really far 
image

AndresOrtegaGuerrero:
Should we clarify that Diamon is the primitive unit cell ?  as well as for the other systems ?

superstar54:
I found that if we use a primitive cell, the default view shows a dark sphere for the Au atom, which may be because the camera is too close. I discussed this with @AndresOrtegaGuerrero, and we think we better still use the conventional cell for Au to avoid confusing the users. I used the experimental lattice constant.

https://github.com/aiidalab/aiidalab-qe/assets/11457659/a17e78e1-9154-4c86-be76-041c6e4ee0fa

superstar54:
> Should we clarify that Diamon is the primitive unit cell ? as well as for the other systems ?

Sure, I added the primitive/conventional label for the systems.

AndresOrtegaGuerrero (approved):
LGTM! nice work!

superstar54:
Hi @cpignedoli, I've merged the PR for now. However, please feel free to suggest any additional enhancements/fixes you think are necessary.

cpignedoli:
@superstar54 @AndresOrtegaGuerrero I Apologize I had no time but this is perfect we cannot delay the release. In any case I will check everything and open issues in case of problems.

PR #706, opened 2024-04-30
use python script to setup all codes

superstar54:
It takes 1-2 sec to set up a code, and we have 14 codes (pw, dos, pp, ph, etc.), so around 20 sec in total. Which is not good, and it takes too long to load the container.

I looked into the issue and found that most of the time is used to set up the verdi command, profile and database. Thus it's would be good to combine all code set up as one.

Since the code setup is run in a different thread, thus, using the `InstalledCode` directly will fail when storing the code, as I tested in #695. This PR uses the `python -c` command to run in the `subprocess`. 

The total time to set up all 14 codes is now reduced to 3-5 seconds. Most importantly, setting more codes will not increase the time too much. **This is a temporary solution, but it is better than no solution.** We can use the `InstalledCode` in the future when we fix the thread problem.

### DIFF
--- a/src/aiidalab_qe/common/setup_codes.py
+++ b/src/aiidalab_qe/common/setup_codes.py
@@ -93,9 +93,11 @@ load_profile()
 
 
 def _generate_string_to_setup_code(code_name, computer_name="localhost"):
-    """Generate the python string to setup a code for a given computer."""
+    """Generate the Python string to setup an AiiDA code for a given computer.
+    Tries to load an existing code and if not existent,
+    generates Python code to create and store a new code setup."""
     try:
-        load_code(f"{code_name}-{QE_VERSION}@localhost")
+        load_code(f"{code_name}-{QE_VERSION}@{computer_name}")
         return ""
     except NotExistent:
         label = f"{code_name}-{QE_VERSION}"
--- a/src/aiidalab_qe/common/setup_codes.py
+++ b/src/aiidalab_qe/common/setup_codes.py
@@ -80,40 +80,59 @@ def codes_are_setup():
     return all(_code_is_setup(code_name) for code_name in CODE_NAMES)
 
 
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/common/setup_codes.py
```diff
@@ -80,40 +80,59 @@ def codes_are_setup():
...
+load_profile()
+
+"""
+    return header_code
+
+
+def _generate_string_to_setup_code(code_name, computer_name="localhost"):
+    """Generate the python string to setup a code for a given computer."""
     try:
         load_code(f"{code_name}-{QE_VERSION}@localhost")
```
AndresOrtegaGuerrero, line 98: Should it be `load_code(f"{code_name}-{QE_VERSION}@{computer_name}") `

AndresOrtegaGuerrero:
Should this PR be dependent on the` aiida-core` version?

src/aiidalab_qe/common/setup_codes.py
superstar54: Yes, you are right.

AndresOrtegaGuerrero:
This was my test 

Codes are setup!

real    0m7.560s
user    0m12.340s
sys     0m0.465s

superstar54:
Hi @AndresOrtegaGuerrero , thanks for the review. I made the changes as you suggested.

> Should this PR be dependent on the aiida-core version?

It could be if the Python API of AiiDA changes, e.g., `InstalledCode`, but for the moment, it should be safe.

superstar54:
> This was my test
Codes are setup!
real 0m7.560s
user 0m12.340s
sys 0m0.465s

Thanks for the test. You can try to run this in the main branch, and check the time difference.

AndresOrtegaGuerrero (approved):
LGTM, my test show less time

PR #711, opened 2024-04-30
fix code for xps, xas, and update the code resources

superstar54:
Update codes of plugin XAS and XPS:
- Fix the code
- Update the resources for the builder
- Select `xas` and `xps` in the test


src/aiidalab_qe/plugins/xas/workchain.py
```diff
@@ -13,6 +13,21 @@
 xch_elements = PSEUDO_TOC["xas_xch_elements"]
 
 
+def update_resources(builder, codes):
```
AndresOrtegaGuerrero, line 16: What if you use an extra function ?
def set_component_resources(component, code_info):
    if code_info:  # Ensure code_info is not None or empty
        component.metadata.options.resources = {
            "num_machines": code_info["nodes"],
            "num_mpiprocs_per_machine": code_info["ntasks_per_node"],
            "num_cores_per_mpiproc": code_info["cpus_per_task"],
        }
        if "parallelization" in code_info:
            component.parallelization = orm.Dict(dict=code_info["parallelization"])

and then in update_resources , you call this function ?

AndresOrtegaGuerrero: Actually , we are calling these lines plenty of times you can make like this function so it can be called by the other plugins , like , 

```
def update_resources(builder, codes):
    # Update resources for 'scf.pw'
    scf_pw_code_info = codes.get("pw")
    if scf_pw_code_info:
        set_component_resources(builder.core.scf.pw, scf_pw_code_info)
```
Then all the plugin can call this "util" function , set_component_resources ?? what do you think ?

superstar54: thanks for the suggestions, we can do this for the moment. In the future, we better use the `options` keyword to set up the resource when creating the builder.

AndresOrtegaGuerrero: the thing with options is that is restricted to the workchain and if it has the `options`,  you rather use the override

superstar54: I agree that `options` depends on how the plugin developer writes their `WorkChain.`

AndresOrtegaGuerrero (approved):
LGTM!

PR #714, opened 2024-05-02
Bump version v24.04.0rc2 -> v24.04.0rc3

superstar54:


superstar54:
Close as new commits have been added to the main branch.

superstar54:
Hi @danielhollas , in this case, I should close this PR and make a new PR from the latest main branch, right?

What if I want to make a release from a commit instead of the latest main? Is it possible to use the current release workflow?

danielhollas:
> Hi @danielhollas , in this case, I should close this PR and make a new PR from the latest main branch, right?

Yep

> What if I want to make a release from a commit instead of the latest main? Is it possible to use the current release workflow?

Sure, you just have to branch out from that commit. What use case would you have in mind then?

superstar54:
One question: After I run `bumpver update—-set-version v24.04.0rc3` and make a PR, I already see the branch and tag for `v2024.04.0rc3`, and I can also see it in the App Store before making the PR. 
I am confused about the meaning of merging the PR.

Issue #716, opened 2024-05-02
HubbardStructureData arent reachable in the AiiDA database widget


AndresOrtegaGuerrero:
Since DFT+U is now available in the QeApp, 
It is important to modify the logic of the AiiDAlab database widget to let the user  find HubbardStructureData , for example if an optimization (or other plugin) generates HubbardStructureData. 

We have to modify the logic but also double check that a user can use a HubbardStructureData, and being able to run other calculations with different the same structure with different Hubbard parameters or without them

PR #718, opened 2024-05-03
HubbardStructureData reach by StructureBrowserWidget

AndresOrtegaGuerrero:
This addresses #716

### DIFF
--- a/src/aiidalab_qe/workflows/__init__.py
+++ b/src/aiidalab_qe/workflows/__init__.py
@@ -147,7 +147,10 @@ class QeAppWorkChain(WorkChain):
                 )
 
             # Determine whether to store and use hubbard_structure based on conditions
-            if isinstance(structure, HubbardStructureData) and hubbard_structure.hubbard == structure.hubbard:
+            if (
+                isinstance(structure, HubbardStructureData)
+                and hubbard_structure.hubbard == structure.hubbard
+            ):
                 # If the structure is HubbardStructureData and hubbard parameters match, assign the original structure
                 builder.structure = structure
             else:
--- a/src/aiidalab_qe/workflows/__init__.py
+++ b/src/aiidalab_qe/workflows/__init__.py
@@ -129,9 +129,16 @@ class QeAppWorkChain(WorkChain):
         builder = cls.get_builder()
         # Set a HubbardStructureData if hubbard_parameters is specified
         hubbard_dict = parameters["advanced"].pop("hubbard_parameters", None)
+
+
+        # Determine whether to store and assign the new structure
+        should_store_and_assign = False
+        # Check if hubbard_dict is provided
         if hubbard_dict is not None:
             hubbard_parameters = hubbard_dict["hubbard_u"]
             hubbard_structure = HubbardStructureData.from_structure(structure)
+            
+            # Initialize on-site Hubbard values
             for key, value in hubbard_parameters.items():
                 kind, orbital = key.rsplit(" - ", 1)
                 hubbard_structure.initialize_onsites_hubbard(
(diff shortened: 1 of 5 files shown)


AndresOrtegaGuerrero:
@superstar54 , I tested with PdosWorkChain, it worked , i will switch this PR as a draft while we discuss what other test i should conduct but also if you would like to add and/or modify test for it

superstar54:
Hi @AndresOrtegaGuerrero is this ready for review, or it's a draft?

AndresOrtegaGuerrero:
@superstar54 This is a draft , i am testing it , but we have to change the logic of the workchain ( i am on it).

AndresOrtegaGuerrero:
@superstar54 Now is ready, please let me know what you think of the logic? Basically if you put an HubbardStructureData and you dont define new the option DFT+U in advanced tab, the workchain will just create a new StructureData (No Hubbard Parameters) from your HubbardStructureData, and run. If you define new parameter in the advanced tab, it will create a new HubbardStructureData with the new parameters.

superstar54 (approved):
Hi @AndresOrtegaGuerrero, I understand your logic in the workchain: If a user doesn't specify Hubbard parameters in the GUI, the workchain will use pure StructureData.

Maybe you could also think in another way: if a structure is generated by geometry optimization using certain Hubbard parameters. Then, it's better to do the following calculation using the same Hubbard parameters, which means it is better to pass the `HubbardStructureData` directly to the builder, to make the results are more consistent.
If you want to keep the logic, we can make the GUI load the Hubbard parameters from the structure if it's a HubbardStructureData.
You don't need to change it in this PR, but we can discuss it in the next relesae.

AndresOrtegaGuerrero:
@superstar54 I modified , so now , if you select a HubbardStructureData, and do not do any modifications , then the app sets DFT+U , and also upload the U values to the widgets. Also, in the workchain if the hubbard is exactly the same the workchain doesnt generate another HubbardStructureData but it uses the same. Have a look

AndresOrtegaGuerrero:
@superstar54 I did the changes you suggested, so now if you select a HubbardStructureData, the app will set your u values in the widget and activate override so in case you continue you will be using the same u parameters. In the workchain if the hubbard parameters set are exactly the same as in your structure, there wont be a new HubbardStructureData, but instead the user will do calculations on the same structure. Have a look

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -141,10 +141,22 @@ def get_builder_from_protocol(
                     hubbard_type="U",
                     use_kinds=True,
                 )
-            hubbard_structure.store()
-            builder.structure = hubbard_structure
+            if hubbard_structure.hubbard == structure.hubbard:
+                builder.structure = structure
+            else:
+                hubbard_structure.store()
+                builder.structure = hubbard_structure
```
superstar54, line 148: Raise error when `structure` is a normal StructureData, which doesn't have `hubbard`. I think you don't need check the `hubbard` here, because you need use the newly created `hubbard_structure`.
```suggestion
            hubbard_structure.store()
            builder.structure = hubbard_structure
```

AndresOrtegaGuerrero:
@superstar54 Should i add this logic in a function ? outside the class, so it is not no long the logic of the class ?

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -141,22 +147,27 @@ def get_builder_from_protocol(
...
-                hubbard_structure.store()
-                builder.structure = hubbard_structure
 
-        # No Hubbard parameters specified
+            # Check if structure matches the new hubbard structure and assign accordingly
+            should_store_and_assign = (
+                not isinstance(structure, HubbardStructureData)
+                or hubbard_structure.hubbard != structure.hubbard
+            )
+
```
superstar54, line 156: I don't understand why we need to check `hubbard_structure.hubbard != structure.hubbard`?

superstar54 (changes requested):

src/aiidalab_qe/workflows/__init__.py
AndresOrtegaGuerrero: because if the user is setting the same hubbard parameters, then i do not store the HubbardStructureData, and just give the builder the same structure, otherwise i will create a copy of the same HubbardStructureData. But there is also the other case where the user uses a HubbardStructureData, but decided to change the U parameters, then i should create a new structure and store it.

superstar54: I see. thanks!

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -141,10 +147,27 @@ def get_builder_from_protocol(
...
+            new_structure.store()
+            builder.structure = new_structure
         else:
             builder.structure = structure
+
+        # Common logic to store and assign the new structure if conditions are met
+        if should_store_and_assign and "hubbard_structure" in locals():
+            hubbard_structure.store()
+            builder.structure = hubbard_structure
+
```
superstar54, line 170: one more question. why not move to line 156?

AndresOrtegaGuerrero: Done

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -141,10 +145,35 @@ def get_builder_from_protocol(
...
+                if hubbard_structure.hubbard != structure.hubbard:
+                    should_store_and_assign = True
+                else:
+                    # If criteria is not met, assign the structure to the builder
+                    builder.structure = structure
+
+            # Common logic to store and assign hubbard_structure if conditions are met
+            if should_store_and_assign:
+                hubbard_structure.store()
+                builder.structure = hubbard_structure
```
superstar54, line 166: this logic can be simplified.
```suggestion
            # If this is a new HubbardStructureData or the hubbard is different
            if not isinstance(structure, HubbardStructureData) or hubbard_structure.hubbard != structure.hubbard:
                hubbard_structure.store()
                builder.structure = hubbard_structure
            else:
                builder.structure = structure
```

PR #720, opened 2024-05-03
fix parameter retrieval in method `_update_process_label`

superstar54:
This PR uses the `get()` method and provides default values to avoid crashing when the `input_parameters` is an empty dict.


AndresOrtegaGuerrero:
@superstar54 have a look in the conflict, i will approve 👍

AndresOrtegaGuerrero (approved):
LGTM! thank you nice work!

PR #725, opened 2024-05-10
cleaning 2D path logic

AndresOrtegaGuerrero:
This PR solves an issue that can occurs with working with oblique cell, where due to rounding number it will not include a k-point. 

If you want to reproduce the issue with the current app, use 
[structure-not-stored (25).cif.zip](https://github.com/aiidalab/aiidalab-qe/files/15273135/structure-not-stored.25.cif.zip)


superstar54 (approved):
Thanks for the fix!

PR #729, opened 2024-05-15
Explicitly pinning down the full-stack image version for qe image

unkcpz:


superstar54 (approved):
Thanks for the improvement, LGTM!

AndresOrtegaGuerrero:
@unkcpz @superstar54 the aiidalab_qe now shows problems with the arm64 in the PRs , from what we discussed the issue is related to the docker_image

PR #730, opened 2024-05-15
RWIP: docker workflow refactor GHA

unkcpz:
An analogous refactor of https://github.com/aiidalab/aiidalab-docker-stack/pull/439 by @danielhollas 

- Not uploading image as artifacts.
- Build multiplatform images together

Since this repo only build one image, I slightly change the `extract-image-names.sh` to `extract-image-name.sh` to get only single image name.


unkcpz:
> Not uploading image as artifacts.

As you mentioned as a side effect in https://github.com/aiidateam/aiida-core/pull/6388, this becomes quite annoying for qeapp, since the test on the built image help a lot with speed up the integration test and we want test happened for the PR from forked repo. 

Here I'll probably do a workaround by both upload as artifacts and upload to ghcr.io if it is from the repo. The test will use the image from the artifact for the moment, and the image publish will use the image from ghcr.io

Maybe there will be potential security issue if we allow PR to push image to registry. Let's image a case that someone maliciously create an image that has racist remarks included, it will be able to be download under `ghcr.io/aiidalab/`. 
But I think we already some restrictions that for the first contribution PR it has to be approved to be run with CI. So we can instead of using github token but using the stored account and secret to make it able to upload to the `ghcr.io`? What you think?

danielhollas:
Hmm, that's tricky. I understand that indeed for integration tests you'd want to use the build image. But uploading to artifacts slows things down and makes everything much more complex. 

I am not sure if there is a good solution currently, I would probably simply suggest to push to origin. :-) Pushes from forks should still work, but will skip the image build and integration tests.

unkcpz:
Okay, I think I can bind the build and test as what was happened before. Then I only separate the upload out and upload to `ghcr.io`.

danielhollas:
Yeah, all valid points. I don't think I have anything more clever to say, it's a tradeoff.

danielhollas:
Superseded by #776

PR #744, opened 2024-05-23
Install plugin with `--user` option for persistence

superstar54:
To ensure the package remains installed after the server restarts, install the package into the home directory using the `--user` option. This directory is mounted as a  Docker volume, ensuring the package is preserved across server restarts.


AndresOrtegaGuerrero (approved):
LGTM! thank you!

PR #751, opened 2024-06-24
Redesign header with logo, toggled info sections, etc.

edan-bainglass:
This PR replaces the current welcome banner with the "official" app logo, a "happy computing" subtitle, and a set of toggle buttons to show/hide info sections - an "About" section (the old welcome banner), and a "Guide" section with links to online documentation and a checkbox to trigger in-app guide widgets (see below).

### Other changes:

- App now takes the full width of the notebook
- (lazy loading) - header/footer components load prior to app with a placeholder "loading" message while the app is being loaded

###  In-app guides

The purpose of these guide widgets is to provide an on-demand example use-case for new users to follow. In-app guide widgets can be added anywhere as follows:

```python
example = ipw.VBox(
    children=[
        ipw.HTML("This is a test guide"),
    ]
)
example.add_class("in-app-guide")
```

The `in-app-guide` CSS class sets the `display` property to `none`. The checkbox in the "Guide" section toggles the `display` property "on/off" via a `show` CSS class.


superstar54 (changes requested):
Thanks, @edan-bainglass for the new header design. Very nice!
Few suggestions.
- Better to move the four-steps instruction in the `About` section to the `Guide` section. Instead, you can add the the more general info, e.g., what the capability of the app (Bands, PDOS, XPS, IR/Raman, etc).
- Regarding the `in-app guide`, we can discuss on detail implementaion . For this PR, could you hide the `show in-app guide` checkbox and related.
- The custom CSS is kind of advanced topic, and would be very useful if you add some docs on it (e.g., the development guide).

danielhollas:
@edan-bainglass LMK once this is ready and I'll have a closer look (just a note that I am at the conference next week, so will get it no sooner than July 14th.

tests/test_infobox.py
```diff
@@ -0,0 +1,53 @@
+from pathlib import Path
+
+from aiidalab_widgets_base.infobox import FirstTimeUserMessage, InAppGuide, InfoBox
```
AndresOrtegaGuerrero, line 3: should this be from `aiidalab_qe.common.infobox` ?

edan-bainglass:
@danielhollas @superstar54 ready for review

edan-bainglass:
> Thanks, @edan-bainglass for the new header design. Very nice! Few suggestions.
> 
> * Better to move the four-steps instruction in the `About` section to the `Guide` section. Instead, you can add the the more general info, e.g., what the capability of the app (Bands, PDOS, XPS, IR/Raman, etc).
> * Regarding the `in-app guide`, we can discuss on detail implementaion . For this PR, could you hide the `show in-app guide` checkbox and related.
> * The custom CSS is kind of advanced topic, and would be very useful if you add some docs on it (e.g., the development guide).

@superstar54 your comments have all been addressed. Note that the last one regarding documenting CSS use - the PR introducing the CSS loader in AWB (https://github.com/aiidalab/aiidalab-widgets-base/pull/624) adds docs there regarding CSS.

Good to merge?

qe.ipynb
```diff
@@ -71,37 +73,38 @@
...
+    "view.main.children = [app_with_work_chain_selector]"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "controller.enable_toggles()"
```
superstar54, line 107: Why only enable the toggler at the last? Is it possible to show the `Guide` and `About` when loading the main app?

superstar54 (approved):
Hi @edan-bainglass , thanks for the work! I added two mirror change requests. Ready to go after you fix them.

qe.ipynb
superstar54: Indeed, this means that the user can not do anything when loading the main app. I hope we can find some solution in the future.

edan-bainglass: Make the app load faster 😅 Working on it

PR #752, opened 2024-06-24
CI: Use uv for installing deps

danielhollas:
Use `uv` installer to speed up installing dependencies for tests + some other CI tweaks brought over from AWB:

- Use FORCE_COLOR to get colored logs in Github UI
- Update actions
- Don't collect code coverage from tests/ folders

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 ---
 # Run basic tests for this app
 
-name: continuous-integration
+name: CI
 
 on: [push, pull_request]
 
@@ -11,10 +11,12 @@ concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: true
 
+env:
+    FORCE_COLOR: 1
+
 jobs:
 
     test-package:
-
 
         strategy:
             matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -35,26 +37,26 @@ jobs:
...
+            - name: Install uv
+              run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.2.13/uv-installer.sh | sh
+
             - name: Install package
-              run: pip install -e .[dev] aiida-core==${{ matrix.aiida-core-version }}
+              run: uv pip install --system -e .[dev] aiida-core==${{ matrix.aiida-core-version }}
 
             - name: Run pytest
-              run: pytest -v tests --cov
+              run: pytest -v tests --cov=aiidalab_qe
```
superstar54, line 54: Why add `aiidalab_qe`?

.readthedocs.yaml
```diff
@@ -10,19 +10,10 @@ build:
...
-python:
-    install:
-        - requirements: docs/requirements.txt
+    commands:
+        - asdf plugin add uv
+        - asdf install uv 0.2.13
+        - asdf global uv 0.2.13
+        - uv venv
+        - uv pip install -r docs/requirements.txt
+        - .venv/bin/python -m sphinx -W --keep-going -d _build/doctrees -D language=en -b html docs/source $READTHEDOCS_OUTPUT/html
```
superstar54, line 19: Are there big benefits to this change? Otherwise it's better to use the one from the official readthedocs docs.

superstar54 (changes requested):
Thanks @danielhollas  for the work. I don't understand some of the changes, could you address my comments?

.github/workflows/ci.yml
danielhollas: Citing from `pytest --help`

```
  --cov=[SOURCE]   Path or package name to measure during execution.                                                 
                                 Use --cov= to not do any source filtering and record everything.
```

Without specifying the package there are two problems:
1. (minor) Coverage of `tests/` folder is collected, which skews the overall coverage percentage.
2. If you don't specify the package, some files from the `aiidalab_qe` package might be omitted from the statistics.

danielhollas: Note that `aiida-core` uses the same approach.

.readthedocs.yaml
danielhollas: It speeds up the build from ~30s to 5s. The configuration suggested here actually comes from the official docs https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv

But it would be fine to revert this change if you don't like it.

.github/workflows/ci.yml
danielhollas: Here's how things look in Codecov without this change:

image

https://app.codecov.io/gh/aiidalab/aiidalab-qe

superstar54 (approved):
Thanks @danielhollas  for the detailed explanation! Nice work!

PR #755, opened 2024-06-29
Feature/post install plugins

mikibonacci:
With the following PR we address the post installation for plugins. 
This can be done by setting a post install key in the plugin.yaml file. 

Additionally, the list of plugins is taken from the local file, not from the one in the github repo (main branch).
In this way we do not risk the add plugins which are not compatible with the given QE app version.


mikibonacci:
Still missing: documentation on how to provide the post_install logic for a given plugin. I will provide the documentation in a next commit on this branch, hopefully by this weekend.

mikibonacci:
Hi @superstar54 , the PR is ready. Do you think we can merge it and make a new prerelease of the app by today? Thanks!

superstar54 (approved):
Hi @mikibonacci , thanks for the nice work!

Issue #756, opened 2024-06-29
Case sensitive filenames breaks in case insensitive filesystems.


mikibonacci:
(base) TheFlash:codes mbonacci$ git clone https://github.com/mikibonacci/aiidalab-qe.git
Cloning into 'aiidalab-qe'...
remote: Enumerating objects: 5372, done.
remote: Counting objects: 100% (379/379), done.
remote: Compressing objects: 100% (231/231), done.
remote: Total 5372 (delta 191), reused 254 (delta 110), pack-reused 4993
Receiving objects: 100% (5372/5372), 9.43 MiB | 1.18 MiB/s, done.
Resolving deltas: 100% (3273/3273), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'src/aiidalab_qe/app/structure/examples/Diamond.cif'
  'src/aiidalab_qe/app/structure/examples/diamond.cif'

This gives me problem when I try to clone the repo in my Mac. So, we should rename one of the diamond cif.

mikibonacci:
closed with https://github.com/aiidalab/aiidalab-qe/pull/757

PR #757, opened 2024-06-29
Deleting the diamond.cif file

mikibonacci:
This will give case-insensitive problems with Diamond.cif - this will happen only for case-insensitive filesystems like my Mac. The problem is that then is it not possible to clone the repo without having to choose (git does it) among the two files.

The diamond.cif file seems anyway not used, so we do not need it.

This will solve https://github.com/aiidalab/aiidalab-qe/issues/756.


superstar54:
Hi @mikibonacci , thanks for the fix!

The test failed because of the wrong `setup.cfg` in one of the dependence `aiidalab-eln.` I made a [PR] to `aiidalab-eln` (https://github.com/aiidalab/aiidalab-eln/pull/85) to fix it.

superstar54 (approved):

PR #760, opened 2024-07-02
Test with aiida-core=2.6.1 and python 3.11

danielhollas:


.github/workflows/ci.yml
```diff
@@ -45,15 +44,13 @@ jobs:
...
+              run: curl --proto '=https' --tlsv1.2 -LsSf https://github.com/astral-sh/uv/releases/download/0.2.28/uv-installer.sh | sh
 
             - name: Install package
               run: uv pip install --system -e .[dev] aiida-core==${{ matrix.aiida-core-version }}
 
             - name: Run pytest
-              run: pytest -v tests --cov=aiidalab_qe
-              env:
-                  TAG: ${{ matrix.tag }}
+              run: pytest -v tests ${{ matrix.aiida-core-version == '2.3' && '--cov=aiidalab_qe' || '' }}
```
danielhollas, line 53: For some reason, running tests with aiida-core >=2.4 takes significantly more time (by 2 minutes), see #766. We can speed up these tests by not collecting the code coverage, i.e. we only collect code coverage with aiida-core==2.3. 
With this change the test times are roughly equal.
(note we use a similar trick in aiida-core)

pyproject.toml
```diff
@@ -4,3 +4,14 @@ requires = [
     "wheel"
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.pytest.ini_options]
+addopts = '--strict-config --strict-markers --durations=30 --durations-min=1 -ra'
+filterwarnings = [
```
danielhollas, line 10: I took these from AWB. There are definitely more warnings that either have to be fixed or ignored.

superstar54 (approved):
Thanks fo the work!. LGTM.

PR #761, opened 2024-07-02
Use uv to install deps during docker build

danielhollas:
This cuts down the build time by a full minute.

### DIFF
--- a/.github/workflows/docker-build-test-upload.yml
+++ b/.github/workflows/docker-build-test-upload.yml
@@ -38,7 +38,7 @@ jobs:
             # Also, they might have running or stopped containers,
             # which are not cleaned up by `docker system prun`
             - name: Reset docker state and cleanup artifacts 🗑️
-              if: ${{ inputs.platform != 'x86_64' }}
+              if: ${{ startsWith(inputs.runsOn, 'ubuntu') }}
               run: |
                   docker kill $(docker ps --quiet) || true
                   docker rm $(docker ps --all --quiet) || true
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,5 @@
 # syntax=docker/dockerfile:1
+FROM ghcr.io/astral-sh/uv:0.2.18 as uv
 FROM base-image
 
 # Copy whole repo and pre-install the dependencies and app to the tmp folder.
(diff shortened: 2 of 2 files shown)


.github/workflows/docker-build-test-upload.yml
```diff
@@ -38,7 +38,7 @@ jobs:
             # Also, they might have running or stopped containers,
             # which are not cleaned up by `docker system prun`
             - name: Reset docker state and cleanup artifacts 🗑️
-              if: ${{ inputs.platform != 'x86_64' }}
+              if: ${{ startsWith(inputs.runsOn, 'ubuntu') }}
```
superstar54, line 41: The new changes only support `ubuntu`?

docker/Dockerfile
```diff
@@ -1,4 +1,5 @@
 # syntax=docker/dockerfile:1
+FROM ghcr.io/astral-sh/uv:0.2.18 as uv
```
superstar54, line 2: `FROM` two images? And nothing is done for the first one?

superstar54 (changes requested):
Hi @danielhollas thanks for the PR. A few question on the changes, please see my comments.

.github/workflows/docker-build-test-upload.yml
danielhollas: This step of resetting Docker is only needed on our self-hosted arm64 runner. This if statement just ensures that this step is skipped run on the official x86 runner.

In a next PR I am going to remove this altogether since we want to stop using the self-hosted runner.

docker/Dockerfile
danielhollas: This first FROM is only used to access the uv binary. Please see the link to the uv documentation that I provided in a comment below, which is where this pattern comes from.

Essentially, this is to avoid the uv binary being in the final image.

superstar54 (approved):
Thanks for the explanation and the excellent work!

.github/workflows/docker-build-test-upload.yml
danielhollas: Ah, made a mistake here, the if condition is inverted, it should be 

```suggestion
              if: ${{ ! startsWith(inputs.runsOn, 'ubuntu') }}
```

This is fixed in #764

PR #762, opened 2024-07-02
Dockerfile: Add --no-user to pip install

danielhollas:
This is an important fix for the Docker build since version [v2024.1017](https://github.com/aiidalab/aiidalab-docker-stack/releases/tag/v2024.1017) of the aiidalab full-stack image, which added the `--user` argument to `pip install` by default.

@superstar54 it seems that the aiidalab_qe app dependencies were being installed in `~/.local` instead of `/opt/conda`. This may have influenced your testing of your tared-home image, as it would blow up the home size.

CC @unkcpz 

NOTE: As an alternative to this PR you can merge #761 and use uv instead of pip which speeds up the docker build by 1 minute.


unkcpz:
I think if we go with the zip and decompress solution, the time won't influenced by where it installed. 

>  This may have influenced your testing of your tared-home image, as it would blow up the home size.

But this is true! @superstar54 can you add the change to your PR and check the image size change? 

> As an alternative to this PR you can merge https://github.com/aiidalab/aiidalab-qe/pull/761 and use uv instead of pip which speeds up the docker build by 1 minute.

If we go with prepare a ready to use home and decompress it for user, we don't need pre-install the dependencies, true? Then I think it is better to install things as what will exactly happened as `aiidalab-launch install`.

danielhollas:
> If we go with prepare a ready to use home and decompress it for user, we don't need pre-install the dependencies, true? Then I think it is better to install things as what will exactly happened as aiidalab-launch install.

Yeah, you don't need to, though I guess it still helps a bit since it will make home much smaller, and hence untar/decompress should be faster (and the overall image size smaller as well).

danielhollas:
Closed in favour of #761

PR #764, opened 2024-07-03
Speedup CI and Docker build

danielhollas:
Bunch of changes and suggestions to speed up CI and Docker build (including local Docker build). Together these changes drop the Docker build/test time to 8 minutes.

Built on top of #761.

@unkcpz @superstar54 given your work on the QE image, please take a look, hopefully this should improve your iteration speed.


.github/workflows/docker-build-test-upload.yml
```diff
@@ -38,41 +39,50 @@ jobs:
             # Also, they might have running or stopped containers,
             # which are not cleaned up by `docker system prun`
             - name: Reset docker state and cleanup artifacts 🗑️
-              if: ${{ inputs.platform != 'x86_64' }}
+              if: ${{ ! startsWith(inputs.runsOn, 'ubuntu') }}
```
danielhollas, line 42: This was a bug, `inputs.platform` does not exist and so this step was always performed, even on Github Actions runner.

.github/workflows/docker-build-test-upload.yml
```diff
@@ -38,41 +39,50 @@ jobs:
...
 
-            - name: Upload image as artifact 💾
+            - name: Upload image as artifact 📤
               uses: actions/upload-artifact@v4
               with:
                   name: ${{ inputs.image }}-${{ inputs.architecture }}
                   path: /tmp/aiidalab/${{ inputs.image }}-${{ inputs.architecture }}.tar
                   retention-days: 3
-              if: always()
+              if: ${{ !github.event.pull_request.head.repo.fork }}
```
danielhollas, line 82: We can skip artifact upload when running from forks since the subsequent jobs that use it are also skipped.

docker/Dockerfile
```diff
@@ -1,14 +1,27 @@
 # syntax=docker/dockerfile:1
+FROM ghcr.io/astral-sh/uv:0.2.18 as uv
 FROM base-image
 
+USER ${NB_USER}
+
+ENV QE_VERSION ${QE_VERSION}
+RUN mamba create -p /opt/conda/envs/quantum-espresso --yes \
```
danielhollas, line 8: It's better to put this step at the top so that it can be cached.

.github/workflows/ci.yml
```diff
@@ -51,9 +50,7 @@ jobs:
               run: uv pip install --system -e .[dev] aiida-core==${{ matrix.aiida-core-version }}
 
             - name: Run pytest
-              run: pytest -v tests --cov=aiidalab_qe
-              env:
-                  TAG: ${{ matrix.tag }}
+              run: pytest -v tests ${{ matrix.aiida-core-version == '2.3' && '--cov=aiidalab_qe' || '' }}
```
danielhollas, line 53: I noticed that newer aiida-core versions take much longer time to run the test suite. I've opened #766 to investigate. For now we can speed them up by not collecting code coverage for them. This makes their speed roughly equivalent.

.github/actions/integration-tests/action.yml
```diff
@@ -24,27 +24,6 @@ runs:
...
-              firefox-version: '96.0'
-
-        - name: Install geckodriver
-          if : ${{ inputs.architecture == 'amd64' }}
-          run: |
-              wget -c https://github.com/mozilla/geckodriver/releases/download/v0.30.0/geckodriver-v0.30.0-linux64.tar.gz
-              tar xf geckodriver-v0.30.0-linux64.tar.gz -C /usr/local/bin
-          shell: bash
-
-        - name: Run pytest for firefox
```
danielhollas, line 43: As long as we've beenrunning these UI tests in AWB and my app, we've never caught any bug that would be specific for Firefox. 

Given Firefoxe's low usage, I think we can skip it here and only test with Chrome. Since these tests do not run in parallel this saves a fair amount of time (>2min). And it also makes the workflow simpler. We'll still continue testing Firefox in AWB to make sure individual widgets work there.

(but suggestion may be controversial)

.github/workflows/docker-build-test-upload.yml
```diff
@@ -38,41 +39,50 @@ jobs:
...
+              uses: docker/bake-action@v5
               with:
-                  architecture: ${{ inputs.architecture }}
-                  runsOn: ${{ inputs.runsOn }}
+                  # Load locally into Docker for testing
+                  load: true
+                  workdir: docker/
+                  set: |
+                      *.platform=linux/${{ inputs.architecture }}
+                      *.cache-to=type=gha,scope=${{ github.workflow }},mode=min
```
danielhollas, line 63: Added caching, saves around 30s. (larger gains are not possible since we're copying the entire repo in the image which invalidates the cache on most on any changes)

.github/workflows/ci.yml
```diff
@@ -20,8 +20,7 @@ jobs:
 
         strategy:
             matrix:
-                tag: [latest]
-                python-version: ['3.9', '3.10']
+                python-version: ['3.9', '3.11']
```
danielhollas, line 23: We're going to skip the 3.10 version so let's start testing with 3.11

unkcpz:
Thanks! @superstar54 said he will check it, I remove myself from reviewer list at the moment.

danielhollas:
Parts of this are now in other PRs so will close this one.

PR #768, opened 2024-07-08
Bump version v24.10.0a0 -> v24.10.1a0

mikibonacci:


unkcpz:
Thanks @mikibonacci I guess it should be `24.10.0a1`. You can use `bumpver update --tag-num -t alpha --dry` if I remember correctly.

PR #773, opened 2024-07-17
Rewamp Docker file for home archiving

danielhollas:
Build on top of #764 
Alternative to #740 

The main goal here is to reduce the complexity of status quo and of #740.

- The strategy of archiving home directory and untarring it at startup actually allows for a lot of simplification of the Dockerfile since everything can be directly prepared in home folder, without intermediary steps, and this allows to get rid of the current startup scripts (70_*, 71_*)
- All startup scripts from full-stack are preserved and reused, which minimizes duplication, resolves the SSH key issue and should be more maintainable
- The only new startup script is the `00_untar_home.sh` which is basically the same here as in #740.
- I've also done some unrelated cleanup:
  - Moved Dockerfile to the root dir, and got rid of `build.json` and  `docker-bake.hcl` which were introducing unnecessary complexity. One can now build the image by simply running `docker build . -t aiidalab/qe:newbuild` . This also allowed me to use podman on my dev machine, which otherwise doesn't support the buildx backend.

I've done some quick benchmarking, at starting the container takes around 12s on my machine. The image takes around 6.5Gb. We could trade around 300Mb images size for extra 3s of startup time if we compressed the home.tar archive. (My timings seems roughly consistent with those observed in #740.

@unkcpz @superstar54 Let me know what you think. I'd recommend to ignore the noise in the diff and just look at the new Dockerfile to see if that makes sense.

I am happy to split this up to smaller pieces if somebody will actually review them. :-P


Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+USER root
+COPY ./before-notebook.d/* /usr/local/bin/before-notebook.d/
+# 6. Copy the whole repo
+COPY --chown=${NB_UID}:${NB_GID} . ${QE_APP_FOLDER}
+# Remove all untracked files and directories.
+RUN git clean -dffx
+
+RUN fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
```
danielhollas, line 48:
```suggestion

# 6. Copy the whole repo
COPY --chown=${NB_UID}:${NB_GID} . ${QE_APP_FOLDER}
# Remove all untracked files and directories.
RUN git clean -dffx

USER root
COPY ./before-notebook.d/* /usr/local/bin/before-notebook.d/RUN fix-permissions "${CONDA_DIR}" && \
    fix-permissions "/home/${NB_USER}"
```

.github/workflows/ci.yml
```diff
@@ -51,9 +50,7 @@ jobs:
               run: uv pip install --system -e .[dev] aiida-core==${{ matrix.aiida-core-version }}
 
             - name: Run pytest
-              run: pytest -v tests --cov=aiidalab_qe
-              env:
-                  TAG: ${{ matrix.tag }}
+              run: pytest -v tests ${{ matrix.aiida-core-version == '2.3' && '--cov=aiidalab_qe' || '' }}
```
unkcpz, line 53: This is a bit confusing can you explain what it achieved? Is it equivalent as original one?

Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ENV QE_APP_FOLDER=${AIIDALAB_APPS}/quantum-espresso
+# 2. Copy files needed for installing stuff, the rest is copied at the end for better caching
+COPY --chown=${NB_UID}:${NB_GID} src/ ${QE_APP_FOLDER}/src
+COPY --chown=${NB_UID}:${NB_GID} setup.cfg pyproject.toml *yaml README.md ${QE_APP_FOLDER}
+
+WORKDIR "${QE_APP_FOLDER}"
+# 3.Install python dependencies
+# Use uv instead of pip to speed up installation, per docs:
+# https://github.com/astral-sh/uv/blob/main/docs/guides/docker.md#using-uv-temporarily
+# Use the same constraint file as pip
```
unkcpz, line 26: Nice stuff!

before-notebook.d/00_untar_home.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#
+home="/home/${NB_USER}"
+
+if [[ ! -d ${home} ]]; then
+    echo "Directory $home does not exist!"
+    exit 1
+fi
```
unkcpz, line 8: I think this check exist in @superstar54's implementation, but why it is necessary, when is the home can be empty?

.github/workflows/docker.yml
```diff
@@ -1,50 +1,13 @@
 ---
-name: Build, and push Docker Image
+name: Docker Image
 
 on:
     pull_request:
-        paths:
```
danielhollas, line 6: Since we're copying the full contents of  the repo in the image, I don't think we can do any filtering here...

unkcpz (commented):
Thanks @danielhollas. Just go through the PR, some minor comments, let's discuss tomorrow.

.github/workflows/ci.yml
danielhollas: Yeah, I can add a comment, this is explained more in #764 on which this PR builds. Could you please take a look and review that separately? This just means that tests with aiida=2.5 will not run with `--cov` which speeds them up quite a lot.

before-notebook.d/00_untar_home.sh
danielhollas: I just wanted to guard against if somebody executed this script by mistake outside of the container.

danielhollas:
> Thanks @danielhollas. Just go through the PR, some minor comments, let's discuss tomorrow.

Thanks! It would be great if you at @superstar54 could give it a test, run the container and open the QeApp and try running some calculation to see if things are working, it's possible I overlooked something.

unkcpz:
> run the container and open the QeApp and try running some calculation to see if things are working, it's possible I overlooked something.

The PR to ghcr is not triggered, if you can bring it up, I think it is even better to test on the Azure k8s.

superstar54:
Hi @danielhollas , thanks for the great work. I have tested locally, and the QEApp is running fine.

One comment: in #740 , it only takes 6 s to launch, and this PR needs around ~ 12 s, any improvement can be done?

danielhollas:
I don't know, feel free to investigate. But I suspect that in #774 you're not doing some work that actually needs to get done, e.g. generating the SSH key and running database migration. Those two things easily add 3s.

danielhollas:
You can test how much time it takes if you run it for the second time (with the volume attached). If there is any time difference, that would suggest that an improvement is possible

danielhollas:
Closing in favour of #778

PR #774, opened 2024-07-18
Docker home build refactor - from origin 

danielhollas:
Same as #773 but from the origin repo to trigger container upload to ghcr.io

CC @unkcpz


danielhollas:
Uh oh

image

We really need to get rid of the artifact upload. I'll work on that tomorrow. @unkcpz @superstar54 please for now test locally.

unkcpz:
> We really need to get rid of the artifact upload

Okay, I think it is now doable, since you show in aiida-core that we can still test it. I'll check it and try again.

PR #776, opened 2024-07-18
Remove arm64 build

### LINKED ISSUE #775: Discussion: Should we build qeapp image for arm64?
The current docker CI machinery is very complex. If we could ditch the arm64 build, it would become much much simpler and easier to maintain. Given that the arm64 build is likely mainly for developers, I would propose to ditch it and only build for x86. Developers can always use the full-stack image. In fact I think it is beneficial if some devs work from the latest full-stack image so that it receives more dogfooding that way.

NOTE: The arm64 is currently failing, if you want to keep it, we need to switch it to buildjet runners.

WDYT @unkcpz @superstar54

danielhollas:
Closes #775 

- Removed arm64 build to reduce complexity
- moved Dockerfile to the root directory and removed `build.json` and `docker-build.hcl`. Image can now be build by `docker build . -t aiidalab/qe:newbuild`
- Only push image to ghcr.io, not to dockerhub. I think this is sufficient for this image.
- Removed the Firefox tests, both for CI speed and simplicity. As long as we've been running these UI tests in AWB and my app, we've never caught any bug that would be specific for Firefox. Given Firefoxe's low usage, I think we can skip it here and only test with Chrome. Since these tests do not run in parallel this saves a fair amount of time (>2min). And it also makes the workflow simpler. We'll still continue testing Firefox in AWB to make sure individual widgets work there.


danielhollas:
@unkcpz @superstar54 I need to re-enable the integration tests but otherwise this is good for initial review. Let's get this PR merged before the other. 

(-400 lines of YAML and configuration!)

danielhollas:
Tests are back and passing.

.github/workflows/docker-build-test-upload.yml
```diff
@@ -1,78 +1,93 @@
...
-                  runsOn: ${{ inputs.runsOn }}
+                  tags: ${{ steps.meta.outputs.tags }}
+                  load: ${{ github.event.pull_request.head.repo.fork }}
+                  push: ${{ ! github.event.pull_request.head.repo.fork }}
+                  context: .
+                  platforms: linux/amd64
+                  cache-to: |
+                      type=gha,scope=${{ github.workflow }},mode=min
+                  cache-from: |
+                      type=gha,scope=${{ github.workflow }}
```
unkcpz, line 69: Thanks for this!

.github/workflows/docker-build-test-upload.yml
```diff
@@ -1,78 +1,93 @@
...
+              if: ${{ !github.event.pull_request.head.repo.fork }}
+              id: meta
+              uses: docker/metadata-action@v4
+              with:
+                  images: |
+                      name=${{ env.IMAGE }}
+                  tags: |
+                      type=edge,enable={{is_default_branch}}
+                      type=ref,event=pr
+                      type=match,pattern=v(\d{2}\.\d{2}.\d+.*),group=1
```
unkcpz, line 54:
```suggestion
                      type=match,pattern=v(\d{4}\.\d{4}.*),group=1
```

I guess you copy this from aiida-core? :smile:

unkcpz (changes requested):
Thanks! @danielhollas Just one minor request, then good to go.

.github/workflows/docker-build-test-upload.yml
danielhollas: Nope, I actually took it from this repo. :-P I think it is correct, no? The current QeApp version is `24.10.0a1`

unkcpz: Ah, true, my bad, I confused with aiidalab-docker-stack :sweat_smile:

unkcpz (approved):
Approved, thanks! Let's wait tests.

PR #778, opened 2024-07-19
Tar home build

danielhollas:
Alternative to https://github.com/aiidalab/aiidalab-qe/pull/740

The main goal here is to reduce the complexity of status quo and of https://github.com/aiidalab/aiidalab-qe/pull/740.

The strategy of archiving home directory and untarring it at startup actually allows for a lot of simplification of the Dockerfile since everything can be directly prepared in home folder, without intermediary steps, and this allows to get rid of the current startup scripts (70_, 71_)
 
All startup scripts from full-stack are preserved and reused, which minimizes duplication, resolves the SSH key issue and should be more maintainable

The only new startup script is the 00_untar_home.sh which is basically the same here as in 

    https://github.com/aiidalab/aiidalab-qe/pull/740.
    I've also done some unrelated cleanup:
        Moved Dockerfile to the root dir, and got rid of build.json and docker-bake.hcl which were introducing unnecessary complexity. One can now build the image by simply running docker build . -t aiidalab/qe:newbuild . This also allowed me to use podman on my dev machine, which otherwise doesn't support the buildx backend.

I've done some quick benchmarking, at starting the container takes around 12s on my machine. The image takes around 6.5Gb. We could trade around 300Mb images size for extra 3s of startup time if we compressed the home.tar archive. (My timings seems roughly consistent with those observed in https://github.com/aiidalab/aiidalab-qe/pull/740.

Reducing the image size will be done in subsequent PR


unkcpz:
Deployed and alive on https://staging-demo.aiidalab.io/
Looks all good!

danielhollas:
I've opened #781 using a better strategy (multistage build).

PR #780, opened 2024-07-19
Fix integration tests for PRs from forks

danielhollas:


danielhollas:
@unkcpz I've merged this to unblock the other PRs, but please let me know if the approach I took here is unclear.

PR #781, opened 2024-07-22
Archive home directory using multi-stage build

danielhollas:
Supersedes #778, hopefully the last iteration!

The main goal here is to reduce the complexity of status quo and of https://github.com/aiidalab/aiidalab-qe/pull/740.

The strategy of archiving home directory and extracting it at startup allows for a bunch of simplification of the Dockerfile since everything can be directly prepared in home folder, without intermediary steps, and this allows to get rid of the current startup scripts (70_, 71_).
 
All startup scripts from full-stack are preserved and reused, which minimizes duplication, resolves the SSH key issue and should be more maintainable

The only new startup script is the 00_untar_home.sh which is basically the same here as in  https://github.com/aiidalab/aiidalab-qe/pull/740.
 
I've done some quick benchmarking, at starting the container takes around 12s on my machine. The image takes around 5.8Gb. We could trade around 300Mb images size for extra 3s of startup time if we compressed the home.tar archive. (My timings seems roughly consistent with those observed in https://github.com/aiidalab/aiidalab-qe/pull/740.

Reducing the image size will be done in subsequent PR.


src/aiidalab_qe/common/setup_codes.py
```diff
@@ -21,9 +20,15 @@
 
 QE_VERSION = "7.2"
 
-CONDA_ENV_PREFIX = Path.home().joinpath(
-    ".conda", "envs", f"quantum-espresso-{QE_VERSION}"
-)
+
+def get_qe_env():
```
danielhollas, line 24: I've created this function so that we don't have to link QE conda env from `/opt/conda` to `~/.conda`.

danielhollas:
@unkcpz @superstar54 This is hopefully the final iteration of the docker build. :sweat_smile: For easier review, I've split chunks of this PR into 3 extra pull requests that should be reviewed and merged first: #782, #783, #784

The startup takes slighly less than 10s. More speedup is I think possible:
- Updating to Python 3.11
- Speeding up `verdi storage migrate` (I'll open an issue on aiida-core)
- Speeding up verdi in general (not sure if possible)

The size is currently 5.8Gb, but I don't understand where the increase comes from, even though I tried very hard to get rid of it. Let's merge the three PRs first, I'll continue to investigate....

Dockerfile
```diff
@@ -1,48 +1,100 @@
...
-        qe=${QE_VERSION} \
-     && mamba clean --all -f -y && \
-     fix-permissions "${CONDA_DIR}" && \
-     fix-permissions "/home/${NB_USER}"
-
-# Download the QE pseudopotentials to the folder for afterware installation.
-ENV PSEUDO_FOLDER=${CONDA_DIR}/pseudo
+RUN --mount=from=uv,source=/uv,target=/bin/uv \
+    uv pip install --strict --system --cache-dir=${UV_CACHE_DIR} .
+
```
unkcpz, line 40: This also happened in stage 4 but I understand it is not avoidable.

Dockerfile
```diff
@@ -1,48 +1,100 @@
...
+
+# STAGE 3:
+# - Prepare AiiDA profile and localhost computer
+# - Install QE codes and pseudopotentials
+# - Archive home folder
```
unkcpz, line 45: Stage 2 and stage 3 better to be merged? It says "to run aiidalab_qe CLI commands" then clear to directly run it in the same stage. I believe the finale size will be the same.

unkcpz (changes requested):
Thanks @danielhollas! The implementation looks super clear, I have one minor request.

@superstar54, you showed interests to learn more docker stuff. I'd say this is a nice PR to read if you have time.

Dockerfile
danielhollas: Yes. It is unfortunate that we need to install all the dependencies to install qe codes and pseudos. But `uv` is so fast and I am re-using its cache that in terms of speed it doesn't matter much.

Dockerfile
danielhollas: Actually, it is beneficial to leave this as a separate stage, because the uv cache can then be immediately used in the final stage, without waiting for the rest of the home_stage build (which is the longest build part). I've rearranged things a bit for better cache utilization. Now, when you modify `Dockerfile` and rebuild, it only takes 10s!

Dockerfile
```diff
@@ -4,50 +4,100 @@ ARG UV_VER=0.2.27
...
+USER ${NB_USER}
+
+WORKDIR /tmp
+# Install python dependencies
+# Use uv cache from the previous build step
+ENV UV_CONSTRAINT=${PIP_CONSTRAINT}
+RUN --mount=from=uv,source=/uv,target=/bin/uv \
+    --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
+    --mount=from=build_deps,source=${QE_APP_SRC},target=${QE_APP_SRC},rw \
+    uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} ${QE_APP_SRC}
```
danielhollas, line 83: Note that I have added the `--compile-bytecode` parameter. Unlike pip, uv does not compile python files (*.py) into bytecode (*pyc) after installation. 

This increases the image size by ~110Mb, but results in slightly fast App startup when user first launched QeApp.

unkcpz:
> the image size dropped from 5.8Gb to 5.1Gb (compared to 4.1Gb on main), although I have no idea why.

If you click on a specific tag in dockerhub, you'll see how much size each layer generated: https://hub.docker.com/r/aiidalab/qe/tags
Apparently, you remove upload images to dockerhub ;)

unkcpz:
Forget to mention, the image was redeployed to the Azure and works as I expected. I think we can merge this and from next week I 'll working on hyperqueue integration.

unkcpz (approved):

.github/workflows/docker-build-test-upload.yml
```diff
@@ -80,7 +80,18 @@ jobs:
             - name: Run pytest for Chrome
               run: pytest -sv --driver Chrome tests_integration/
               env:
-                  QE_IMAGE: ${{ env.IMAGE }}@${{ steps.build-upload.outputs.digest }}
+                  # We'd like to identify the image by its unique digest, i.e ghcr.io/aiidalab/qe@sha256:<digest>
+                  # but that sadly does not work when the image is loaded to Docker locally and not published on ghcr.io
+                  # as is the case for PRs from forks. Hence this super-ugly ternary expression...
+                  # For forks, we take the image as ghcr.io/aiidalab/qe:pr-XXX
+                  # which is stored in the steps.meta.outputs.tags variable
```
unkcpz, line 87: Are you sure this will work for the PR from fork? I think from fork there is no permission to upload anything to the GH package.

unkcpz: Ah, it is in this PR https://github.com/aiidalab/aiidalab-qe/pull/780. There is no pr-780 in GH registry, so I guess it is not working? If so can you revert this change back for this PR?

danielhollas: Yes, you can revert it here and merge. I wanted to check that the fix works for PRs from origin as well.

(we indeed don't upload for PRs from fork, we just load to Docker locally, see the build step above how this is achieved)

danielhollas: To clarify, the fix from #780 is working and #780 can be merged as well. :-)

danielhollas: I've reverted the commit here, the fixed should be merged in #780.

unkcpz: Thanks!

unkcpz:
> Apparently, you remove upload images to dockerhub ;)

Hi @danielhollas, I guess you miss one comment above?

danielhollas:
Hi. I am aware, although I should have been more explicit. Does it bring any benefits of publishing to Dockerhub? Given what you told me about this image being most important for demo server deployment, I think that only publishing to ghcr.io is fine?

Publishing to Dockerhub would complicate the GitHub actions workflow, so unless there is a clear benefit I'd advise against it.

unkcpz:
One thing a bit annoy is I can not find what is available tags in ghcr.io registry, since we have a lot images with pr-xx tags and disgest directly.

danielhollas:
Yeah, ghcr.io interface is not great. But with the significantly simpler workflow, you don't really need to search for tags, do you? If you look at the workflow, we don't push by digest or commit sha anymore, only pr-xxx on PRs and edge on main, and version when a new tag is pushed.

unkcpz:
Make sense, I think we want to have highly maintainable repo that involve less outside tools as possible to fit the goal. Would you then mind to add a paragraph to README to tell which tag can be used and is from which branch? Sort of like "Supported tags" section of `aiidalab-docker-stack`.

PR #782, opened 2024-07-22
Print container logs if service fails to start

danielhollas:
When a Docker container fails to start properly, the integration tests fail with a timeout error. In that case, we will print container logs for debugging. We use the same strategy in AWB and it works well.

Extracted from #781.


unkcpz (approved):
Looks all good!

PR #783, opened 2024-07-22
Move QE conda install into separate stage

danielhollas:
When building the QEApp Docker container, the Quantum Espresso installation into a conda environment is a step that's completely independent from the others. By putting it into a separate Docker build stage, it can be build in parallel (when using buildkit), as:

    DOCKER_BUILDKIT=1  docker build .

Also added a `.dockerignore` file. This is important since we're copying the whole directory into the container and we don't want to copy junk inside. Even if we later run `git clean`, once the files are copied, that stay in the fs layers.

Extracted from #781

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,14 +39,14 @@ RUN  --mount=from=uv,source=/uv,target=/bin/uv \
      uv pip install --system --no-cache . && \
      fix-permissions "${CONDA_DIR}"
 
-COPY --from=qe_conda_env "${QE_DIR}" "${QE_DIR}"
-# TODO: Remove this once we get rid of 70_prepare-qe-executable.sh
-ENV QE_VERSION="$QE_VER"
-
 # Download the QE pseudopotentials to the folder for afterware installation.
 ENV PSEUDO_FOLDER=${CONDA_DIR}/pseudo
 RUN mkdir -p ${PSEUDO_FOLDER} && \
     python -m aiidalab_qe download-pseudos --dest ${PSEUDO_FOLDER}
+
+COPY --from=qe_conda_env "${QE_DIR}" "${QE_DIR}"
+# TODO: Remove this once we get rid of 70_prepare-qe-executable.sh
+ENV QE_VERSION="$QE_VER"
 
 COPY before-notebook.d/* /usr/local/bin/before-notebook.d/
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,7 +1,26 @@
...
+ARG QE_VER=7.2
+ARG QE_DIR=/opt/conda/envs/quantum-espresso-${QE_VER}
 
+FROM ghcr.io/astral-sh/uv:0.2.27 AS uv
+
+# STAGE 1
+# Install QE into conda environment in /opt/conda
+# This step is independent from the others and can be run in parallel.
+FROM ghcr.io/aiidalab/full-stack:${FULL_STACK_VER} AS qe_conda_env
+ARG QE_VER
```
danielhollas, line 13: This is needed so that the variables are passed from the top level.

Dockerfile
```diff
@@ -17,26 +36,12 @@ RUN  --mount=from=uv,source=/uv,target=/bin/uv \
...
-     # (for `aiidalab/aiidalab<23.03.2` the `setup.py` should be in the root folder of the app https://github.com/aiidalab/aiidalab/pull/382).
-     # and then the app and restart the daemon in the end.
-     # But since the aiida profile not yet exists, the daemon restart will fail but it is not a problem.
-     # Because we only need the dependencies to be installed.
-     # aiidalab install --yes --python ${CONDA_DIR}/bin/python  "quantum-espresso@file://${PREINSTALL_APP_FOLDER}" && \
-     # However, have to use `pip install` explicitly because `aiidalab install` call `pip install --user` which will install the app to `/home/${NB_USER}/.local`.
-     # It won't cause issue for docker but for k8s deployment the home folder is not bind mounted to the pod and the dependencies won't be found. (see issue in `jupyter/docker-stacks` https://github.com/jupyter/docker-stacks/issues/815)
      uv pip install --system --no-cache . && \
-     fix-permissions "${CONDA_DIR}" && \
-     fix-permissions "/home/${NB_USER}"
```
danielhollas, line 32: This is not needed since the `uv pip install --system` does not touch the home directory.

Dockerfile
unkcpz: Cool, learned. I was always defined in the stage and don't know how to move all ARGs to the front of file.

unkcpz (approved):
Thanks!

Dockerfile
danielhollas: Yeah, it's definitely not obvious, took me a while to figure out.

PR #784, opened 2024-07-22
Support QE conda env in /opt/conda

danielhollas:
Change `setup_codes.py` to support QE conda environment located in `/opt/conda`

Built on top of #783. Separated from #781.


src/aiidalab_qe/common/setup_codes.py
```diff
@@ -134,7 +139,7 @@ def setup_codes():
     try:
         run(["python", "-c", python_code], capture_output=True, check=True)
     except CalledProcessError as error:
-        raise RuntimeError(f"Failed to setup codes: {error}")
+        raise RuntimeError(f"Failed to setup codes: {error}") from None
```
danielhollas, line 142: This is just to make stack traces nicer. Here's the correspoding flake8-bugbear rule as explained by ruff

```
❯ ruff rule B904
# raise-without-from-inside-except (B904)

Derived from the **flake8-bugbear** linter.

## What it does
Checks for `raise` statements in exception handlers that lack a `from`
clause.

## Why is this bad?
In Python, `raise` can be used with or without an exception from which the
current exception is derived. This is known as exception chaining. When
printing the stack trace, chained exceptions are displayed in such a way
so as make it easier to trace the exception back to its root cause.

When raising an exception from within an `except` clause, always include a
`from` clause to facilitate exception chaining. If the exception is not
chained, it will be difficult to trace the exception back to its root cause.

## Example
```python
try:
    ...
except FileNotFoundError:
    if ...:
        raise RuntimeError("...")
    else:
        raise UserWarning("...")
```

Use instead:
```python
try:
    ...
except FileNotFoundError as exc:
    if ...:
        raise RuntimeError("...") from None
    else:
        raise UserWarning("...") from exc
```

## References
- [Python documentation: `raise` statement](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement)

```

Comment (unkcpz) [reply src/aiidalab_qe/common/setup_codes.py:142]: Just curious, how you find these good thing? Is it from your LSP?

File: src/aiidalab_qe/common/setup_codes.py [hunk 2]
```diff
@@ -21,9 +20,15 @@
...
-    ".conda", "envs", f"quantum-espresso-{QE_VERSION}"
-)
+
+def get_qe_env():
+    # QE is already pre-installed in the QE image
+    path = Path(f"/opt/conda/envs/quantum-espresso-{QE_VERSION}")
+    if path.exists():
+        return path
+    else:
+        return Path.home().joinpath(".conda", "envs", f"quantum-espresso-{QE_VERSION}")
```
unkcpz, line 30: This can cause an edge problem that when the container is started with qeapp image and then shifted to using full-stack image. The QE not able to be accessed. 
We were trying to provide `full-stack`, `qe-app`, `base-with-server` images in demo server. But as suggested by @giovannipizzi, we agree on only keep `qe-app` image. This change then wont be a problem.

unkcpz (approved):
A small note from my side. All good, thanks @danielhollas

src/aiidalab_qe/common/setup_codes.py
danielhollas: This is from ruff, which I have setup as LSP in neovim.

While the pre-commit in this repo uses ruff, it doesn't have any ruff-specific configuration. I think we should copy over the configuration from AWB. I think we have a good experience with that set and it will make things consistent across the two repos.

src/aiidalab_qe/common/setup_codes.py
unkcpz: >  in this case the current solution with the symlink would not work anyway since the symlink would be broken, right?

I think it won't work.

PR #785, opened 2024-07-24
Adopt more ruff rules to align with AWB codebase

danielhollas:
I've copied over the ruff configuration from AWB which served us well so far. Most of the necessary fixes were done automatically, I've done around 20 of them manually.

TODO: fix violations in notebooks and use ruff to lint notebooks directly.


pyproject.toml
```diff
@@ -4,3 +4,32 @@ requires = [
...
 build-backend = "setuptools.build_meta"
+
+[tool.ruff]
+line-length = 88
+show-fixes = true
+output-format = "full"
+target-version = "py39"
+
+[tool.ruff.lint]
+ignore = ["E501", "E402", "TRY003", "RUF012", "N806"]
```
AndresOrtegaGuerrero, line 15: Should we add docstring about this one to keep in mind when to use them ?

src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -111,7 +110,7 @@ def __init__(self, **kwargs):
         )
 
     @tl.observe("previous_step_state")
-    def _observe_previous_step_state(self, change):
+    def _observe_previous_step_state(self, _change):
```
AndresOrtegaGuerrero, line 113: I dont know if this one is necessary and we just keep it as change

src/aiidalab_qe/common/process.py
```diff
@@ -87,9 +87,9 @@ def __init__(self, process_label, **kwargs):
         self.refresh_work_chains()
         # the following is needed to disable the button.
 
-    def parse_extra_info(self, pk: int) -> dict:
+    def parse_extra_info(self, _pk: int) -> dict:
```
AndresOrtegaGuerrero, line 90: I think we should keep pk and not use _

src/aiidalab_qe/common/setup_pseudos.py
```diff
@@ -287,7 +286,7 @@ def _default_error(self):
     @traitlets.observe("busy")
     @traitlets.observe("error")
     @traitlets.observe("installed")
-    def _update(self, change):
+    def _update(self, _change):
```
AndresOrtegaGuerrero, line 289: same here

src/aiidalab_qe/common/setup_codes.py
```diff
@@ -130,6 +129,8 @@ def _generate_string_to_setup_code(code_name, computer_name="localhost"):
             prepend_text,
         )
         return python_code
+    else:
+        return ""
```
AndresOrtegaGuerrero, line 133: Should we add docstring here, to explain the aiida code node exist?

tests/conftest.py
```diff
@@ -107,10 +106,6 @@ def _generate_xy_data(xvals=None, yvals=None, xlabel=None, ylabel=None):
         """
         from aiida.orm import XyData
 
-        xvals = xvals
-        yvals = yvals
```
AndresOrtegaGuerrero, line 111: I guess these lines werent used?

pyproject.toml
danielhollas: I am not sure what you mean by this comment? These are rules that we do not enforce, for various reasons, mostly because they would be too tedious to enforce. You can learn by each of them by running

```console
ruff rule <rule_code>
```

src/aiidalab_qe/app/configuration/__init__.py
danielhollas: This one comes from `ARG002 Unused method argument:`

```
# unused-method-argument (ARG002)

Derived from the **flake8-unused-arguments** linter.

## What it does
Checks for the presence of unused arguments in instance method definitions.

## Why is this bad?
An argument that is defined but not used is likely a mistake, and should
be removed to avoid confusion.
```

We definitely want to keep this rule enabled since it can catch bugs (it already caught some minor bugs in this PR). 

In cases where unused argument cannot be removed, the convention is to prefix the variable with underscore (or just use a single underscore instead of full name). Alternatively, we can add `# noqa` comment on such lines. I am slightly leaning towards using the underscore, since it immediately let's you know that this argument is not used anywhere in the function. (and it is what is used in AWB).

src/aiidalab_qe/common/setup_codes.py
danielhollas: I added a short comment.

danielhollas:
@AndresOrtegaGuerrero thank you for the review, please see my replies. 

I've changed the configuration so that ruff now checks notebooks directly as well, and removed the `nbqa` pre-commit hook which is now not needed.

src/aiidalab_qe/app/configuration/__init__.py
AndresOrtegaGuerrero: ok i see, i see your point of having a syntax for not being used , then i guess is ok that in some cases we will have _change and in others change.

AndresOrtegaGuerrero (approved):
Thank you for your work @danielhollas LGTM! , I was wondering if @superstar54 or @PNOGillespie can check the XPS part in particular the test where there are lines not used.

tests/conftest.py
superstar54: > I am assuming it is a leftover after some refactoring?

Yes. Thanks for removing them.

superstar54 (approved):
Thanks for the work. LGTM!

PR #786, opened 2024-07-25
XAS/XPS: Fix Bug Caused by VdW Module

### LINKED ISSUE #765: Bug for Core-Hole SCF Calculations Combined with DFT-D3 Correction
**Describe the bug**
(From AiiDA-QE) When running core-hole SCF calculations in the `XpsWorkChain` and `XspectraCrystalWorkChain` with default settings in combination with the Grimme D3 vdW correction, the SCF calculation returns `NaN` for the total energy. This is because the DFT-D3 module of QE sets D3 parameters based on the `ATOMIC SPECIES` name, but if the name is not a valid element symbol then the program continues to the SCF procedure rather than crashing due to the unidentified element symbol.

This will happen using the default settings for both `WorkChain`s because the default `abs_atom_marker` is set to `X`:
SCF_Input_File-Default_Absorber
D3_Terms_Table-Missing_Absorber
D3_Terms_Total-Missing_Absorber
D3_Terms_Total-Absorber
SCF-Final_Energy_NaN

**Proposed Solution**
Changing the absorbing atom to `<element_symbol><number>` easily fixes the issue, as numbers are automatically trimmed when the DFT-D3 module reads the species name:
SCF_Input_File-C1_Absorber
D3_Terms_Table-C1_Absorber

For `AiiDA-QE`, this is possible through overrides, however there is currently no way to set the `abs_atom_marker` in `AiiDALab-QE` and thus the `WorkChain`s will fail unless the absorbing atom marker is changed.

PNOGillespie:
Resolves #765

Forces core-hole SCF calculation steps in the XPS and XAS plugins to not use VdW corrections of any type, keeping any such setting for the relax step.

Since the VdW correction only affects the total energy/forces and doesn't affect the electronic structure, simply removing the term from the core-hole SCF calculation solves the issue without any downsides.


superstar54 (approved):
LGTM!

PR #787, opened 2024-07-31
adding ProjwfcBandWorkchain

AndresOrtegaGuerrero:
This PR includes the ProjwfcBandWorkChain from aiida-wannier90-workflows to add fat band calculation option in the Bands plugin


AndresOrtegaGuerrero:
Pending issues to continue implementation merging this PR 
https://github.com/aiidateam/aiida-wannier90-workflows/pull/56#issuecomment-2210383837

src/aiidalab_qe/plugins/bands/__init__.py
```diff
@@ -15,8 +16,15 @@ class BandsOutline(OutlinePanel):
...
 
+projwfc_code = QEAppComputationalResourcesWidget(
+    description="projwfc.x",
+    default_calc_job_plugin="quantumespresso.projwfc",
+)
+
+
 bands = {
     "outline": BandsOutline,
+    "code": {"projwfc": projwfc_code},
```
superstar54, line 27: Hi @AndresOrtegaGuerrero ,  this should fix the failed tests. Because there is already a code named `projwfc` in pdos plugin, so you need use another name. 
```suggestion
    "code": {"projwfc_bands": projwfc_code},
```
I will open another PR to handle this automatically.

AndresOrtegaGuerrero: Thank you!

AndresOrtegaGuerrero:
There is a new release of aiida-wannier90-workflows , so we can evaluate this PR for merging it with aiidalab-qe

setup.cfg
```diff
@@ -60,6 +61,9 @@ aiidalab_qe.properties =
     electronic_structure = aiidalab_qe.plugins.electronic_structure:electronic_structure
     xas = aiidalab_qe.plugins.xas:xas
 
+aiida.workflows =
+    aiidalab_qe.bands_workchain = aiidalab_qe.plugins.bands.bands_workchain:BandsWorkChain
+
```
edan-bainglass, line 66: We already had bands calculation via the built-in bands plugin. I'm surprised there wasn't already an associated workflow.

src/aiidalab_qe/plugins/bands/bands_workchain.py
```diff
@@ -0,0 +1,382 @@
+import math
```
edan-bainglass, line 1: Conceptually, what's the difference between workchain.py and bands_workchain.py? The latter breaks from the plugin template a bit, no?

src/aiidalab_qe/plugins/bands/result.py
```diff
@@ -16,7 +16,10 @@ def __init__(self, node=None, **kwargs):
     def _update_view(self):
         # Check if the workchain has the outputs
         try:
-            bands_node = self.node.outputs.bands
+            if "bands" in self.node.outputs.bands:
+                bands_node = self.node.outputs.bands.bands
+            elif "bands_projwfc" in self.node.outputs.bands:
+                bands_node = self.node.outputs.bands.bands_projwfc
```
edan-bainglass, line 22: So it's regular of fat? Am I understanding it correctly? Not possible to combine the mechanics? I guess it's a different code, so no.

src/aiidalab_qe/plugins/bands/setting.py
```diff
@@ -1,13 +1,16 @@
...
+import traitlets as tl
 
+from aiida import orm
 from aiidalab_qe.common.panel import Panel
 
 
 class Setting(Panel):
     title = "Bands Structure"
     identifier = "bands"
+    input_structure = tl.Instance(orm.StructureData, allow_none=True)
```
edan-bainglass, line 13: You using it anywhere?

edan-bainglass (approved):
Generally looks okay. See my comments, but no changes required, just clarification.

setup.cfg
AndresOrtegaGuerrero: No , because , we were calling a workflow from aiida-quantumespresso plugin, therefore that workchain is already present. Here I did a new one to be able to switch between the from aiida-quantumespresso and the one from aiida-wannier90-workflows (capable of the fat bands)

src/aiidalab_qe/plugins/bands/bands_workchain.py
AndresOrtegaGuerrero: yeah, the workchain.py, is that one used within the "architecture" of the plugin system within aiidalab-qe. While this bands_workchain is the actual aiida workflow, that workchain.py (is more like a controller that calls the workflow builder) passes the builder for the app.

src/aiidalab_qe/plugins/bands/result.py
AndresOrtegaGuerrero: yeah, the featuer somehow are in different aiida plugins, the "regular" is from aiida-quantumespresso, while the "fat bands" from the wannier90-workflows

src/aiidalab_qe/plugins/bands/setting.py
AndresOrtegaGuerrero: Actually, No, with this PR, that trailets should be removed, since now the logic of kpath_2d is automatic , so there is no need , i will double check it is indeed not used

src/aiidalab_qe/plugins/bands/bands_workchain.py
edan-bainglass: Why not migrate this new workflow to `aiida-quantumespresso`?

edan-bainglass: Never mind, I should have read the comment above first 😅 So it is because it combines workflows from two plugins, correct?

src/aiidalab_qe/plugins/bands/setting.py
AndresOrtegaGuerrero: Removed

src/aiidalab_qe/plugins/bands/setting.py
```diff
@@ -18,39 +18,33 @@ def __init__(self, **kwargs):
...
-            options=[
-                ("Hexagonal", "hexagonal"),
-                ("Square", "square"),
-                ("Rectangular", "rectangular"),
-                ("Centered Rectangular", "centered_rectangular"),
-                ("Oblique", "oblique"),
-            ],
-            value="hexagonal",
+        self.projwfc_bands = ipw.Checkbox(
+            description="Fat bands calculation",
```
mikibonacci, line 29: Hi @AndresOrtegaGuerrero, could you add some more description here? maybe is not clear what "fat bands" mean, for a non expert user (I am thinking about experimentalists). Thanks!

src/aiidalab_qe/plugins/electronic_structure/result.py
```diff
@@ -20,7 +20,10 @@ def _update_view(self):
             pdos_node = None
 
         try:
-            bands_node = self.node.outputs.bands
+            if "bands" in self.node.outputs.bands:
```
mikibonacci, line 23: Hi @AndresOrtegaGuerrero, I don't understand why we need to put this block of code, as it it already in plugins/bands/result.py

mikibonacci (commented):
Thanks for the PR @AndresOrtegaGuerrero, I added few comments on something that is unclear to me

src/aiidalab_qe/plugins/electronic_structure/result.py
AndresOrtegaGuerrero: This is because now we have a BandWorkChain, that selects between the one from aiida-quantumespresso or the one from aiida-wannier90-workflows (for doing fat bands)

src/aiidalab_qe/plugins/bands/setting.py
AndresOrtegaGuerrero: I added a description let me know what you think

mikibonacci: Hi @AndresOrtegaGuerrero, for me it is good! thanks!

src/aiidalab_qe/plugins/electronic_structure/result.py
AndresOrtegaGuerrero: I will modify this code , because , if we merge it like this , it wont be compatible with previous versions that contain Bands calculations

AndresOrtegaGuerrero:
@mikibonacci @edan-bainglass , I modify the result.py part from the bands plugin. With this change there is backwards compatibility of bands calculations using previous versions of the app.

edan-bainglass (approved):
LGTM! Thanks

mikibonacci (approved):
LGTM, thanks!

Issue #789, opened 2024-07-31
Slim size of image


unkcpz:
I did a layer by layer investigation using docker slim to see how aiidalab-qe image grows, find an extra layer added with same content by https://github.com/aiidalab/aiidalab-qe/blob/94b92f3cf66a572881680711f5380b582e7fec75/Dockerfile#L89 
It is identical in terms of files as the layer with `uv pip install ${QE_APP_SRC}`. 
By remove the `fix-permission` line, the image size goes down for more than 600 MB.
Since `fix-permission` is mostly for changing permission so user can access to the conda env, maybe it is better to put QE in home directly? @danielhollas 

But the current size fit for our purpose to deploy on to demo server, this is thus not urgent.

danielhollas:
Thank you for the investigation! That's crazy that changing permissions has such a huge effect on size.

Because we run most of the commands in the image under NB_USER, I suspect we don't even have to run the `fix_permissions` script. @unkcpz can you open the the PR where this line is removed? If the image passes the tests and works then let's go with it.

(Perhaps we should then do the same thing in the aiidalab-docker-stack images. We call the `fix_permissions` script a lot, but I suspect we actually don't need to.

PR #790, opened 2024-07-31
Slim image without fix-permissions on CONDA_DIR

unkcpz:
see #789


danielhollas (approved):
If this works let's go for it! In particular, could you test that you can still run `conda install` and `pip install --no-user` into the base environment?

I've opened https://github.com/aiidalab/aiidalab-docker-stack/issues/486 for follow-up on aiidalab-docker-stack. Perhaps there are wins to be had there as well!

unkcpz:
> could you test that you can still run conda install and pip install --no-user into the base environment?

Yes, I can run them without problem.

PR #792, opened 2024-07-31
No untar

### LINKED ISSUE #791: ~0.5 s overhead with tar
I looked into container runtime a bit and learned that it actually doing `tar` for every layer. It means the tar we did for home did not help with saving size.
Meanwhile untar the `home.tar` cost ~0.5s in my nice workstation. 
I think we can simply cp home to `/opt/store/home` (try to add in #790 as the folder in the `/opt` with correct permission) and using `mv` which takes basically no time since I assume it just move the pointer to file and update metadata of location.

unkcpz:
fixes #791 

- Not using tar/untar which is already take care by container runtime.
- Move the home leave thus leave no duplicated data in `/opt/` which occupied disk space.


unkcpz:
close because it copy data anyway since home is in volume. I did wrong test by using container without volume.

PR #795, opened 2024-08-12
Hq as light scheduler

unkcpz:
- [x] naming?? `local-hq` or replace `localhost`? (use `localhost-hq` and hide `localhost` for old calculation backward compatibility)
- [x] left todos and XXX annotations.
- [x] Rebase after code/pseudo setup refactoring. 
- [x] Merge https://github.com/aiidateam/aiida-hyperqueue/pull/33 and release v0.2


src/aiidalab_qe/common/setup_codes.py
```diff
@@ -171,40 +222,26 @@ def install(force=False):
...
+                    yield f"Preparing setup script for ({code_name}) on ({computer})..."
+                    code_string = _generate_string_to_setup_code(code_name, computer)
                     python_code += code_string
             try:
                 yield "Setting up all codes..."
-                run(["python", "-c", python_code], capture_output=True, check=True)
-            except CalledProcessError as error:
+                subprocess.run(
+                    ["python", "-c", python_code], capture_output=True, check=True
+                )
```
unkcpz, line 237: I think https://github.com/aiidalab/aiidalab-qe/pull/706 bring the footgun, increase the difficulty of maintenance. There are no clear error message when the setup script is fail. I'd argue to revert that change back.
I forget to pass the `computer` argument and the script simply failed with an opaque error message.

superstar54: What's the error message? Reverting to the old code really helps debuging the error? In the old code, 
```python
def setup_codes():
    for code_name in CODE_NAMES:
        _setup_code(code_name)
```
it didn't pass the computer too, thus it will probably show the same `opaque` error message using the `verdi` command.

#706 significantly enhances the speed of code setup, which remains beneficial, especially when users install the app from the AppStore, thus, I wouldn't suggest reverting to the old code, instead, we can improve it using `InstalledCode` in the future when we fix the thread problem.
Once we have more information from the error messages and can compare them, we can discuss further.

unkcpz: If the `verdi` command failed it shows which command I run and I just need to rerun it to see what's wrong. With scripts, nothing I can do, it even not showing the content of the script. Can you check if https://github.com/aiidalab/aiidalab-qe/pull/695 implementation can be done?

Dockerfile
```diff
@@ -82,9 +101,27 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${QE_APP_SRC},target=${QE_APP_SRC},rw \
     uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} ${QE_APP_SRC}
 
+# TODO: this seems need to do twice
```
unkcpz, line 104: I don't know how to simplify this, pip install in stage 2 above is needed to have the scheduler so the `verdi computer setup` can run. But if I understand correctly, in stage 2 it installed to /opt/conda and will not appeared in stage 3, so I need to do it again in stage 3. Is that correct? @danielhollas

Dockerfile
```diff
@@ -43,19 +44,37 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
...
 RUN mkdir -p ${PSEUDO_FOLDER} && \
     python -m aiidalab_qe download-pseudos --dest ${PSEUDO_FOLDER}
 
+ENV UV_CONSTRAINT=${PIP_CONSTRAINT}
+# Install the aiida-hyperqueue
+# XXX: fix me after release aiida-hyperqueue
+RUN --mount=from=uv,source=/uv,target=/bin/uv \
+    --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
+     git clone https://github.com/aiidateam/aiida-hyperqueue && \
+     uv pip install --system --strict --compile-bytecode --cache-dir=${UV_CACHE_DIR} aiida-hyperqueue
```
danielhollas, line 68: You don't need to manually clone, you can specify a git dependency to pip directly

```suggestion
     uv pip install --system --strict --compile-bytecode --cache-dir=${UV_CACHE_DIR} "aiida-hyperqueue @ git+https://github.com/aiidateam/aiida-hyperqueue"
```

Now you don't need to have that as a separate step and you can just add the extra dependency to uv invocation on line 43 in build_deps stage.

Dockerfile
```diff
@@ -82,9 +101,27 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
...
     uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} ${QE_APP_SRC}
 
+# TODO: this seems need to do twice
+# ENV UV_CONSTRAINT=${PIP_CONSTRAINT}
+# # Install the aiida-hyperqueue
+# # XXX: fix me after release aiida-hyperqueue
+# RUN --mount=from=uv,source=/uv,target=/bin/uv \
+#     --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
+#      git clone https://github.com/aiidateam/aiida-hyperqueue && \
+#      uv pip install --system --strict --compile-bytecode --cache-dir=${UV_CACHE_DIR} aiida-hyperqueue
```
danielhollas, line 111:
```suggestion
    uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} \
     ${QE_APP_SRC} "aiida-hyperqueue @ git+https://github.com/aiidateam/aiida-hyperqueue" 
```

I think this should do the trick

Dockerfile
danielhollas: I think you're correct. But you can significantly simplify this by combining with the previous RUN step. See my other comments.

src/aiidalab_qe/common/setup_codes.py
danielhollas: I've had in my head an approach for refactoring this logic for some time, I've opened #796 as a first step in that direction.

Dockerfile
unkcpz: I also did a change in this PR, error from running script can be output by printing `err.stderr`.

unkcpz:
Just find a cache when pre-setting the computer, the issue exist also for the current `localhost`. `LOCALHOST_MPI_PROCS_PER_MACHINE` in `40_prepare-aiida.sh` was triggered in the build phase in QeApp, so the number of procs is the machine that build the image, for the image that released therefore it is `2` the number of CPUs of github action runner I assume. 

In this PR, the same problem faced for the hq computer setup, therefore, I skip the setup for the default N_PROCS and MEM.

before-notebook.d/41_setup-hq-computer.sh
```diff
@@ -0,0 +1,32 @@
...
+# XXX: need to make daemon start late in docker stack
+# https://github.com/aiidalab/aiidalab-docker-stack/pull/498
+verdi daemon stop || echo "stop fail"
+
+computer_list=$(verdi computer list)
+if echo ${computer_list} | grep -q ${HQ_COMPUTER}; then
+    echo "${HQ_COMPUTER} already setup"
+else
+    # computer
+    verdi computer show ${HQ_COMPUTER} || verdi computer setup        \
```
danielhollas, line 14: I don't think you need this branch at all. You can just use the `verdi computer show` command to determine if the computer exists or not.

```suggestion
verdi computer show ${HQ_COMPUTER} || verdi computer setup        \
```

before-notebook.d/42_start-hq.sh
```diff
@@ -0,0 +1,59 @@
...
+# TODO: this will be run every time the container start, we need a lock file to prevent it.
+job_poll_interval="2.0"
+computer_name=${HQ_COMPUTER}
+python -c "
+from aiida import load_profile; from aiida.orm import load_computer;
+load_profile();
+load_computer('${computer_name}').set_minimum_job_poll_interval(${job_poll_interval})
+load_computer('${computer_name}').set_default_mpiprocs_per_machine(${CPU_LIMIT})
+load_computer('${computer_name}').set_default_memory_per_machine(${MEMORY_LIMIT})
+"
```
danielhollas, line 59: I would recommend running all this setup in the `41_setup-hq-computer.sh` script right after you setup the computer. That way you can ensure that you run it only once (I believe that's what we do in the base image)

src/aiidalab_qe/setup/codes.py
```diff
@@ -43,11 +44,19 @@ def get_qe_env():
...
 def qe_installed():
-    return get_qe_env().exists()
+    env_exist = get_qe_env().exists()
+    proc = subprocess.run(
+        ["conda", "list", "-n", f"{get_qe_env().name}", "qe"],
+        check=True,
+        capture_output=True,
+    )
+
+    # XXX: "qe" in check is not future proof if there are similar packages such as qe-tool, better solution?? JSON output??
```
danielhollas, line 54: I think you can take a look how we do it in aiidalab-docker-stack tests, where I think we use the json output.

Dockerfile
```diff
@@ -82,11 +105,30 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${QE_APP_SRC},target=${QE_APP_SRC},rw \
     uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} ${QE_APP_SRC}
```
danielhollas, line 106: I think you can just add it here to avoid invoking `uv` twice.
```suggestion
    uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} ${QE_APP_SRC} "aiida-hyperqueue@git+https://github.com/aiidateam/aiida-hyperqueue"
```

Dockerfile
```diff
@@ -43,19 +44,41 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
...
+RUN --mount=from=uv,source=/uv,target=/bin/uv \
+    --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
+     uv pip install --system --strict --compile-bytecode --cache-dir=${UV_CACHE_DIR} \
+     "aiida-hyperqueue@git+https://github.com/aiidateam/aiida-hyperqueue"
+
+COPY ./before-notebook.d/* /usr/local/bin/before-notebook.d/
+
+ENV HQ_COMPUTER="localhost-hq"
+# export
+ARG HQ_COMPUTER
```
danielhollas, line 74: Does this line do anything? It looks like you're defining it again on line 130 anyway.

Dockerfile
```diff
@@ -43,19 +44,41 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
...
 ENV PSEUDO_FOLDER=/tmp/pseudo
 RUN mkdir -p ${PSEUDO_FOLDER} && \
     python -m aiidalab_qe download-pseudos --dest ${PSEUDO_FOLDER}
 
+ENV UV_CONSTRAINT=${PIP_CONSTRAINT}
+# Install the aiida-hyperqueue
+# XXX: fix me after release aiida-hyperqueue
+RUN --mount=from=uv,source=/uv,target=/bin/uv \
+    --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
+     uv pip install --system --strict --compile-bytecode --cache-dir=${UV_CACHE_DIR} \
```
danielhollas, line 67: I don't think you need to compile the bytecode here since this is just a preparatory layer. This should save some time. I would also move this up and merge this with the previous uv invocation as part of `build_deps`. That way `uv` can resolve both packages together and ensure the dependencies are consistent.

```suggestion
     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} \
```

danielhollas (changes requested):
Some quick comments, overall looks reasonable, I'll do a second pass once all the todo's are done.

before-notebook.d/41_setup-hq-computer.sh
unkcpz: worth to try, I'll change it.

before-notebook.d/42_start-hq.sh
unkcpz: The problem is the the `CPU_LIMIT` and `MEMORY_LIMIT` is the runtime variables that didn't know during the building stage.

Dockerfile
unkcpz: True! I was going to defined it once and used it for the other stage but it seems not pass to next stage as `ENV`.
Now I hardcode the computer name in two place which is fine but not good enough.

unkcpz:
@danielhollas all requests addressed.

Dockerfile
```diff
@@ -1,11 +1,14 @@
 # syntax=docker/dockerfile:1
-ARG FULL_STACK_VER=2024.1022
+# TODO: change full-stack version to latest one when released with bc and late dameon start
+ARG FULL_STACK_VER=edge
```
unkcpz, line 3:
```suggestion
ARG FULL_STACK_VER=2024.1023
```

unkcpz:
I convert back to draft, since test on the production deployment reveal a problem that untar not happened for the presistent volume since it checks home is empty but there will be a `lost+found` folder in k8s deployment. I change to use `.FLAG_HOME_INITIALIZED` flag file. If it works, I'll make it a separated PR.

unkcpz:
Ready for another review.

before-notebook.d/00_untar-home.sh
```diff
@@ -15,12 +15,20 @@ if [[ $(ls -A ${home} | wc -l) = "0" ]]; then
...
   echo "$home folder is not empty!"
   ls -lrta "$home"
 fi
+
+if [ -d $AIIDALAB_APPS/quantum-espresso ]; then
+  echo "Quantum ESPRESSO app does exist"
+else
+  echo "Copying directory '$QE_APP_FOLDER' to '$AIIDALAB_APPS'"
+  cp -r "$QE_APP_FOLDER" "$AIIDALAB_APPS"
+fi
```
unkcpz, line 32: This do the trick, because in the k8s deployment, taring for the empty files like `work` for example has permission issue. Which not cause any problem I can see but it prevent the qeapp folder copy after. I remove it out as a independent operation.

before-notebook.d/00_untar-home.sh
```diff
@@ -4,7 +4,7 @@ set -eux
 home="/home/${NB_USER}"
 
 # Untar home archive file to restore home directory if it is empty
-if [[ $(ls -A ${home} | wc -l) = "0" ]]; then
+if [ ! -e $home/.FLAG_HOME_INITIALIZED ]; then
```
danielhollas, line 7: Not a huge fan of this solution since it seems brittle (user can remove this file). 
I am somewhat confused, why does the previous one not work anymore?
(feel free to ignore, this is just me rambling :D)

before-notebook.d/43_start-hq.sh
danielhollas, line 1: I didn't look into this file too deeply, might want to get other eyes on this.

danielhollas (approved):
Did a super quick review, LGTM. (btw: will be away for the next three weeks)

before-notebook.d/00_untar-home.sh
unkcpz: Two problems with previous one
1. the left side is `0` and right side is `\0` as str.
2. On the k8s deployment, if the persistent volume is used, it has a `lost+found` folder exist before this script is running.

So another solution I did was `if [ $(ls -A ${home} | wc -l ) -lt 1 ]; then`, but it is more brittle I assume :-p

before-notebook.d/43_start-hq.sh
unkcpz: Sure, please do it, I am not in hurry to get this merged.

unkcpz:
>  (btw: will be away for the next three weeks)

No problem, and congrats with the atomspec paper!

danielhollas:
Thanks! I've been meaning to announce it on the forum and thank the team but I did not get to it yet. : 🙈

unkcpz:
I merge this now, since in the demo server checkin meeting we set the next test from @cpignedoli  and @giovannipizzi at next Wednesday. It is good to have an alpha release include this one and many new fixes and use version tag to be deployed in the server.

PR #796, opened 2024-08-13
Code/Pseudos setup: Separate frontend from backend

danielhollas:
This is a preparatory PR for refactoring the code setup. It is imho better to separate the frontend (widgets that are displayed in the app) from the routines that actually install stuff in the backend and in the CLI. In this PR, I only move code, and don't do any actual changes yet.

As an immediate advantage, just executing `python -m aiidalab_qe --help` got noticeably faster since we no longer import the rest of the app.

In the next PR, instead of importing the install routines in the widgets, we can simply invoke the CLI from a subprocess. That way we do not need to worry about the threading issues, which in turn will allow us to get rid of generating the code setup code and execute it directly.


danielhollas:
> Thanks @danielhollas, would you mind to separate the renaming in to a commit

I am not quite sure what you mean, and not sure how to separate it. It's not really renaming, it's a code split. The backend code was moved from `aiidalab_qe.common` to `aiidalab.setup`. 

For example, most of the code from `aiidalab_qe/src/common/setup_code.py` has been moved to `aiidalab_qe/src/setup/code.py`, except for the `QeSetupWidget`. Does that make sense?

src/aiidalab_qe/common/setup_codes.py
```diff
@@ -1,215 +1,18 @@
...
-        # Assume that the installation was triggered by a different process.
-        yield "Installation was already started, waiting for it to finish..."
-        with FileLock(FN_LOCKFILE, timeout=120):
-            if not codes_are_setup():
-                raise RuntimeError(
-                    "Installation process did not finish in the expected time."
-                ) from None
-
 
 class QESetupWidget(ipw.VBox):
```
danielhollas, line 17: This file now only contains `QESetupWidget`, the rest was moved to `src/aiidalab_qe/setup/codes.py`

src/aiidalab_qe/common/setup_pseudos.py
```diff
@@ -1,230 +1,12 @@
...
-        with FileLock(FN_LOCKFILE, timeout=300):
-            if len(pseudos_to_install()) > 0:
-                raise RuntimeError(
-                    "Installation process did not finish in the expected time."
-                ) from None
+from ..setup.pseudos import install, pseudos_to_install
+from .widgets import ProgressBar
 
 
 class PseudosInstallWidget(ProgressBar):
```
danielhollas, line 12: This file now only contains `PseudosInstallWidget`, the rest was moved to `src/aiidalab_qe/setup/pseudos.py`

unkcpz:
Ah, okay, I thought it was rename and changes. If it is split, then nothing too much  can be simplified. I am reviewing it.

unkcpz (approved):
Thanks @danielhollas, looking forward to see next PR how you invoke from CLI to avoid threading issue.

Issue #798, opened 2024-08-22
Adding control for maximum number of iterations in scf cycle


croncagl:
When using the qe app, I noticed that in some systems the 'electron_maxstep = 80' is not enough to converge the scf cycle.

PR #800, opened 2024-08-30
Electron MaxStep Widget

AndresOrtegaGuerrero:
This PR address #798 , 
I have encounter that this limit with some materials is quite small , for example lanthanides , so is good to have this control.


mikibonacci:
Hi @AndresOrtegaGuerrero , LGTM!

mikibonacci (approved):

PR #802, opened 2024-09-09
Implementing lazy-loading

edan-bainglass:
This PR aims to improve the user's experience by isolating loading of widgets/resources to where they are needed (requested).

The overall approach is as follows (per widget):
1. Move bulk of `__init__` logic to a new `render` method
2. Add a loading widget (introduced in this PR) as a child in `__init__`
3. Observe the container of the widget (widget might be in an accordion, a tab, etc. -> observe the `selected_index` trait of the container)
4. On observable notification (e.g., tab switch), trigger widget's `render` method

Additionally, I take here an MVC (Model-View-Controller) approach to the app. The app's logic is migrated out to Model classes that represent the so-called "business-logic" of the app. Ideally, though perhaps not quite there yet, it should be possible to prepare and submit a QE calculation entirely via the Model classes!

A combination of a View and a Controller, here represented in a single class, is then simply a "window" to the Model's data. The View (the widgets) represent the UI. The Controller (the `_on_<event>` methods) is in charge of reacting to events either triggered by the user via the View, or triggered internally via the Model network.

Note that as I worked on this PR, several competing "flavors" of MVC came about. The present implementation is not a settled one. There are still some issues to address in order to properly establish our standard. One major issue is described below.

### Open design concern
Currently, the `ConfigurationModel`, which represents the data of step 2, is injected into all other `SettingsPanel`s. There, traits of the corresponding `SettingsModel` can be linked to the corresponding traits in the `ConfigurationModel`. This effectively syncs the specific settings with the app's configuration. There are presently a couple of idea to resolve this coupling between settings. Glad to discuss with those who are interested.

Glad to adjust where necessary 🙂

#### Update

Recent changes decouple the `ConfigurationModel` from the `SettingsPanel`s. The `AdvancedSettings` class is still coupled with its "sub-settings", e.g., smearing, magnetization, etc. This is primarily due to the highly coupled `set_model_state` and `get_model_state` methods of the `AdvancedModel`. But this okay for now, and perhaps in general, as the advanced settings is presented all together in one panel.

The decoupling hinges on a mechanism of announcing dependencies. For example, the PDOS plugin requires both the input structure and the protocol. Hence, it defines

```python
class PdosModel(SettingsModel):
    dependencies = [
        "input_structure",
        "workchain.protocol",
    ]

    input_structure = tl.Instance(orm.StructureData, allow_none=True)
    protocol = tl.Unicode(allow_none=True)
...
```

During plugin discovery, these dependencies are used to set up directional links, connecting the model to the configuration model network.

Note, however, that in decoupling the `ConfigurationModel`, I made a decision to extract the structure relaxation and property selection parts out from the basic settings. The new design of the configuration step is now split into sub-steps (thanks @mikibonacci for the suggestion to refer to these as step 2.x). 

Glad to discuss and adjust where necessary 🙂

### Documentation
Of course, once the PR is approved, documentation of the settled approach (or at least the current one) will be added!


PNOGillespie:
Hi @edan-bainglass. I can move the logic for updating the pseudopotentials (which is what lines 67 - 112 in `settings` do on starting the plugin) to be under `_update_element_select_panel()`. This should mean that the logic is only triggered by `_update_structure()`. Is that what you meant? Please let me know if I'm missing something.

I'm somewhat busy both this week and next, but I should be able to put together a separate PR soon for the changes if it's just a case of moving things around.

edan-bainglass:
Hi @PNOGillespie 👋 

> Hi @edan-bainglass. I can move the logic for updating the pseudopotentials (which is what lines 67 - 112 in `settings` do on starting the plugin) to be under `_update_element_select_panel()`. This should mean that the logic is only triggered by `_update_structure()`. Is that what you meant? Please let me know if I'm missing something.

Yes, this sounds right. As stated, @superstar54's approach in the XPS plugin is likely the way to go, for reference 🙂 Once the PR is ready, I test against my lazy-loading work and we see more clearly if it's sufficient 👍

> I'm somewhat busy both this week and next, but I should be able to put together a separate PR soon for the changes if it's just a case of moving things around.

No worries. The lazy-loading work will take a few pass (possibly incremental PRs), so the XAS part can be addressed once your PR is in.

Thanks 🙏

edan-bainglass:
@PNOGillespie @superstar54 just wondering, why were the XPS and XAS plugins included in the core app? Would it be better to maintain them as external plugins, say `aiidalab-qe-xps` and `aiidalab-qe-xas`, to be installed by those interested? What do you think?

edan-bainglass:
Hi all. The PR is ready for review. As mentioned, it is massive! If necessary, maybe we split the work a bit, say, one app step per person? Not sure if this helps. Anyhow, I'm available for questions :)

src/aiidalab_qe/plugins/bands/__init__.py
```diff
@@ -16,15 +17,15 @@ class BandsOutline(OutlinePanel):
...
-)
-
-
 bands = {
     "outline": BandsOutline,
-    "code": {"projwfc_bands": projwfc_code},
+    "model": BandsModel,
+    "code": {
+        "projwfc_bands": CodeModel(
+            description="projwfc.x",
```
AndresOrtegaGuerrero, line 25: I have a question, in the new CodeModel, if you notice the Pdos plugin also uses the projwfc.x, in the new model, is it possible that for the BandsModel, we use only one projwfc code , instead of the need of repeating it ? (I am asking , just in a follow up PR, there is a conflict of just having one projwfc.x )

src/aiidalab_qe/plugins/bands/model.py
```diff
@@ -0,0 +1,18 @@
...
+    projwfc_bands = tl.Bool(False)
+
+    def get_model_state(self):
+        return {"projwfc_bands": self.projwfc_bands}
+
+    def set_model_state(self, parameters: dict):
+        self.projwfc_bands = parameters.get("projwfc_bands", False)
+
+    def reset(self):
+        self.kpath_2d = False
```
AndresOrtegaGuerrero, line 18: This self.kpath_2d shount be here , since it was removed in the PR of fat_bands

tests/configuration/test_advanced.py
```diff
@@ -1,244 +1,218 @@
...
+    model = app.config_model
 
-    w.input_structure = structure
+    structure = generate_structure_data(name="silicon")
+    model.input_structure = structure
 
-    w.override.value = True
-    assert w.mesh_grid.value == "Mesh [16, 16, 16]"
+    model.advanced.override = True
+    assert model.advanced.mesh_grid == "Mesh [14, 14, 14]"
```
AndresOrtegaGuerrero, line 156: why this change to 16,16,16 ? the mesh is estimated based on a distance and since the structure was not modified it should be the same ? Which version of aiida-quantumespresso are you testing it ? maybe it was changed ?

src/aiidalab_qe/plugins/bands/__init__.py
edan-bainglass: I'm not sure. Had some discussion with @mikibonacci regarding duplication of codes. Some ideas there, but would take more time/discussion to handle. Is this a blocking matter?

src/aiidalab_qe/plugins/bands/model.py
edan-bainglass: Good catch. Will remove!

tests/configuration/test_advanced.py
edan-bainglass: Version `4.6.0`. I left some questions in the code regarding some of the tests. regarding expected results that didn't make sense to me. Will post them here in the discussion soon.

tests/conftest.py
```diff
@@ -206,45 +217,17 @@ def _generate_projection_data():
...
         dirpath = pathlib.Path(d)
 
-        for values in elements.values():
-            element = values["symbol"]
-
-            if element in actinides:
-                continue
-
+        for element in ELEMENTS:
             upf = generate_upf_data(element)
```
AndresOrtegaGuerrero, line 231: Is it a good idea just to restric the test of the sssp pseudos to just a few elements ?

tests/conftest.py
```diff
@@ -206,45 +217,17 @@ def _generate_projection_data():
...
     with tempfile.TemporaryDirectory() as d:
         dirpath = pathlib.Path(d)
 
-        for values in elements.values():
-            element = values["symbol"]
-
-            if element in actinides:
-                continue
-
+        for element in ELEMENTS:
```
AndresOrtegaGuerrero, line 230: Is it a good idea just to restrict to a few elements ? since this test is to test the sssp library?

tests/test_app.py
```diff
@@ -1,78 +1,68 @@
...
-        == 0
-    )
-    assert app.submit_step.pw_code.num_cpus.value == 4
+    assert app.struct_model.structure is None
+    assert app.config_model.workchain.relax_type == "positions_cell"
+    assert app.config_model.workchain.spin_type == "none"
+    assert app.config_model.get_model("bands").include is False
+    assert app.config_model.get_model("pdos").include is False
+    assert app.config_model.advanced.pseudos.ecutwfc == 30
+    assert len(app.config_model.advanced.pseudos.dictionary) == 1
```
AndresOrtegaGuerrero, line 33: Same here

src/aiidalab_qe/plugins/bands/__init__.py
AndresOrtegaGuerrero: No, i just wanted to know , if in a follow up PR , we can unified the codes for those two plugins , since is being requested

tests/test_app.py
```diff
@@ -1,78 +1,68 @@
...
-        )
-        == 0
-    )
-    assert app.submit_step.pw_code.num_cpus.value == 4
+    assert app.struct_model.structure is None
+    assert app.config_model.workchain.relax_type == "positions_cell"
+    assert app.config_model.workchain.spin_type == "none"
+    assert app.config_model.get_model("bands").include is False
+    assert app.config_model.get_model("pdos").include is False
+    assert app.config_model.advanced.pseudos.ecutwfc == 30
```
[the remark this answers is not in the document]
edan-bainglass: Right. Strange that the test is passing. Will check!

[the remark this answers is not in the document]
AndresOrtegaGuerrero: because it was changed , to 30

tests/configuration/test_advanced.py
AndresOrtegaGuerrero: I just tested it , and is 14, 14,14 for silicon with aiida-quantumespresso 4.6.0

danielhollas:
> Hi all. The PR is ready for review. As mentioned, it is massive! If necessary, maybe we split the work a bit, say, one app step per person? Not sure if this helps. Anyhow, I'm available for questions :)

I was planning to take a look at the overall approach but as you say this is a huuge PR. Would it make sense to split it up so that you have any necessary infrastructere (if any) and perhaps converting a single widget as an example. And then leave the rest for separate PRs?

edan-bainglass:
> I was planning to take a look at the overall approach but as you say this is a huuge PR. Would it make sense to split it up so that you have any necessary infrastructere (if any) and perhaps converting a single widget as an example. And then leave the rest for separate PRs?

I have no idea how to approach this. It is a complete restructuring of the app. Not sure how to isolate it into separate PRs without breaking other parts.

tests/test_app.py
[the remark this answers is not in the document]
edan-bainglass: This whole part was removed in the new commits. There is no longer a mechanism to reset the app via a process, since now a new workflow is done in a new tab. But this should be tested anyhow (resetting when no structure). I'll check and add a test if necessary. However, this might be low priority, as no structure means no submission.

tests/test_app.py
edan-bainglass: See comment above

edan-bainglass:
@superstar54 @mikibonacci @danielhollas @AndresOrtegaGuerrero open for review. Looking forward to your comments 🙂

src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -102,84 +136,148 @@ def __init__(self, **kwargs):
...
+                ipw.dlink(
+                    (self._model, "input_structure"),
+                    (outline.include, "disabled"),
+                    lambda _: not self._model.has_pbc,
+                )
+
+            def toggle_plugin(change, identifier=identifier, info=info):
+                if change["new"]:
+                    info.value = (
+                        f"Customize {identifier} settings in step 2.2 if needed"
```
AndresOrtegaGuerrero, line 243: "in step 2.2" is this 2.2 fix , or should be described differently ?

edan-bainglass: Step 2.2 is where you customize calculation parameters, so this is correct. Unless I'm misunderstanding your question.

AndresOrtegaGuerrero: I needed to tested to understand the Step 2.2 
image

Cant it have like a variable, instead of just 2.2 ? so in case the app add more things or something we know that is link to an {identifier} as well?

AndresOrtegaGuerrero: maybe put the S in step capital. -> Step

AndresOrtegaGuerrero:
I tried to load a previous calculation where I label one of the kinds of the structure as Au2 and i get this error , 

~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/hubbard.py in _update(self)
     91             return
     92         self._show_loading()
---> 93         self._model.hubbard.update()
     94         self._build_hubbard_widget()
     95         if isinstance(self._model.input_structure, HubbardStructureData):

~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in update(self)
    517     def update(self):
    518         with self.hold_trait_notifications():
--> 519             self._update_defaults()
    520             self.parameters = self._get_default_parameters()
    521             self.eigenvalues = self._get_default_eigenvalues()

~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in _update_defaults(self)
    560             )
    561         else:
--> 562             self.orbital_labels = self._get_labels()
    563             self._defaults["parameters"] = {label: 0.0 for label in self.orbital_labels}
    564             self.applicable_elements = [

~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in _get_labels(self)
    594     def _get_labels(self):
    595         symbols = self.input_structure.get_kind_names()
--> 596         hubbard_manifold_list = [
    597             self._get_manifold(Element(symbol))
    598             for symbol in self.input_structure.get_kind_names()

~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in <listcomp>(.0)
    595         symbols = self.input_structure.get_kind_names()
    596         hubbard_manifold_list = [
--> 597             self._get_manifold(Element(symbol))
    598             for symbol in self.input_structure.get_kind_names()
    599         ]

/opt/conda/lib/python3.9/enum.py in __call__(cls, value, names, module, qualname, type, start)
    382         """
    383         if names is None:  # simple value lookup
--> 384             return cls.__new__(cls, value)
    385         # otherwise, functional API: we're creating a new Enum type
    386         return cls._create_(

/opt/conda/lib/python3.9/enum.py in __new__(cls, value)
    700                 ve_exc = ValueError("%r is not a valid %s" % (value, cls.__qualname__))
    701                 if result is None and exc is None:
--> 702                     raise ve_exc
    703                 elif exc is None:
    704                     exc = TypeError(

ValueError: 'Au2' is not a valid Element

AndresOrtegaGuerrero:
> I tried to load a previous calculation where I label one of the kinds of the structure as Au2 and i get this error ,
> 
> ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/hubbard.py in _update(self) 91 return 92 self._show_loading() ---> 93 self._model.hubbard.update() 94 self._build_hubbard_widget() 95 if isinstance(self._model.input_structure, HubbardStructureData):
> 
> ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in update(self) 517 def update(self): 518 with self.hold_trait_notifications(): --> 519 self._update_defaults() 520 self.parameters = self._get_default_parameters() 521 self.eigenvalues = self._get_default_eigenvalues()
> 
> ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in _update_defaults(self) 560 ) 561 else: --> 562 self.orbital_labels = self._get_labels() 563 self._defaults["parameters"] = {label: 0.0 for label in self.orbital_labels} 564 self.applicable_elements = [
> 
> ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in _get_labels(self) 594 def _get_labels(self): 595 symbols = self.input_structure.get_kind_names() --> 596 hubbard_manifold_list = [ 597 self._get_manifold(Element(symbol)) 598 for symbol in self.input_structure.get_kind_names()
> 
> ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in (.0) 595 symbols = self.input_structure.get_kind_names() 596 hubbard_manifold_list = [ --> 597 self._get_manifold(Element(symbol)) 598 for symbol in self.input_structure.get_kind_names() 599 ]
> 
> /opt/conda/lib/python3.9/enum.py in **call**(cls, value, names, module, qualname, type, start) 382 """ 383 if names is None: # simple value lookup --> 384 return cls.**new**(cls, value) 385 # otherwise, functional API: we're creating a new Enum type 386 return cls._create_(
> 
> /opt/conda/lib/python3.9/enum.py in **new**(cls, value) 700 ve_exc = ValueError("%r is not a valid %s" % (value, cls.**qualname**)) 701 if result is None and exc is None: --> 702 raise ve_exc 703 elif exc is None: 704 exc = TypeError(
> 
> ValueError: 'Au2' is not a valid Element

The problem is also present when trying to set up labels for the structure

AndresOrtegaGuerrero:
> > I tried to load a previous calculation where I label one of the kinds of the structure as Au2 and i get this error ,
> > ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/hubbard.py in _update(self) 91 return 92 self._show_loading() ---> 93 self._model.hubbard.update() 94 self._build_hubbard_widget() 95 if isinstance(self._model.input_structure, HubbardStructureData):
> > ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in update(self) 517 def update(self): 518 with self.hold_trait_notifications(): --> 519 self._update_defaults() 520 self.parameters = self._get_default_parameters() 521 self.eigenvalues = self._get_default_eigenvalues()
> > ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in _update_defaults(self) 560 ) 561 else: --> 562 self.orbital_labels = self._get_labels() 563 self._defaults["parameters"] = {label: 0.0 for label in self.orbital_labels} 564 self.applicable_elements = [
> > ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in _get_labels(self) 594 def _get_labels(self): 595 symbols = self.input_structure.get_kind_names() --> 596 hubbard_manifold_list = [ 597 self._get_manifold(Element(symbol)) 598 for symbol in self.input_structure.get_kind_names()
> > ~/apps/quantum-espresso/src/aiidalab_qe/app/configuration/advanced/model.py in (.0) 595 symbols = self.input_structure.get_kind_names() 596 hubbard_manifold_list = [ --> 597 self._get_manifold(Element(symbol)) 598 for symbol in self.input_structure.get_kind_names() 599 ]
> > /opt/conda/lib/python3.9/enum.py in **call**(cls, value, names, module, qualname, type, start) 382 """ 383 if names is None: # simple value lookup --> 384 return cls.**new**(cls, value) 385 # otherwise, functional API: we're creating a new Enum type 386 return cls._create_(
> > /opt/conda/lib/python3.9/enum.py in **new**(cls, value) 700 ve_exc = ValueError("%r is not a valid %s" % (value, cls.**qualname**)) 701 if result is None and exc is None: --> 702 raise ve_exc 703 elif exc is None: 704 exc = TypeError(
> > ValueError: 'Au2' is not a valid Element
> 
> The problem is also present when trying to set up labels for the structure , We need to add a test for the Tags editor, or in the Hubbard test, give labels to the atoms to the atoms like Li2 for example
> 
> image

The issue come from this part that you modified , 

hubbard_manifold_list = [
            self._get_hubbard_manifold(Element(x.symbol))
            for x in self.input_structure.kinds
        ]

Since we can provide atoms "Tags" you need to do it like in the original code

src/aiidalab_qe/app/configuration/advanced/model.py
```diff
@@ -0,0 +1,927 @@
...
+            self.applicable_elements = [
+                *filter(
+                    lambda element: (
+                        element.is_transition_metal
+                        or element.is_lanthanoid
+                        or element.is_actinoid
+                    ),
+                    [
+                        Element(symbol)
+                        for symbol in self.input_structure.get_kind_names()
```
AndresOrtegaGuerrero, line 573: Check the original code , the condition should be , self.input_structure.get_symbols_set()

src/aiidalab_qe/app/configuration/advanced/model.py
```diff
@@ -0,0 +1,927 @@
...
+        return deepcopy(self._defaults["parameters"])
+
+    def _get_default_eigenvalues(self):
+        return deepcopy(self._defaults["eigenvalues"])
+
+    def _get_labels(self):
+        symbols = self.input_structure.get_kind_names()
+        hubbard_manifold_list = [
+            self._get_manifold(Element(symbol))
+            for symbol in self.input_structure.get_kind_names()
```
AndresOrtegaGuerrero, line 598: you should use the symbold , but not from the kinds , 
```
hubbard_manifold_list = [
            self._get_hubbard_manifold(Element(x.symbol))
            for x in self.input_structure.kinds
        ] 
```

src/aiidalab_qe/app/configuration/advanced/pseudos.py
```diff
@@ -0,0 +1,415 @@
...
+
+        children = []
+
+        elements = (
+            self._model.input_structure.get_kind_names()
+            if self._model.input_structure
+            else []
+        )
+
+        for index, element in enumerate(elements):
```
AndresOrtegaGuerrero, line 285: Here I would also be careful, since the model has a definition of element , is the element also has like a tag like 'Fe2' you might have an issue with the pseudos.

tests/test_app.py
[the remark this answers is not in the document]
AndresOrtegaGuerrero: but somehow is setting the model to a state where is setting a cutoff based on a previous input_structure , is important why the logic to reset the state of the model is not being updated

src/aiidalab_qe/plugins/bands/__init__.py
edan-bainglass: Let's discuss this further once the PR is merged

tests/test_app.py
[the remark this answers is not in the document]
edan-bainglass: This is indeed interesting. I can look into it. But I propose something different (to be discussed with the team) - if there is no input structure, there should be no messing with steps > 1! Same goes for other dependent reqs, e.g. no configuration parameters = no messing with steps > 2, etc.

src/aiidalab_qe/app/configuration/advanced/model.py
edan-bainglass: Good catch. Will modify!

src/aiidalab_qe/app/configuration/advanced/model.py
edan-bainglass: Not sure I understand what you're getting at. Here's the old code:

```python
        kind_list = self.input_structure.get_kind_names()
        hubbard_manifold_list = [
            self._get_hubbard_manifold(Element(x.symbol))
            for x in self.input_structure.kinds
        ]
        result = [
            f"{kind} - {manifold}"
            for kind, manifold in zip(kind_list, hubbard_manifold_list)
        ]
```

`.kinds` returns effectively `.get_kind_names()`, so other than the redundant call to it instead of using `symbols` defined above, I'm not seeing what's wrong. Can you please clarify?

src/aiidalab_qe/app/configuration/advanced/pseudos.py
edan-bainglass: If I understand you correctly, the lines above where I define `elements` should be referencing the `get_symbols_set` method instead of `get_kind_names` in order to obtain ALL kinds, including modified (tagged) kinds. Is this correct?

src/aiidalab_qe/app/configuration/advanced/model.py
AndresOrtegaGuerrero: Test the app , try to put a tag and see the app fail. I am testing with AiiDA v2.5.1 ,

AndresOrtegaGuerrero: In [2]: a.get_kind_names()
Out[2]: ['Co', 'O', 'Li', 'O3', 'Li1', 'Co2'] (these are the names) 

In [3]: a.kinds
Out[3]: 
[<Kind: name 'Co', symbol 'Co'>,
 <Kind: name 'O', symbol 'O'>,
 <Kind: name 'Li', symbol 'Li'>,
 <Kind: name 'O3', symbol 'O'>,
 <Kind: name 'Li1', symbol 'Li'>,
 <Kind: name 'Co2', symbol 'Co'>] 

So in your code it should fail since you get the the strings of the names, so it should fail for O3

edan-bainglass: I see. I misunderstood the `kind` objects. This makes sense. Thanks! I will adjust my code.

src/aiidalab_qe/app/configuration/advanced/pseudos.py
edan-bainglass: I get it now. `kind.name` includes tagged atoms, where `kind.symbol` is the elemental symbol. Updating code! Thanks again 🙏

superstar54 (changes requested):
Thanks for the nice work! I try to add my comments during the review. Some are discussed with Edan, but add here for the record.

1) One main benefit of the new design is that user can run the job without rendering the widget. I try to mimic this using the following code:
    
```python
from aiidalab_qe.app.main import App
from aiida import orm, load_profile

load_profile()
structure = orm.load_node(52587)
app = App(qe_auto_setup=False)
app.struct_model.structure = structure
app.struct_model.confirmed = True
app.config_model.confirmed = True
# app.submit_step.render()
app.submit_model.submit()
```
- For the moment, one still needs to render the submission step.
- There is `submit()` method for submit model, but not in struct and config mdoel, Suggest adding ones so that they share the same syntax.

2) in the PR, any change in the confirmed step, will trigger the reset of following step. This maybe too sensitive, suggest to change to only clicking `confirm` button trigger the reset of the following step.

superstar54:
Currently, you use two methods to control (trigger) the updates.
1) `on_xxx_change` in the settings. This is used in most of the cases, and I believe is more fit in the MVC design.
2) link between two traits from models, for some special parameters, e.g., the `protocol`. This is convenient, but sometimes it's hard to follow the logic in the code because one cannot easily find what links are there and how many.

In principle, the functionality of method 2 can also be implemented using method 1, with more code and clear logic. This is open for discussion.

edan-bainglass:
> Currently, you use two methods to control (trigger) the updates.
> 
> 1. `on_xxx_change` in the settings. This is used in most of the cases, and I believe is more fit in the MVC design.
> 2. link between two traits from models, for some special parameters, e.g., the `protocol`. This is convenient, but sometimes it's hard to follow the logic in the code because one cannot easily find what links are there and how many.
> 
> In principle, the functionality of method 2 can also be implemented using method 1, with more code but clear logic. This is open for discussion.

Thank you @superstar54. Note that 99% of the time, I use 1 + 2, e.g., the `protocol` trait of the basic settings model is linked with that of the advanced settings model, but the actual trigger of the change of a protocol is the `on_protocol_change` method of the advanced settings. This approach is what I mostly use.

Indeed, there are a small number of cases where I use 1 without 2. This is perhaps limited to the main `App` widget when transferring dependencies between wizard steps. But in principle, these can also be achieved using the 1 + 2 approach. I recall taking the 1 without 2 approach here due to performance, but I think this is no longer an issue. A 1 + 2 approach will work all the same.

That said, both approaches are standard. 1 + 2 is the so-called Observer pattern, where the other is the Mediator pattern. There are pros and cons to both. Perhaps we can discuss these when you have time 🙂

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -47,140 +46,164 @@ class StructureSelectionStep(ipw.VBox, WizardAppWidgetStep):
...
+        )
+        ipw.dlink(
+            (self._model, "structure"),
+            (self.manager, "structure"),
+            lambda structure: structure.get_ase() if structure else None,
+        )
+        ipw.dlink(
+            (self.manager, "structure_node"),
+            (self._model, "structure"),
+        )
```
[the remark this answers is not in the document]
edan-bainglass: This is indeed an issue. I need to think about this a bit. The model -> manager link was added to handle cases where you load a structure from a process before rendering the manager. Maybe there is another way to do it. Will think about it.

src/aiidalab_qe/app/main.py
```diff
@@ -22,40 +28,50 @@ class App(ipw.VBox):
     process = tl.Union([tl.Unicode(), tl.Int()], allow_none=True)
 
     def __init__(self, qe_auto_setup=True):
+        # Initialize the models
+        self.struct_model = StructureModel()
```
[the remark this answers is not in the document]
edan-bainglass: Sure 🙂

edan-bainglass:
Thanks for the review @superstar54. See my comments below.

> 1. One main benefit of the new design is that user can run the job without rendering the widget. I try to mimic this using the following code:
> 
> ```python
> from aiidalab_qe.app.main import App
> from aiida import orm, load_profile
> 
> load_profile()
> structure = orm.load_node(52587)
> app = App(qe_auto_setup=False)
> app.struct_model.structure = structure
> app.struct_model.confirmed = True
> app.config_model.confirmed = True
> # app.submit_step.render()
> app.submit_model.submit()
> ```

This is indeed a benefit, though I am not sure of the use case where one would do this. Another benefit that @mikibonacci correctly pointed out is that now that the logic of the app is contained in the models (almost - see note regarding submit model below), the view is no longer limited to the ipywidgets package. We could in principle, perhaps with some minimal work, replace it with another tech, say, Solara 🙂
 
> * For the moment, one still needs to render the submission step.

Indeed, this sucks. I left it this way for now due to the extra work required to extract certain resources traits from the code widgets. But this can be done now that the rest is in place 👍 

> * There is `submit()` method for submit model, but not in struct and config mdoel, Suggest adding ones so that they share the same syntax.

Agreed 👍

> 2. in the PR, any change in the confirmed step, will trigger the reset of following step. This maybe too sensitive, suggest to change to only clicking `confirm` button trigger the reset of the following step.

This can also be done 👍 But let's discuss in person.

mikibonacci (commented):
Hi @edan-bainglass, thank you very much for the amazing work. 

I was testing the branch in a fresh container. When I confirm the structure in step 1, the app crashes because no pseudo is found in the db. Indeed, there are no pseudo families installed. 

I have to check more the code, but I guess these are installed when step 3 is rendered? It should happen here, in the `render` method of the `SubmitQeAppWorkChainStep`:

```python
 if self.qe_auto_setup:
      self.sssp_installation.refresh()
```

When I installed the 4 pseudo flavours - (sssp)x(precision,efficiency)x(PBE,PBEsol) -  and I restart the app, I don't have the crash. 

The problem is that we install the pseudos in step 3, but we need them in step 2. 

The question is: as we are doing the installation in a separated Thread, should we just start it when we start the App? The same for the codes, actually.

mikibonacci:
Quite sure it is not the case for standard `aiidalab/full-stack` image (was it one of the reason why the container was taking a lot of time to start?). It is actually the case for the image directly generated via the [aiidalab-qe Dockerfile](https://github.com/aiidalab/aiidalab-qe/blob/main/Dockerfile) - The one used for the demo-server. Anyway, let's wait for @superstar54

edan-bainglass:
Running some tests. Ran a Si2, no relax, fat bands calculation. Calculation failed, but at the moment I'm more concerned about the following error raised on clicking the "Kill workchain" button:

```python
TypeError                                 Traceback (most recent call last)
~/apps/quantum-espresso/src/aiidalab_qe/app/result/__init__.py in _on_kill_button_click(self, _)
    150 
    151     def _on_kill_button_click(self, _):
--> 152         self._model.kill_process()
    153         self._update_kill_button_layout()
    154 

~/apps/quantum-espresso/src/aiidalab_qe/app/result/model.py in kill_process(self)
     28     def kill_process(self):
     29         if process := self._get_process_node():
---> 30             control.kill_processes(process)
     31 
     32     def clean_remote_data(self):

/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/control.py in kill_processes(processes, message, all_entries, timeout, wait)
    205 
    206     controller = get_manager().get_process_controller()
--> 207     _perform_actions(processes, controller.kill_process, 'kill', 'killing', timeout, wait, msg=message)
    208 
    209 

/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/control.py in _perform_actions(processes, action, infinitive, present, timeout, wait, **kwargs)
    231     futures = {}
    232 
--> 233     for process in processes:
    234         if process.is_terminated:
    235             LOGGER.error(f'Process<{process.pk}> is already terminated.')

TypeError: 'WorkChainNode' object is not iterable
```

I started from a fresh DB, so there is only one `WorkChainNode`.

AndresOrtegaGuerrero:
> Hi @edan-bainglass, thank you very much for the amazing work.
> 
> I was testing the branch in a fresh container. When I confirm the structure in step 1, the app crashes because no pseudo is found in the db. Indeed, there are no pseudo families installed.
> 
> I have to check more the code, but I guess these are installed when step 3 is rendered? It should happen here, in the `render` method of the `SubmitQeAppWorkChainStep`:
> 
> ```python
>  if self.qe_auto_setup:
>       self.sssp_installation.refresh()
> ```
> 
> When I manually install the 4 pseudo flavours - (sssp)x(precision,efficiency)x(PBE,PBEsol) - and I restart the app, I don't have the crash.
> 
> The problem is that we install the pseudos in step 3, but we need them in step 2. I think this also makes not possible to submit a `QeAppWorkChain` just via the "backend" models, as @superstar54 did, if we don't have the pseudo and codes installed? The `refresh` method (which triggers the installation) is assigned to the Widget, not to the model, right?
> 
> The question is: as we are doing the installation in a separated Thread, should we just start it when we start the App (as it was done before)? The same for the codes, actually. Or trigger the installation and put a blocker in the very same step 2.
> 
> I am also thinking if we should pack these installations when we download the QE app from the store, in such a way to completely discard these fragile procedures. But maybe this is another story.

@edan-bainglass I will also track if the PseudoDojo, in particular the FR (Full relativistic) are installed for the SOC case.

edan-bainglass:
@AndresOrtegaGuerrero @mikibonacci my last commit addresses the issue by installing SSSP and setting up QE on startup, as suggested by @mikibonacci. We take a slight hit in startup time just to trigger these two tasks, but as @mikibonacci stated, these are done in separate threads, so the app is free to proceed. Hope this resolves the issue.

mikibonacci:
> @AndresOrtegaGuerrero @mikibonacci my last commit addresses the issue by installing SSSP and setting up QE on startup, as suggested by @mikibonacci. We take a slight hit in startup time just to trigger these two tasks, but as @mikibonacci stated, these are done in separate threads, so the app is free to proceed. Hope this resolves the issue.

Thanks @edan-bainglass ! 
I think the last bit to solve the issue is some blocker for step 2 (or maybe even for the `Confirm` of step 1?). Otherwise, if the connection is too slow, we are still going to run into the crash - if we go too fast through the app

ps. maybe we can just open an issue and solve this later

edan-bainglass:
> Thanks @edan-bainglass ! I think the last bit to solve the issue is some blocker for step 2 (or maybe even for the `Confirm` of step 1?). Otherwise, if the connection is too slow, we are still going to run into the crash - if we go too fast through the app
> 
> ps. maybe we can just open an issue and solve this later

We might want to move the SSSP installation and QE setup widgets to the `App` component if they are truly blockers. So on start, if these have not yet been installed/setup, the user will see the progress bars there. Currently these are in step 3.

I'll open an issue 👍

mikibonacci:
> Running some tests. Ran a Si2, no relax, fat bands calculation. Calculation failed, but at the moment I'm more concerned about the following error raised on clicking the "Kill workchain" button:
> 
> ```python
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> ~/apps/quantum-espresso/src/aiidalab_qe/app/result/__init__.py in _on_kill_button_click(self, _)
>     150 
>     151     def _on_kill_button_click(self, _):
> --> 152         self._model.kill_process()
>     153         self._update_kill_button_layout()
>     154 
> 
> ~/apps/quantum-espresso/src/aiidalab_qe/app/result/model.py in kill_process(self)
>      28     def kill_process(self):
>      29         if process := self._get_process_node():
> ---> 30             control.kill_processes(process)
>      31 
>      32     def clean_remote_data(self):
> 
> /opt/conda/lib/python3.9/site-packages/aiida/engine/processes/control.py in kill_processes(processes, message, all_entries, timeout, wait)
>     205 
>     206     controller = get_manager().get_process_controller()
> --> 207     _perform_actions(processes, controller.kill_process, 'kill', 'killing', timeout, wait, msg=message)
>     208 
>     209 
> 
> /opt/conda/lib/python3.9/site-packages/aiida/engine/processes/control.py in _perform_actions(processes, action, infinitive, present, timeout, wait, **kwargs)
>     231     futures = {}
>     232 
> --> 233     for process in processes:
>     234         if process.is_terminated:
>     235             LOGGER.error(f'Process<{process.pk}> is already terminated.')
> 
> TypeError: 'WorkChainNode' object is not iterable
> ```
> 
> I started from a fresh DB, so there is only one `WorkChainNode`.
> 
> ## Update
> Okay, might be a bug on my end. I ran another and got the same issue with the Kill button, so not a single-process issue. Will investigate!

Hi @edan-bainglass , looking into this, the thing is that the `process` in `control.kill_processes(process)` should be a list (or an iterable). if you just replace with `control.kill_processes([process])`. 
When I introduced this, I indeed put this code:

```python
def _on_click_kill_button(self, _=None):
        """callback for the kill button.
        First kill the process, then update the kill button layout.
        """
        workchain = [orm.load_node(self.process)] # <- HERE
        control.kill_processes(workchain)
```

superstar54:
> I'm told by @superstar54 that the new image has these pre-installed?

- The aiidalab-qe image has QE and pseudos already installed.
- The aiidalab/full-stack, in which the user installs QE App from the app store, will trigger the `post-install` script to install QE and pseudos.

QE App will check whether the QE and pseudos are available or not when launching (by default `qe_auto_setup = True`).

The reason the blocker is in step 3 is that we want to allow the user to play with the app as soon as they install it. since the installation is running in another thread. But we can improve if there are better options.

mikibonacci:
> > I'm told by @superstar54 that the new image has these pre-installed?
> 
> * The aiidalab-qe image has QE and pseudos already installed.
> * The aiidalab/full-stack, in which the user installs QE App from the app store, will trigger the `post-install` script to install QE and pseudos.
> 
> QE App will check whether the QE and pseudos are available or not when launching (by default `qe_auto_setup = True`).
> 
> The reason the blocker is in step 3 is that we want to allow the user to play with the app as soon as they install it. since the installation is running in another thread. But we can improve if there are better options.

Right @superstar54 , I always forget about the `post_install`. I guess then the solution is just putting a blocker then, or actually even better just removing the `disown` [here](https://github.com/aiidalab/aiidalab-qe/blob/ef677e75ae306bcaf36504d89288cc7bfaf2c38b/post_install#L4), so we stay there until is ready. The user needs to wait a bit for the installation, but we are sure he has internet connection to download stuff and greatly simplifies our lifes.

Anyway, this is not the place to talk about this, let's discuss this after this PR.

edan-bainglass:
@superstar54 please review how I handled the circular structure link in step 1 🙏

edan-bainglass:
Hmm, that's strange. All tests pass locally 🤔 @superstar54 maybe you can have a look?

Anyhow, a few comments regarding the latest push:

### Primary change
- Decoupled the code models from the code widgets - options, selected option, and computational resources are now all traits in the model of a given code, with the widget syncing up with the model on render
- Currently, the computation resources widgets automatically trigger the retrieval of code options from the DB. I plan to open a PR to make this optional, as in my design, the model is the source of truth and takes care of fetching the options

### Peripheral changes
- Added a `--skip-slow` pytest cli option that can be used during development to skip a few tests that take >3 minutes (now marked with `@pytest.mark.slow` (mostly pseudos installation stuff)
- Session-scoped the pseudo-group-mocking fixtures to improve performance and reduce issues with profile clean up

@superstar54 step 3 is ready for review 🙏
@AndresOrtegaGuerrero step 2 is ready for a more detailed review 🙏
To all others, if you haven't yet looked at the changes, please do. And please test locally. I'm actively updating where needed, so please let me know if anything breaks.

### Looking into the following next
- When loading a process, opening the structure step triggers a reset of the app (due to the injection of the structure into the manager. Need to fix this! Also, @superstar54 pointed out that the label is missing when the structure is loaded, though this seems to be an issue already in `main`
- The kill-process/clean-remote buttons do not respond dynamically to the state of a running job. Need to look into this. @mikibonacci did they when you implemented them? In other words, when the job finishes, did the kill-process button vanish and clean-remote button appear? I'll also look into this.

### Planned PRs (good to get your comments here already)
- Rethink the testing suite
- Consider "disabling" steps following a not-yet-confirmed step

edan-bainglass:
As I was fixing the issue with the kill/clean process buttons not reacting to the process state, I found that the issue was a SQLAlchemy one - the process monitor thread, when attempting to update the state of the results section, was using a process node fetched by the primary thread. This one's on me! To improve performance by reducing DB calls, I had modified the results section's model to store the process node on first fetch. I now understand how this may lead to SQLAlchemy session conflicts. It is now fixed!

That said, I suspect similar performance improvements when handling codes (and perhaps pseudos) may be the reason why tests are failing (all with SQLAlchemy errors). Working on it! Note that it is still curious that it does not occur locally, where tests pass).

src/aiidalab_qe/app/structure/__init__.py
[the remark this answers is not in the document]
edan-bainglass: This is now resolved. The model -> manager link is removed! It was necessary for the case of a structure loaded from a process. This is now injected into the manager after it is initialized, **but before it is linked to the model**, so that the step is not reset on render.

edan-bainglass:
@superstar54 your requests have been addressed. See comments below 🙂 

> 1. One main benefit of the new design is that user can run the job without rendering the widget. I try to mimic this using the following code:
> 
> ```python
> from aiidalab_qe.app.main import App
> from aiida import orm, load_profile
> 
> load_profile()
> structure = orm.load_node(52587)
> app = App(qe_auto_setup=False)
> app.struct_model.structure = structure
> app.struct_model.confirmed = True
> app.config_model.confirmed = True
> # app.submit_step.render()
> app.submit_model.submit()
> ```
> 
> * For the moment, one still needs to render the submission step.

This is in principle no longer true. The submission step required rendering to trigger the rendering (and setup) of the codes. This requirement was due to the dependency of code models on their widget counterparts for options, selected option, and computational resources. This has been resolved by fully decoupling the code models from the widgets. @superstar54 please test this in your toy models 🙏

> * There is `submit()` method for submit model, but not in struct and config mdoel, Suggest adding ones so that they share the same syntax.

Step 1 and 2 now have `.confirm()`. Matter of fact, they are now both `Confirmable` - a new mixin class adding the trait and logic of being (un)confirmed. This mixin is used throughout where appropriate.

> 2. in the PR, any change in the confirmed step, will trigger the reset of following step. This maybe too sensitive, suggest to change to only clicking `confirm` button trigger the reset of the following step.

As mentioned before, this may lead to inconsistent states and is considered bad behavior. Internal discussion suggested handling this matter differently, perhaps by "disabling" steps post unconfirmed steps. Discussion and testing remains, but this will be a separate PR.

edan-bainglass:
@PNOGillespie could you give this PR a go, specifically your XAS part? Great if we can get a quick check to verify all is well 🙏🙏🙏

superstar54:
Hi @edan-bainglass , thanks for the update! I did thoughtful tests on the PR for different calculations and plugins. Everything works great, except for some issues, which I reported here:

Running a Job:

- The `clean remote data` button does not show when the job is finished.
- When modifying a confirmed step, what's the strategy to reset the following steps? In my test, modifying step 1 reset the step 2 completely, but step 3 is not reset.
- Another issue related to XAS, see #920 

Loading the previous job:
- pdos plugin: `NSCF K-points distance ` is not set to the value customly set in the previous job.
- xps plugin: the `Select core-level` section, checkbox before `Si_2p` is not selected.
- The Results panels are not shown automatically. I need to manually click on the `Update results` button.
- At some tests (randomly), the step 3 always shows the `Loading pw code` without success, see the attached video. I need to refresh the page to make it work.

https://github.com/user-attachments/assets/7a2826a6-847c-4927-929c-110634eb7f15

Free feel to discuss in more detail, or if you need any help to fix these issues.

superstar54:
I tried to run the app without GUI. Could you check if my script is correct? becuase I got an error:
```python
from aiidalab_qe.app.main import App
from aiida import orm, load_profile

load_profile()
structure = orm.load_node(2402)
app = App(qe_auto_setup=False)
app.structure_model.structure = structure
app.structure_model.confirm()
app.configure_model.confirm()
app.submit_model.confirm()
```


```---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_76841/3511553373.py in <cell line: 11>()
      9 app.configure_model.confirm()
     10 # app.submit_step.render()
---> 11 app.submit_model.confirm()

~/apps/aiidalab-qe/src/aiidalab_qe/app/submission/model.py in confirm(self)
     82     def confirm(self):
     83         if not self.process_node:
---> 84             self._submit()
     85         super().confirm()
     86 

~/apps/aiidalab-qe/src/aiidalab_qe/app/submission/model.py in _submit(self)
    290     def _submit(self):
    291         parameters = self.get_model_state()
--> 292         builder = self._create_builder(parameters)
    293 
    294         with self.hold_trait_notifications():

~/apps/aiidalab-qe/src/aiidalab_qe/app/submission/model.py in _create_builder(self, parameters)
    312 
    313     def _create_builder(self, parameters) -> ProcessBuilderNamespace:
--> 314         builder = QeAppWorkChain.get_builder_from_protocol(
    315             structure=self.input_structure,
    316             parameters=deepcopy(parameters),  # TODO why deepcopy again?

~/apps/aiidalab-qe/src/aiidalab_qe/workflows/__init__.py in get_builder_from_protocol(cls, structure, parameters, **kwargs)
    125                 value["code"] = orm.load_node(value["code"])
    126         # update pseudos
--> 127         for kind, uuid in parameters["advanced"]["pw"]["pseudos"].items():
    128             parameters["advanced"]["pw"]["pseudos"][kind] = orm.load_node(uuid)
    129         #

KeyError: 'pseudos'
```

edan-bainglass:
@superstar54 thank you for reporting the issues. I will investigate the cause and update accordingly.

src/aiidalab_qe/common/mixins.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __new__(cls, name, bases, classdict):
+        if len(bases) == 1 and not issubclass(bases[0], tl.HasTraits):
+            raise MissingMixinError()
+        for base in bases:
+            if issubclass(base, tl.HasTraits):
+                for name, trait in base.class_traits().items():
+                    if name not in classdict:
+                        classdict[name] = trait
+        bases = tuple(filter(lambda base: base is not tl.HasTraits, bases))
```
superstar54, line 33: I think you want to remove the classes which are the subclass of `tl.HasTraits`.
```suggestion
        bases = tuple(filter(lambda base: not issubclass(base, tl.HasTraits), bases))
```

src/aiidalab_qe/common/mixins.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __init__(self, *args: object) -> None:
+        super().__init__("expected at least one mixin in class definion", *args)
+
+
+class MetaMixinTraitsRegister(tl.MetaHasTraits):
+    """A metaclass to register traits from `HasTraits`-subclassed mixins.
+
+    The metaclass removes the `HasTraits` base class from the mixin to
+    avoid MRO conflicts.
```
superstar54, line 22: What are the kinds of `MRO conflicts`? Could you explain in more detail?

superstar54 (changes requested):
Hi @edan-bainglass , One issue I found is the `HasTraitsAndMixins`, 
- there is a bug which affects its functionality, it does not filter the base class correctly. However, the class still works, so I am wandered what the initial problem you want to solve using this class. You mentioned the `MRO conflicts`, but it not very clear to me.
- If it filters the base classes correctly, then there is another issue. You only registered the traits, but missed others, e.g., the `has_structure` method in the `HasInputStructure`.

src/aiidalab_qe/app/result/__init__.py
```diff
@@ -35,183 +52,160 @@ def __init__(self, **kwargs):
...
+        ]
+
+        self.rendered = True
+
         self._update_kill_button_layout()
+        self._update_clean_scratch_button_layout()
 
     def can_reset(self):
-        "Do not allow reset while process is running."
         return self.state is not self.State.ACTIVE
```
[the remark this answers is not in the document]
edan-bainglass: Potentially accidental

src/aiidalab_qe/common/mixins.py
edan-bainglass: Sure. Take a class that extends `HasTraits`. Now add a mixin that extends `HasTraits`. You now have `HasTraits` twice, which will result in an MRO conflict - cannot resolve the two `HasTraits` instances. The metaclass is intended to resolve this, yielding in the end only a single `HasTraits` while still accounting for all traits.

superstar54:
> Can you describe your test that resulted in the issue?
See the comment, `I think you want to remove the classes which are the subclass of tl.HasTraits.`

edan-bainglass:
> > Can you describe your test that resulted in the issue?
> 
> See the comment, `I think you want to remove the classes which are the subclass of tl.HasTraits.`

@superstar54 not sure if this is why you got an issue, but while working on the results section, I ran into an issue with the mixin and updated it to the following:
```python
class MetaMixinTraitsRegister(tl.MetaHasTraits):
    """A metaclass to register traits from `HasTraits`-subclassed mixins.

    The metaclass removes the `HasTraits` base class from the mixin to
    avoid MRO conflicts.
    """

    def __new__(cls, name, bases, classdict):
        for base in bases:
            if issubclass(base, tl.HasTraits):
                for name, trait in base.class_traits().items():
                    if name not in classdict:
                        classdict[name] = trait
        bases = (
            *filter(lambda base: base is not tl.HasTraits, bases),
            tl.HasTraits,
        )
        return super().__new__(cls, name, bases, classdict)
```
This allowed me to use non-trait-ful mixins, such as the `HasModels` mixin. But maybe let's discuss the original problem and the `HasTraitsAndMixins` solution anyhow. Perhaps there's improvement to be had 🙂

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -0,0 +1,306 @@
...
+                    ERROR: required pseudo family `{self.family}` is
+                    not installed. Please use `aiida-pseudo install` to install
+                    it."
+                </div>
+            """
+        except ValueError as exception:
+            self._status_message = f"""
+                <div class='alert alert-danger'>
+                    ERROR: failed to obtain recommended cutoffs for pseudos
+                    `{pseudo_family}`: {exception}
```
mikibonacci, line 160: I would add some details here, like "Please select a custom ecut and ecutrho"... But actually, 
the `get_builder_from_protocol` will except because indeed no cutoff are found. So, 
We can put actuallly a "please select another pseudopotential family". This is kind of a bug
from aiida-pseudo, hopefully we can solve it easily

edan-bainglass: Agreed! Could you open an issue (to be PR'd soon)? 🙏

danielhollas:
@edan-bainglass it would be very helpful to me if you could point me to the things you want me to look at, I am feeling a bit lost right now. :-) Or we can wait till Monday, where you will present the approach in general?

edan-bainglass:
> @edan-bainglass it would be very helpful to me if you could point me to the things you want me to look at, I am feeling a bit lost right now. :-) Or we can wait till Monday, where you will present the approach in general?

Best if I walk you through it on Monday. However, in the meantime, you can take a look at #919, where I've started documenting the core principles.

AndresOrtegaGuerrero:
the behavior of the bandswidget it has been modifie. When moving the fat band widht , it doesnt keep the zoom breaking #904

AndresOrtegaGuerrero:
The result tabs are not displayed automatically when loading a job. They only become visible after clicking on the QeAppWorkChain node. Similarly, clicking the "Update Results" button does not trigger the display of the result tab

edan-bainglass:
@AndresOrtegaGuerrero thank you for the comments. Let's discuss at the retreat 🙂

AndresOrtegaGuerrero:
> The behavior of the BandsWidget has been modified. Adjusting the fat band width widget no longer preserves the zoom level, causing it to reset. The zoom level should remain unchanged when modifying the plot contributions, maintaining consistency across interactions. This breaks the behaviour of #904

This is solved now

danielhollas:
Huge congrats on shipping this! :rocket: Amazing work. 
75 commits, 119 files changes, 97 review comments! :exploding_head: 

image

PR #806, opened 2024-09-24
XAS: Move Startup Logic to `_update_structure()`

PNOGillespie:
Addresses requirements for lazy-loading (PR #802) by moving logic for checking core-hole pseudopotentials to `_update_structure()`.


PNOGillespie:
Hi @edan-bainglass, apologies for the slow speed on this one, but this should be ready to go. Let me know if you need anything else.

One other thing I did notice (which @superstar54 should know about) when running test workflows is that the changes which add the `tot_charge` feature (#402) are actually interfering with the input parameters for XAS calculations. I think this is simply because the default (if no `tot_charge` is set by the user) explicitly sets `tot_charge = 0` in the overrides, rather than removing the parameter from the overrides entirely, which ultimately changes the inputs for XAS and causes calculations to fail on the QE side of things.

I'll check this myself today, open an issue, and get some sort of fix ready.

edan-bainglass:
@PNOGillespie on the contrary, thank you so much for quickly addressing it 🙏 I'll give it a look later today and test on my end. Thanks again 🙂

PR #808, opened 2024-09-24
Add slider to adjust fat bands width

t-reents:
@AndresOrtegaGuerrero as discussed, I added a slider to control the width of the fat bands. It will only become visible once you enable the fat bands. In agreement with the previous logic, you need to click the update button to adjust the width. The slider covers a range from 10 to 750 meV/atom as the maximum width of a fat band.


AndresOrtegaGuerrero (approved):
I tested it it works really well, thank you for your Job , LGTM

Issue #817, opened 2024-09-26
The sentence “Currently only 3D structures supported” is not true anymore


superstar54:
The sentence “Currently only 3D structures supported” (Text written at beginning of Step 1) is not true anymore, right? Please remove this sentence.
- Instead, in the structure editor, under “Edit StructureData”, add also “no periodicity” as an option. And test for a molecule if this works. (E.g. all mesh of k-points become 1x1x1 = Gamma only). E.g. try PDOS of a molecule, we have ETFA molecule (add also a simple molecule like O2 or CO2 to the examples!).
- In any case, ensure that the issue discussing what needs to be done for low-dimensional systems (e.g. what to relax, whether to add a Martyna-Tuckermann or similar, how many k-points, …) is in the milestone for the release in 6 months (~march 2025).ma

cpignedoli:
also connected to https://github.com/aiidalab/aiidalab-qe/issues/734

Issue #734, opened 2024-05-16
Handling Molecules


AndresOrtegaGuerrero:
The current app misses the possibility of treating properly Molecules and tricks have to be used in this case: the user has to override defaults and reduce k-point mesh accuracy to have only 1 k-point.

Moreover, Bandstructure is not deactivated.

cpignedoli:
I would add a simple pre-check of the structure:

- if the structure has no cell add a std cell (e.g. +5A in each direction) treat as a molecule and inform teh user of the action taken ointing out how this can be overridden

- if the structure has a cell, with simple distance criteria guess whether it is PBC (111), (110), (011),... (000). If periodicity is not (111) preset accordingly the cell periodicity tabs inform the user of the presetting done and inform on how to override presetting in case needed
- If the system is treated as a molecule, prset k-points to 1 1 1 and blovk bandstructure calculations

superstar54:
Thanks, @cpignedoli , for the suggestions.

> what about other plugins?

This pre-check is done in step 1 when the user confirms the structure, so it applies to all plugins.

> if the structure has no cell add a std cell (e.g. +5A in each direction) treat it as a molecule and inform the user of the action taken pointing out how this can be overridden

One suggestion of the cell size: since in PW, we usually use [MT](https://www.quantum-espresso.org/Doc/INPUT_PW.html#:~:text=t%27%2C%20%27-,mt,-%27%20%3A) correction for an isolated system, we need to add a cell that guaranteed the cell is twice as large as charge density (~the molecule's size).

AndresOrtegaGuerrero:
From #817 

Instead, in the structure editor, under “Edit StructureData”, add also “no periodicity” as an option. And test for a molecule if this works. (E.g. all mesh of k-points become 1x1x1 = Gamma only). E.g. try PDOS of a molecule, we have ETFA molecule (add also a simple molecule like O2 or CO2 to the examples!).

In any case, ensure that the issue discussing what needs to be done for low-dimensional systems (e.g. what to relax, whether to add a Martyna-Tuckermann or similar, how many k-points, …) is in the milestone for the release in 6 months (~march 2025).ma

PR #828, opened 2024-09-26
removing text 3D structures only

AndresOrtegaGuerrero:
This PR address #817 

I think #817 should be split it into two issues, this one can be to remove, and another one more related about allowing to work with molecules likehttps://github.com/aiidalab/aiidalab-qe/issues/734


superstar54 (approved):
LGTM, thanks!

PR #833, opened 2024-09-30
Feature/degauss widget pdos tab

AndresOrtegaGuerrero:
This tab includes a degauss widget to the pdos plugin
This useful for some case where the PDOS calculation doesnt obtain some projections (is a projwfc,x bug , that Giovanni mentioned) , this is a way around to avoid this. 

image

This solves the issue of core electron density not displayed before for Au 

Now
image

Before 
image

This PR is quite important since it also affects states around fermi level like in this case , if you noticed there are some missing states in the pdos

image

### DIFF
--- a/src/aiidalab_qe/plugins/pdos/setting.py
+++ b/src/aiidalab_qe/plugins/pdos/setting.py
@@ -33,12 +33,17 @@ class Setting(Panel):
             disabled=False,
             style={"description_width": "initial"},
         )
+        self.use_pdos_degauss = ipw.Checkbox(
+            value=False,
+            description="Use custom PDOS degauss",
+            style={"description_width": "initial"},
+        )
         self.pdos_degauss = ipw.BoundedFloatText(
             min=0.001,
             step=0.001,
             value=0.01,
             description="PDOS degauss (Ry):",
-            disabled=False,
+            disabled=True,
             style={"description_width": "initial"},
         )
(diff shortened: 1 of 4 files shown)


cpignedoli (changes requested):
The PDOS smearing affects both DOS and PDOS so teh name of teh widget shoudl be different.
-it must be NOT active by default
-It should have a default value that provides similar results compared to default tetraedron parameters (so something close to 0.005 for protocol moderate --I tested Au for this--))

src/aiidalab_qe/plugins/pdos/setting.py
```diff
@@ -31,15 +33,45 @@ def __init__(self, **kwargs):
...
         )
+        self.use_pdos_degauss = ipw.Checkbox(
+            value=False,
+            description="Use custom PDOS degauss",
+            style={"description_width": "initial"},
+        )
+        self.pdos_degauss = ipw.BoundedFloatText(
+            min=0.001,
+            step=0.001,
+            value=0.01,
```
cpignedoli, line 44:
```suggestion
            value=0.005,
```

cpignedoli (approved):
approved if you merge my new default value

mikibonacci (approved):
Hi @AndresOrtegaGuerrero, thank you for the PR. 
For me it can be merged, I just left a small comment (nothing relevant actually).

PR #834, opened 2024-10-01
Feature/handling molecules

AndresOrtegaGuerrero:
This feature handles #734 

- [x] Include the option molecule for StructureData editor
- [x] Band structure calculation should be disabled
- [x] Cell Optimizatio not available if structure is molecule
- [x] kpoint_distance should be disable and force [1,1,1]
- [x] Realoading an old workflow should restore normal behavior of app
- [x] PdosCalculation Test

Modification to molecule editor
molecule_editor

Cell option no available for Molecule
cell_opt_option_no_available

Kpoint_Distance disable for Molecules
kpoint_distance_disable_for_molecules

New Workflow shuold restore behavior 
new_workflow_or_reset_restore_normal_behavior

### DIFF
--- a/src/aiidalab_qe/app/configuration/__init__.py
+++ b/src/aiidalab_qe/app/configuration/__init__.py
@@ -118,11 +118,6 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     def _observe_previous_step_state(self, _change):
         self._update_state()
 
-    @tl.observe("input_structure")
-    def _observe_input_structure(self, _change):
-        if _change["new"] is not None and _change["new"].pbc == (False, False, False):
-            self.workchain_settings._on_input_structure_change(_change)
-
     def get_configuration_parameters(self):
         """Get the parameters of the configuration step."""
 
--- a/src/aiidalab_qe/app/configuration/__init__.py
+++ b/src/aiidalab_qe/app/configuration/__init__.py
@@ -113,6 +113,11 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     def _observe_previous_step_state(self, _change):
         self._update_state()
 
+    @tl.observe("input_structure")
+    def _observe_input_structure(self, _change):
+        if self.input_structure is not None and self.input_structure.pbc == (False, False, False): 
+            pass
+
     def get_configuration_parameters(self):
         """Get the parameters of the configuration step."""
 
@@ -156,7 +161,7 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     @tl.default("state")
     def _default_state(self):
         return self.State.INIT
-
+    
     def reset(self):
         """Reset the widgets in all settings to their initial states."""
         with self.hold_trait_notifications():
--- a/src/aiidalab_qe/app/configuration/__init__.py
+++ b/src/aiidalab_qe/app/configuration/__init__.py
@@ -45,6 +45,11 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
             (self.advanced_settings, "input_structure"),
         )
         #
+        ipw.dlink(
+            (self, "input_structure"),
+            (self.workchain_settings, "input_structure"),
+        )
+        #
         self.built_in_settings = [
             self.workchain_settings,
             self.advanced_settings,
--- a/src/aiidalab_qe/app/configuration/__init__.py
+++ b/src/aiidalab_qe/app/configuration/__init__.py
@@ -120,8 +120,8 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
 
     @tl.observe("input_structure")
     def _observe_input_structure(self, _change):
-        if self.input_structure is not None and self.input_structure.pbc == (False, False, False): 
-            pass
+        if _change["new"] is not None and _change["new"].pbc == (False, False, False):
+            self.workchain_settings._on_input_structure_change(_change)
 
     def get_configuration_parameters(self):
         """Get the parameters of the configuration step."""
@@ -166,7 +166,7 @@ class ConfigureQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
     @tl.default("state")
     def _default_state(self):
         return self.State.INIT
-    
+
     def reset(self):
         """Reset the widgets in all settings to their initial states."""
         with self.hold_trait_notifications():
--- a/src/aiidalab_qe/app/configuration/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced.py
@@ -110,7 +110,7 @@ class AdvancedSettings(Panel):
         ipw.dlink(
             (self.override, "value"),
             (self.kpoints_distance, "disabled"),
-            lambda override: not override,
+            lambda override: not override if self.input_structure is not None and self.input_structure.pbc != (False, False, False) else True,
         )
         self.kpoints_distance.observe(self._callback_value_set, "value")
 
@@ -332,6 +332,9 @@ class AdvancedSettings(Panel):
             self.hubbard_widget.update_widgets(change["new"])
             if isinstance(self.input_structure, HubbardStructureData):
                 self.override.value = True
+            if self.input_structure.pbc == (False, False, False):
+                self.kpoints_distance.value = 10.0
+                self.kpoints_distance.disabled = True
         else:
             self.magnetization.input_structure = None
             self.pseudo_setter.structure = None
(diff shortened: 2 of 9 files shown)


AndresOrtegaGuerrero:
For this PR We need #833 merged to get the PDOS peaks otherwise is all 0.0

AndresOrtegaGuerrero:
@edan-bainglass  I included you since it involves widgets and some behaviour of the app , so you can give your opinion for changes or things to consider

cpignedoli (approved):
tested in combination with #833 
Automatic selection of periodicity will have to follow as a separate PR. The user can then override the automatic selection.

src/aiidalab_qe/app/configuration/workflow.py
```diff
@@ -150,6 +154,54 @@ def update_reminder_info(change, name=name):
             **kwargs,
         )
 
+    @tl.observe("input_structure")
+    def _on_input_structure_change(self, change):
+        """Update the relax type options based on the input structure."""
```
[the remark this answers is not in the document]
edan-bainglass: Just giving you a heads up of what to expect in my beast PR, so no worries on changing it here 🙂

edan-bainglass (approved):
Okay, all good from me! Note that it will likely all change in my PR - not the behavior, but the implementation, due to the introduction of MVC.

One comment though - I find it odd that a control that modifies the app behavior (disable this, remove that) is somewhat "hidden" deep down (expand step, expand editors, switch tab...). I would consider elevating such controls. I am planning a layout redesign PR after my MVC PR. Should be a fun one!

src/aiidalab_qe/app/configuration/workflow.py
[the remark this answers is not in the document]
AndresOrtegaGuerrero: Yeah, I know! that part of removing was something that i wanted to check with you , so for molecules we cannot allow the user to do Cell Optimization , however in ipywidgets the ToogleButtons , there is no way to disable one option .. so I opted to go my way around and remove that option for molecules, while making sure it should be present for 1d,2d, and 3d systems.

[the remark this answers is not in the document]
edan-bainglass: Right, I've encountered the limitation before. Would be nice in general if any option-selection widget allowed disabling certain options. Oh well...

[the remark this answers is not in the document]
AndresOrtegaGuerrero: So , i would rename as 
@tl.observe("input_structure")
    def _update_relax_type(self, change):

or keep the name of _on_input_structure_change name , since i need the traitlets to trigger this function ?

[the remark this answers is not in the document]
edan-bainglass: Again, I wouldn't change it at this stage. But if you did...

```python
@tl.observe("input_structure")
def _on_input_structure_change(self, change):
    self._update_relax_type(change["new"])  # or `change` if you need more from it 

def _update_relax_type(self, input_structure):  # or change (see above)
    ...
```

Sometimes, the triggered handler (`_on_<state>_change`) might trigger several methods. This pattern makes it clear what does are. For example:

```
on structure change:
  do this()
  then do that()
  then finally do the last thing()
```

Easier to maintain. Also, `this`, `that`, etc. are effectively "business logic", so they are moved to the model. Then you get

```
on structure change:
  model.do_this()
  model.do_that()
  model.the_last_thing()
```

**Exception to the rule**: controllers can also modify the view directly. In such a case, the method would not go through the model.

----

Please don't let any of the above block your PR 🙏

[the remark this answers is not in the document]
AndresOrtegaGuerrero: ok i will keep the name , and i guess later we sit together for lazy loading :D

AndresOrtegaGuerrero:
I will wait for the input from @superstar54 or @PNOGillespie  , to know there is no conflict with the XAS or XPS plugins. or extra modifications so is 100% compatible

src/aiidalab_qe/app/configuration/workflow.py
```diff
@@ -141,6 +145,54 @@ def update_reminder_info(change, name=name):
...
+                ("Atomic positions", "positions"),
+                ("Full geometry", "positions_cell"),
+            ]
+            # Ensure the value is in the options
+            if self.relax_type.value not in [
+                option[1] for option in self.relax_type.options
+            ]:
+                self.relax_type.value = "positions_cell"
+
+            self.properties["bands"].run.disabled = False
```
superstar54, line 194: `elif` and `else` are duplicated.

src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -113,6 +118,11 @@ def __init__(self, **kwargs):
     def _observe_previous_step_state(self, _change):
         self._update_state()
 
+    @tl.observe("input_structure")
+    def _observe_input_structure(self, _change):
+        if _change["new"] is not None and _change["new"].pbc == (False, False, False):
+            self.workchain_settings._on_input_structure_change(_change)
+
```
superstar54, line 125: This is not needed because you already linked the `input_structure` to the `input_structure` of the `workchain_settings.`
```suggestion
```

src/aiidalab_qe/app/configuration/advanced.py
```diff
@@ -285,6 +281,20 @@ def __init__(self, default_protocol=None, **kwargs):
...
+            (self.kpoints_distance, "disabled"),
+            lambda override: not override,
+        )
+
+    def remove_kpoints_distance_link(self):
+        """Remove the kpoints_distance_link."""
+        if hasattr(self, "kpoints_distance_link"):
+            self.kpoints_distance_link.unlink()
+            del self.kpoints_distance_link
+
```
superstar54, line 297: I suggest not creating the link; instead, modify the `_override_changed` method. In this way, you only need to keep the logic in one place.

superstar54 (changes requested):
Hi @AndresOrtegaGuerrero , thanks for the nice work! I requested a few changes.

- For the kpoints, I think we need to use `gamma_only` for molecule instead of [1, 1, 1], because the former one is faster.



- XAS plugin does not support molecule calculation. I think @PNOGillespie could add a warning message in the XAS GUI if users activate the molecule option (in a future PR).
- XPS plugin supports molecule, and has a molecule option in the XPS GUI  explicitly, I can modify it in a future PR to use the molecule option from this PR. 

So, I think your PR can go ahead without worry XAS and XPS.

src/aiidalab_qe/app/configuration/advanced.py
AndresOrtegaGuerrero: I need this link to be created and removed , so that the kpoints distance is not modify for molecules , if not the override keep it blocked

AndresOrtegaGuerrero:
> Hi @AndresOrtegaGuerrero , thanks for the nice work! I requested a few changes.
> 
> * For the kpoints, I think we need to use `gamma_only` for molecule instead of [1, 1, 1], because the former one is faster.
> * XAS plugin does not support molecule calculation. I think @PNOGillespie could add a warning message in the XAS GUI if users activate the molecule option (in a future PR).
> * XPS plugin supports molecule, and has a molecule option in the XPS GUI  explicitly, I can modify it in a future PR to use the molecule option from this PR.
> 
> So, I think your PR can go ahead without worry XAS and XPS.

I would not use gamma_only , then there will be problems for vibroscopy app

src/aiidalab_qe/app/configuration/__init__.py
AndresOrtegaGuerrero: I need this trailet here , to remove the cell optimization option to be deleted

src/aiidalab_qe/app/configuration/advanced.py
AndresOrtegaGuerrero: since is liked to the reset()

AndresOrtegaGuerrero:
> > there will be problems for vibroscopy app
> 
> `gamma_only` is one time faster than [1, 1, 1] and is suggested to be used in a normal case. The plugin could have its logic to handle the molecule case, e.g., the vibroscopy app can override the `gamma_only` to [1, 1,1].
> 
> Another parameter is `assume_isolated`; it's also suggested to use `mt` for the molecular system.

The idea is to set this first and in a follow up set the MT

superstar54:
Ok to add MT in another PR.

btw, I can not find the `gamma_only` in the latest docs of pw, :laughing:

AndresOrtegaGuerrero:
> Ok to add MT in another PR.
> 
> btw, I can not find the `gamma_only` in the latest docs of pw, 😆

and the MT for the assume isolate ,
so it will be K_POINT{gamma}

Yest the best is to set that logic and how it affect other plugins in a follow up PR

src/aiidalab_qe/app/configuration/advanced.py
superstar54: I understand that the link is used to disable the `kpoints_distance` widget. You can achieve this in `_override_changed`. If input_structure is a molecule, the `kpoints_distance` widget is disabled no matter the change is.

AndresOrtegaGuerrero: but the issue, is when uploading a job with pbc that is not molecule , so I tried with the override ,but i realize that since is linked with the reset , it was better to create and remove the link ,

superstar54 (approved):
LGTM!

Issue #816, opened 2024-09-26
show warning messages if the calculation takes too long time


superstar54:
Is it possible to show warning messages e.g. if people select a combination of parameters that would make calculations too slow? 
Can defaults be changed if we are in the demo server? If so, Carlo and Giovanni will recommend different defaults (and possibly different messages to show in some cases) E.g., clean workdir, and default protocol = Fast, and ‘structure as is’. If we can, Carlo and Giovanni can suggest when to write, and what. [@Jason @Edan @Xing]. Add a warning (in step 3) if the calculation takes too long time, e.g. running 5 atoms in a localhost with 2 cpus.

PR #840, opened 2024-10-03
Features/warning long calc

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/816.

### DIFF
--- a/src/aiidalab_qe/app/submission/__init__.py
+++ b/src/aiidalab_qe/app/submission/__init__.py
@@ -51,7 +51,7 @@ class SubmitQeAppWorkChainStep(ipw.VBox, WizardAppWidgetStep):
 
     # Warn the user if they are trying to run calculations for a large
     # structure on localhost.
-    RUN_ON_LOCALHOST_NUM_SITES_WARN_THRESHOLD = 5
+    RUN_ON_LOCALHOST_NUM_SITES_WARN_THRESHOLD = 10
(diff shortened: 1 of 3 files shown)


mikibonacci:
Hi @AndresOrtegaGuerrero  and @superstar54 , the PR is ready for review. 

In the submission step, I added (completed what I already found there) checks for: number of sites (lower than 10 is safe), cell volume (lower then 1000 A^3 is safe) and number of cpus (lower than 4 is not safe for big systems - see previous checks and Figure 1 - higher than 1 in localhost ask for checks, see Figure 2)
Figure 1 (similar warning even if not in localhost - see the code) is:
Screenshot 2024-10-07 at 16 30 53

Figure 2 is:
Screenshot 2024-10-07 at 16 32 10

AndresOrtegaGuerrero:
you have some issues in the tests, @mikibonacci is it possible to create a test , to check this logic ?

mikibonacci:
Hi @AndresOrtegaGuerrero, yeah will put some test, should be possible

mikibonacci:
Hi @AndresOrtegaGuerrero , I updated the descriptions. For the tests: I added the testing for the localhost part, not for the non-localhost (I am not sure how to do it, as we should load an external code...). Let me know if you have some ideas on how to do it, or if we can go like this.

thanks!

AndresOrtegaGuerrero (approved):
LGTM! Thank you Miki

PR #842, opened 2024-10-07
update labels for fat_bands

AndresOrtegaGuerrero:
In #787 , 
I forgot to update the logic for the fat_bands , here is included
Now for only bands calculations and fat bands the labels are propertly display
image


mikibonacci (approved):
Hi @AndresOrtegaGuerrero , for me can be merged! We should keep in mind to have some tests in the future, but for now it is not so crucial. 

Thanks!

PR #844, opened 2024-10-08
Trigger the update of node view

superstar54:


src/aiidalab_qe/app/result/workchain_viewer.py
```diff
@@ -83,29 +84,36 @@ def toggle_camera():
...
-        self._update_view()
 
         super().__init__(
             children=[self.title, self.result_tabs],
             **kwargs,
         )
-        self._process_monitor = ProcessMonitor(
-            process=self.node,
+        self.process_monitor = ProcessMonitor(
+            timeout=1.0,
```
AndresOrtegaGuerrero, line 93: How this timeout works ?

superstar54: For the moment, the `ProcessMonitor` will call the callback functions every timeout seconds. This, indeed, will be a problem if the callback function is computationally expensive.

In principle, we only need to run the callback function when the process is finished (or even better, when the process is modified). I have opened an [issue](https://github.com/aiidalab/aiidalab-widgets-base/issues/636) in the aiidalab-widget-base.

We can wait for that issue to be solved before merging this PR.

AndresOrtegaGuerrero: or at least we should increase the time , 1 second is to often

superstar54: We use a small value for the timeout, so that the results can be immediately shown when the process is finished.

As pointed out by @danielhollas , there is a `on_sealed` argument for the callbacks. I changed the callback to `on_sealed`, so that the `_update_view` method will only call when the process is finished. In this case, the timeout 1 s should be fine.

Issue #813, opened 2024-09-26
Remove the job selection in the top


superstar54:
- remove completely the dropdown, just have a button to go to the full job list, 
- also remove the header of the App, i.e., the ‘previous/next step’ and ‘reset’ buttons, because people can just refresh the page.

Remove the two lines, (job selection, reset, previous.etc), and then add two buttons: 1) to the job lists, 2) [@Xing]
Add a clear big link or button, on the right of the job dropdown at the top, to jump to the “Job list” page. Call it “Full job list”.

PR #849, opened 2024-10-08
Fix/638/remove job selection

superstar54:
This PR fixs #813 .

- Removed the workchain selector on the top
- Remove the `previous`, `next` and `reset` header.

Then, I added
- A new `Job list` button on the top
- A button to create a new calculation. It will create a new QE App page, instead of reset the current APP.

https://github.com/user-attachments/assets/ef821b62-4119-4659-9dc5-bb1c25414315

### DIFF
--- a/src/aiidalab_qe/app/main.py
+++ b/src/aiidalab_qe/app/main.py
@@ -4,13 +4,12 @@ Authors: AiiDAlab team
 """
 
 import ipywidgets as ipw
+from IPython.display import Javascript, display
 
-from aiida.orm import load_node
 from aiidalab_qe.app.configuration import ConfigureQeAppWorkChainStep
 from aiidalab_qe.app.result import ViewQeAppWorkChainStatusAndResultsStep
 from aiidalab_qe.app.structure import StructureSelectionStep
 from aiidalab_qe.app.submission import SubmitQeAppWorkChainStep
-from aiidalab_qe.common import QeAppWorkChainSelector
 from aiidalab_widgets_base import WizardAppWidget, WizardAppWidgetStep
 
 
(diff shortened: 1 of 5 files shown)


src/aiidalab_qe/app/main.py
```diff
@@ -64,23 +68,26 @@ def __init__(self, qe_auto_setup=True):
...
+        # hide the header
+        self._wizard_app_widget.children[0].layout.display = "none"
         self._wizard_app_widget.observe(self._observe_selected_index, "selected_index")
 
-        # Add process selection header
-        self.work_chain_selector = QeAppWorkChainSelector(
-            layout=ipw.Layout(width="auto")
+        # Add a button to start a new calculation
+        self.new_work_chains_button = ipw.Button(
+            description="New calculation",
```
AndresOrtegaGuerrero, line 77: `Start New Calculation`
Might be more explicit for the user

src/aiidalab_qe/app/utils/search_jobs.py
```diff
@@ -70,6 +72,8 @@ def load_data(self):
         ]
 
     def setup_widgets(self):
+        import ipywidgets as ipw
```
AndresOrtegaGuerrero, line 75: why not to keep some of these import on top ? or somehow you expect better performance like this ?

AndresOrtegaGuerrero:
I think the button named "Job list" it might not be intuitive to the users, Maybe we should re label to something more explicit? 
"Previous Calculations" or "View Previous Runs" or "Job History" , so we dont need some text to guide the user where to find the calculations

src/aiidalab_qe/app/utils/search_jobs.py
superstar54: I only import the module when I use it—kind of lazy loading to speed the loading.

AndresOrtegaGuerrero:
I think the same apply to the button Guide , maybe we should use 

"How to Use" or "Usage Guide" or even "Getting Started" , and maybe we can change the icon to `icon="fa-book",`

src/aiidalab_qe/app/main.py
superstar54: Thanks. I changed it to `Start New Calculation`.

superstar54:
> I think the button named "Job list" it might not be intuitive to the users, Maybe we should re label to something more explicit?

Thanks for the suggestion. I changed the `Job list` to `Job History`. 

> "How to Use" or "Usage Guide" or even "Getting Started" , and maybe we can change the icon to icon="fa-book",

For the `Guide` button, I will let @edan-bainglass decide and make changes if needed.

AndresOrtegaGuerrero (approved):
LGTM! thank you Xing

edan-bainglass:
> I think the same apply to the button Guide , maybe we should use\n\n\"How to Use\" or \"Usage Guide\" or even \"Getting Started\" , and maybe we can change the icon to icon=\"fa-book\",

I like Getting Started, and the book icon might be a nice touch. Maybe check what others do along Getting Started.

AndresOrtegaGuerrero:
> > I think the same apply to the button Guide , maybe we should use\n\n"How to Use" or "Usage Guide" or even "Getting Started" , and maybe we can change the icon to icon="fa-book",
> 
> I like Getting Started, and the book icon might be a nice touch. Maybe check what others do along Getting Started.

Should we open an issue to replace it ? @superstar54

Issue #823, opened 2024-09-26
Add download buttons


superstar54:
In workflow summary, add 2 download buttons:
.aiida file
Verdi process dump (zipped)

AndresOrtegaGuerrero:
Should instead the download button be ,

` verdi archive create qe_app_calc.aiida --nodes {qe_node}`
 
but also include the Import QeAppNode to allow the users share results ?

`verdi archive import {upload node}.aiida`

mikibonacci:
We just need to make sure that they are importing `QeAppWorkChain`, otherwise they will not see it in the app... well, they can always just want to import structures and codes, or entire profiles

PR #850, opened 2024-10-08
Adding the download button for QeAppWorkChains.

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/823

We provide a DownloadDataWidget to download, separately:

- AiiDA archive: export_{node.pk}.aiida
- AiiDA raw data export_{node.pk}_raw.zip

both are zip files.


mikibonacci:
Hi @superstar54 and @AndresOrtegaGuerrero , this PR is ready for review. 

The main new thing is the `DownloadDataWidget`, in the new `utils` subfolder of the `result` one. I left mostly untouched the implementation of a download output button found in the `WorkChainOutputs`, as I have not fully understood how it works. I preferred to implement a widget which contains both archive and raw_data download buttons. 

Files are downloaded as zip archives. 

A part the implementation, which probably needs to be improved, only one issue: with google chrome the downloads are seen as insecure (I think because they are zip files, and we have http instead of https). 
Could you please confirm that also you have this issue?

Thanks!

AndresOrtegaGuerrero:
@mikibonacci thank you , do you think we can include the import aiida node , is too much for this PR ?

src/aiidalab_qe/app/result/utils/download_data.py
```diff
@@ -0,0 +1,113 @@
+import base64
```
AndresOrtegaGuerrero, line 1: Should we have this in common folder instead

mikibonacci:
Hi @AndresOrtegaGuerrero , thanks for looking at this PR. for the import button, I think I can do a second PR so we don't add too much to this one. But also, I am a bit concerned on what the user may try to upload... for example, maybe they expect that if a PwBaseWorkChain is imported, this will be visible in the app... but this is not always true (true only if we import the whole QeAppWorkChain). Actually, we should be safe enough, as the user typically will just export and import using the app buttons.

To summarize, I think we can do it in another PR, maybe not a critical one. We have already the upload widget from euphonic vibro-app, so should be easy to use similar code and just play with the import API of AiiDA

src/aiidalab_qe/app/result/utils/download_data.py
mikibonacci: what do you mean?

AndresOrtegaGuerrero: I mean of having this widget in the folder common , but i am not sure , i think having it in results is ok , is just it could be nice to have the widgets in a single folder.

src/aiidalab_qe/app/result/utils/download_data.py
```diff
@@ -0,0 +1,113 @@
...
+        super().__init__(
+            children=[
+                self.download_archive_button,
+                self.download_raw_button,
+            ],
+        )
+
+    def download_archive_data(self, _=None):
+        """
+        Download both the phonopy.yaml and fc.hdf5 files.
```
AndresOrtegaGuerrero, line 39: Maybe lets update this docstring

src/aiidalab_qe/app/result/utils/download_data.py
```diff
@@ -0,0 +1,113 @@
...
+        archive_data = self.produce_bitestream(self.node, what="archive")
+        self._download(payload=archive_data, filename=f"export_{self.node.pk}.aiida")
+        del archive_data
+        self.download_archive_button.description = (
+            self.download_archive_button.description.replace(
+                "... Downloading now...", ""
+            )
+        )
+
+    def download_raw_data(self, _=None):
```
AndresOrtegaGuerrero, line 51: these download methods might be be simplified into one ? where you can receive arguments for the desire download ?

src/aiidalab_qe/app/result/utils/download_data.py
mikibonacci: oh right, I totally forgot to do it, sorry

src/aiidalab_qe/app/result/utils/download_data.py
mikibonacci: yeah I think should not be difficult, I will do it

src/aiidalab_qe/app/result/utils/download_data.py
mikibonacci: yeah I see... maybe we can move it later, during the lazy loading refactoring merging

AndresOrtegaGuerrero:
> Hi @superstar54 and @AndresOrtegaGuerrero , this PR is ready for review.
> 
> The main new thing is the `DownloadDataWidget`, in the new `utils` subfolder of the `result` one. I left mostly untouched the implementation of a download output button found in the `WorkChainOutputs`, as I have not fully understood how it works. I preferred to implement a widget which contains both archive and raw_data download buttons.
> 
> Files are downloaded as zip archives.
> 
> A part the implementation, which probably needs to be improved, only one issue: with google chrome the downloads are seen as insecure (I think because they are zip files, and we have http instead of https). Could you please confirm that also you have this issue?
> 
> Thanks!

One question why the aiida file is saved as a zip?

src/aiidalab_qe/app/result/utils/download_data.py
```diff
@@ -0,0 +1,139 @@
...
+        self.download_archive_button = ipw.Button(
+            description="Download AiiDA archive zip data",
+            icon="download",
+            button_style="primary",
+            disabled=False,
+            layout=ipw.Layout(width="auto"),
+        )
+        self.download_archive_button.on_click(self._download_data)
+
+        self.download_raw_button = ipw.Button(
```
superstar54, line 19: It would be good to add a tooltip for both buttons. In the tooltip, explain what the `archive` and `raw` mean.

mikibonacci:
Hi @AndresOrtegaGuerrero , it is just the standard archive format when you create an archive (see this in [aiida-core create.py file](https://github.com/aiidateam/aiida-core/blob/main/src/aiida/tools/archive/create.py) at line 173. 
Maybe I can just put the button description to be "Download AiiDA archive data" and not "Download AiiDA archive zip data", as it may be a bit misleading (the user may try to unzip the archive file). 

What do you think?

src/aiidalab_qe/app/result/utils/download_data.py
```diff
@@ -0,0 +1,139 @@
...
+            children=[
+                self.download_archive_button,
+                self.download_raw_button,
+            ],
+        )
+
+    def _download_data(self, button_instance):
+        """
+        This method handles the download process when a download button is clicked.
+        It updates the button's description to indicate that the download is in progress,
```
superstar54, line 40: I would suggest not modifying the description but instead showing this information in another container, e.g., an HTML.

mikibonacci: Done

src/aiidalab_qe/app/result/utils/download_data.py
```diff
@@ -0,0 +1,160 @@
...
+        with a filename based on the node's primary key, and then resets the button description
+        to its original state.
+
+        Args:
+            button_instance (ipywidgets.Button): The button instance that was clicked.
+        """
+
+        if "archive" in button_instance.description:
+            what = "archive"
+            filename = f"export_{self.node.pk}.aiida"
```
superstar54, line 70: It's would be good to add `qeapp` to the name of the file, e.g.,
```suggestion
            filename = f"export_qeapp_calculation_pk_{self.node.pk}.aiida"
```
Same for the raw data

src/aiidalab_qe/app/result/workchain_viewer.py
```diff
@@ -199,7 +200,7 @@ def __init__(self, node, export_dir=None, **kwargs):
         super().__init__(
             children=[
                 ipw.HBox(
-                    children=[title, self._download_button_container],
+                    children=[self._download_button_widget, title],
```
superstar54, line 203: It's too compact to show the download widget and the title in an HBox. Please use `VBox`.

superstar54 (changes requested):
Hi @mikibonacci , thanks for the update. It's a good way to download it in another thread!

Maybe one bug: when download the `raw` data, it always in this status: `Downloading data...`

mikibonacci:
Hi @superstar54 , thanks for reviewing the PR. I addressed your valuable comments and suggestions.

However, I don't understand the bug you mention: are the data downloaded in your local machine but the "Downloading data..." still there?

superstar54:
In my test, tt's a small calculation, the `aidia` file is downloaded within 5 s. But for the `raw` data, after 10 mins, it still shows `Downloading data...`, and no data is not downloaded.

mikibonacci:
Hi @superstar54 , I think I solved the issue.  I just disable the raw data download if the ProcessDumper cannot be imported. I have not found a way to catch the error from the second thread (for now).

superstar54 (approved):
LGTM!

PR #851, opened 2024-10-09
changing guide button label and layout

AndresOrtegaGuerrero:
this PR modifies the description of the guide toggle button

image

### DIFF
--- a/src/aiidalab_qe/app/wrapper.py
+++ b/src/aiidalab_qe/app/wrapper.py
@@ -129,11 +129,12 @@ class AppWrapperView(ipw.VBox):
 
         self.guide_toggle = ipw.ToggleButton(
             button_style="",
-            icon="question",
+            icon="book",
             value=False,
-            description="Guide",
+            description="Getting Started",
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/app/wrapper.py
```diff
@@ -129,11 +129,12 @@ def __init__(self) -> None:
...
         self.guide_toggle = ipw.ToggleButton(
             button_style="",
-            icon="question",
+            icon="book",
             value=False,
-            description="Guide",
+            description="Getting Started",
             tooltip="Learn how to use the app",
             disabled=True,
+            layout=ipw.Layout(width="50%"),
```
superstar54, line 137: The width will depend on the parent container. Since the number of text is fixed, we maybe use an absolute value. e.g, 150px. Please try to determine which value is best.

AndresOrtegaGuerrero:
image
I use 140px on both so is more symmetrical

superstar54 (approved):
LGTM!

PR #856, opened 2024-10-10
allows fermi_energy to be zero

AndresOrtegaGuerrero:
This PR allows the widget to be use for vibroscopy plugin to set fermi to zero

### DIFF
--- a/src/aiidalab_qe/common/bandpdoswidget.py
+++ b/src/aiidalab_qe/common/bandpdoswidget.py
@@ -479,15 +479,15 @@ class BandPdosPlotly:
         # PDOS plot type
         if self.plot_type == "pdos":
             fermi_energy = self.fermi_energy.get(
-                "fermi_energy"
-            ) or self.fermi_energy.get("fermi_energy_up")
+                "fermi_energy", self.fermi_energy.get("fermi_energy_up")
+            )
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/common/bandpdoswidget.py
```diff
@@ -479,15 +479,15 @@ def handle_spin_polarization(fermi_energy, update_func, layout_type):
         # PDOS plot type
         if self.plot_type == "pdos":
             fermi_energy = self.fermi_energy.get(
-                "fermi_energy"
-            ) or self.fermi_energy.get("fermi_energy_up")
+                "fermi_energy", self.fermi_energy.get("fermi_energy_up")
```
mikibonacci, line 482: `self.fermi_energy.get("fermi_energy_up",0)` ? So we are sure that we get 0 fermi, in the worst case scenario

AndresOrtegaGuerrero: no , it should check in case there is one or the other , there is always one or the other

mikibonacci: Got it, thanks!

mikibonacci (approved):

mikibonacci:
Hi @AndresOrtegaGuerrero LGTM, thanks!

Issue #858, opened 2024-10-10
Improvement on process labels


mikibonacci:
`relax: none` -> `unrelaxed`. This is more intuitive, cannot be misunderstood.

This is a follow-up of https://github.com/aiidalab/aiidalab-qe/issues/811

PR #861, opened 2024-10-14
Optimizing process labels

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/858

The main change is the label for unrelaxed. From `relax: none` to `unrelaxed`. 

Examples of process labels: 

*No relax, magnetic*:
Screenshot from 2024-10-14 10-23-58

*relax*:
Screenshot from 2024-10-14 10-24-23

*No properties except vc-relax*:
Screenshot from 2024-10-14 10-24-46

As you can see from above, the structure formula is used by default for the labels. However, if the users attach a specific label to the structure in step 1, this is used instead. For example, here we attach the LABEL string:

Screenshot from 2024-10-14 10-32-03

Pinning @giovannipizzi and @cpignedoli for some feedback on this, thanks!


superstar54 (approved):
LGTM!

Issue #859, opened 2024-10-10
Improvement on warning for long/large calcs


mikibonacci:
We should also check the available local resources, to trigger the warning. But be careful, sometimes the whole resource of the machine are checked, not only the one available for the docker container.

This is a follow-up of https://github.com/aiidalab/aiidalab-qe/issues/816

PR #862, opened 2024-10-14
Improved warning messages

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/859

- refactoring of the warnings and suggestions
- comparison also with the cpus_count()

Three cases:
1 (2): run (not) on localhost with large systems (more than 10 atoms or volume larges than 1000 Angstroms) and asking than 4 CPUs:

Screenshot from 2024-10-14 14-11-24

3: run on localhost but asking more than half of the available CPUs
Screenshot from 2024-10-14 14-11-45

Here is the code: https://github.com/mikibonacci/aiidalab-qe/blob/e6d94e1ac485521d36873ce53311751b6d7643f5/src/aiidalab_qe/app/submission/__init__.py#L241

the suggestions are contained in the `suggestions` variable, and the three messages are constructed with respect the logic conditions just below.

Improvement on the messages and the conditions are more than welcome, @giovannipizzi and @cpignedoli !

PS: I will update the tests once we agree on the messages


src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -238,66 +241,72 @@ def _show_alert_message(self, message, alert_class="info"):
...
         on_localhost = (
             orm.load_node(self.pw_code.value).computer.hostname == "localhost"
         )
         num_sites = len(self.input_structure.sites)
         volume = self.input_structure.get_cell_volume()
+        localhost_cpus = len(os.sched_getaffinity(0))  # or os.cpu_count()
+
+        # List of possible suggestions for warnings:
+        suggestions = {
+            "more_resources": "<li>Increase the resources (CPUs should be equal or more than 4, if possible) </li>",
```
superstar54, line 256: `should be equal or more than 4` could mislead the user, and let the user think that 4 cpus is enough for a large system. Could you use a simple function to calculate the number of cpus based on the number of atoms and volume?
Assume the maximum number of computation time is ~ 5 days.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -238,66 +241,72 @@ def _show_alert_message(self, message, alert_class="info"):
...
+                self._show_alert_message(
+                    warnings_1_2
+                    + "<ul>"
+                    + suggestions["go_remote"]
+                    + suggestions["more_resources"]
+                    + suggestions["change_configuration"]
+                    + "</ul>",
+                    alert_class="warning",
+                )
+        elif on_localhost and num_cpus > np.ceil(localhost_cpus / 2):
```
superstar54, line 299: How about showing the warning message when `80%` cpus is used?

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -238,66 +241,72 @@ def _show_alert_message(self, message, alert_class="info"):
...
-            return  # No code selected, nothing to do.
+            return  # No code selected or no structure, so nothing to do.
 
         num_cpus = self.pw_code.num_cpus.value * self.pw_code.num_nodes.value
         on_localhost = (
             orm.load_node(self.pw_code.value).computer.hostname == "localhost"
         )
         num_sites = len(self.input_structure.sites)
         volume = self.input_structure.get_cell_volume()
+        localhost_cpus = len(os.sched_getaffinity(0))  # or os.cpu_count()
```
superstar54, line 252: @unkcpz Could you check if this gives a correct number of CPUs in a docker container?

superstar54 (changes requested):
Hi @mikibonacci , thanks for the work! I added a few change request.

src/aiidalab_qe/app/submission/__init__.py
mikibonacci: just as a reference: https://github.com/getsentry/self-hosted/issues/2126#issue-1696973715

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -238,70 +240,91 @@ def _show_alert_message(self, message, alert_class="info"):
...
-                "</ul>",
-                alert_class="warning",
-            )
-        elif (
-            self.input_structure
-            and on_localhost
-            and (
-                num_sites > self.RUN_ON_LOCALHOST_NUM_SITES_WARN_THRESHOLD
-                or volume > self.RUN_ON_LOCALHOST_VOLUME_WARN_THRESHOLD
+                f"or a significant cell volume ({int(volume)} Å<sup>3</sup>), "
```
superstar54, line 280: suggest to:
`The selected structure is large, with {num_sites} atoms and a volume of {int(volume)} Å<sup>3</sup>, making it...`

superstar54 (approved):
LGTM! Two small change requests.

PR #868, opened 2024-10-21
Add button to trigger the update of the results tabs.

superstar54:
In PR #844 , the process monitor updates the result tabs in another thread. This leads to an issue: when the user interacts with the result, the widget runs in the thread differently from the one running in the process monitor.

Therefore, triggering the automatic updating of the result tabs is not possible.

In this PR, we don't use process monitor anymore. Instead, we add a button to let the user trigger the update manually. 

Screenshot from 2024-10-21 13-04-52

I also modified the layout of the buttons (kill, clean) to make the UI pretty.

### DIFF
--- a/src/aiidalab_qe/app/result/__init__.py
+++ b/src/aiidalab_qe/app/result/__init__.py
@@ -51,8 +51,8 @@ class ViewQeAppWorkChainStatusAndResultsStep(ipw.VBox, WizardAppWidgetStep):
             description="Kill workchain",
             tooltip="Kill the below workchain.",
             button_style="danger",
-            icon="window-close",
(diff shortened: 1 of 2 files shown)


mikibonacci (approved):
Hi @superstar54 , thanks for this and for improving the other buttons. 

Just thinking if we can somehow avoid to render again the tabs if they are already there, but I also think that this will be not a problem when the lazy loading will be there.

So, LGTM!

superstar54:
>  Just thinking if we can somehow avoid to render again the tabs if they are already there, but I also think that this will be not a problem when the lazy loading will be there.

Indeed, this is the problem from the node tree. There is a PR to fix it: https://github.com/aiidalab/aiidalab-qe/pull/560. Will reopen it soon.

Issue #874, opened 2024-10-21
Issue when installing new plugins in new version


superstar54:
![Image](https://github.com/user-attachments/assets/696c214d-8f5b-49af-97ae-a5bb8dc389fd)

mikibonacci:
I think I know the issue. I renamed the titles, but these were used to check the plugins and import them. I will fix it

PR #880, opened 2024-10-22
Fixing the bug in the plugins installation

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/874

Hi @superstar54,
I added `package_name` so that we can define a new and descriptive title for each plugin, without having an issue in the plugin list page. 

Please test it also yourself so we are sure it works, thanks!


superstar54 (changes requested):
Hi @mikibonacci , the `key` better to be unique and without space. I suggest using the package's name as the `top-level-key`, and adding another `title` field, e.g.:
```yaml
aiidalab-qe-vibroscopy:
    title: Phonons and IR/Raman (aiidalab-qe-vibroscopy)
    description: Plugin to compute phonons, IR/Raman spectra, Inelastic Neutron Scattering of materials
    author: Miki Bonacci, Andres Ortega Guerrero, Lorenzo Bastonero and Nicola Marzari
    pip: aiidalab-qe-vibroscopy==1.1.1
    post_install: setup-phonopy
```

superstar54 (approved):
LGTM!

Issue #875, opened 2024-10-21
Improve the bands plot


superstar54:
- can we remove the splines on the bands, that make bands weird when there are discontinuities? (See attachment)
- can we increase the max value of the slider of flat bands, when one is looking to the whole large band structure (here LiO)?

![image](https://github.com/user-attachments/assets/92c8851b-a9bb-45c4-8bce-648c0633f0fe)

![image](https://github.com/user-attachments/assets/1da05b3b-4245-42a3-8110-a6afebbdc8d2)

AndresOrtegaGuerrero:
Could you add the attachment?

PR #883, opened 2024-10-22
increasing fat_bands_slide/ removing spline

AndresOrtegaGuerrero:
This PR address #875 

We are increasing the max value of fat bands to 2.0
We are removing the spline option
We additionally are switing to plotly.graph_objects.Scattergl since is lighter and it shows a better performance response ( > 80 atoms)


superstar54:
Hi Andres, thanks for the PR.

I noticed there is a little difference at the bottom of the bands plot:
old one:
Screenshot from 2024-10-23 10-15-10
new one
Screenshot from 2024-10-23 10-16-29

Is it acceptable that there is not lines there?

AndresOrtegaGuerrero:
@superstar54 Yeah I noticed this , emm it might be due to using Scattergl , for me it is , actually it might less confusing that it just shows a bar, i check and the behaviour of the limiting bar still does the same work. Since i am observing way better performance for big system , this is acceptable for me. Also because in the past some people were confused that the display of the lines so i see it as a win/win :D

superstar54 (approved):
Thanks! LGTM!

PR #885, opened 2024-10-22
Update plugins.yaml

mikibonacci:


plugins.yaml
```diff
@@ -11,7 +11,7 @@ aiidalab-qe-vibroscopy:
     title: Phonons and IR/Raman (aiidalab-qe-vibroscopy)
     description: Plugin to compute phonons, IR/Raman spectra, Inelastic Neutron Scattering of materials via finite displacement and finite field approach.
     author: Miki Bonacci, Andres Ortega Guerrero, Lorenzo Bastonero and Nicola Marzari
-    pip: aiidalab-qe-vibroscopy==1.1.2
+    pip: aiidalab-qe-vibroscopy==1.1.3
```
superstar54, line 14: In this way, we allow any patch version within 1.1.x (e.g., 1.1.4).
```suggestion
    pip: aiidalab-qe-vibroscopy~=1.1.3
```

superstar54 (approved):
LGTM! I suggest using `~`, so that it allows patch updates within 1.1.x series

plugins.yaml
mikibonacci: In the rush I ignored this, but actually is the way to go, Thanks!

PR #904, opened 2024-10-24
Layout axes remain the same after update_plot methods is called 

AndresOrtegaGuerrero:
This address this issue [#903](https://github.com/aiidalab/aiidalab-qe/issues/903)


superstar54:
Hi @AndresOrtegaGuerrero , should we reset the axis range every time the plot updates? Maybe in some cases, when we change the plot setting, the axis range should be updated.

Another issue (not related to this PR), when I change the flat parameters, the screen flash a lot:


https://github.com/user-attachments/assets/a5909271-2f4a-4623-9fac-677e138702e7

In the flat band case, is it possible to update the data instead of re-drawing the plot?

AndresOrtegaGuerrero:
> Hi @AndresOrtegaGuerrero , should we reset the axis range every time the plot updates? Maybe in some cases, when we change the plot setting, the axis range should be updated.
> 
> Another issue (not related to this PR), when I change the flat parameters, the screen flash a lot:
> 
>  qeapp-flat-bands-falsh.mp4 
> In the flat band case, is it possible to update the data instead of re-drawing the plot?

To answer the first question , we have the reset axis button from plotly that does this function.

The second question , we do a replot since we have to widget a BandsPdosWidget (meant for qe app) and the BandsPdosPlotly , that kinda resemble the bandsplot-widget from the past that just receives a dictionary with the info. 

For a new thinckness i need to re-loop and re compute contribution per band...

superstar54:
> For a new thinckness i need to re-loop and re compute contribution per band.

If the number of data is the same, then it is possible to only update the data instead of re-plot the image. I did this in the XPS plot, as shown here: https://github.com/aiidalab/aiidalab-qe/blob/15f46b03e4b1367c6270b6f0db71068960900c06/src/aiidalab_qe/plugins/xps/result.py#L219

You can look at it in the future when you have time.

superstar54 (approved):

AndresOrtegaGuerrero:
@superstar54 Thank you! , regarding the flashing , what about adding like a debounce of 0.5 sec (or less) ? 

Regarding the updating the plot , so i will try to find a way to avoid repeating a re-loops. Maybe in a different PR ?

superstar54:
Hi @AndresOrtegaGuerrero , yes, Okay to merge the PR.  You can implement the update in a future PR when you have time.

Issue #909, opened 2024-10-31
Problems when importing the generated aiida node from the app


AndresOrtegaGuerrero:
When using the node generated download button , and importing it into another account it shows this bug 

~/apps/quantum-espresso/src/aiidalab_qe/app/result/workchain_viewer.py in __init__(self, node, **kwargs)
     49             """
     50         )
---> 51         self.workflows_summary = SummaryView(node)
     52 
     53         self.summary_tab = ipw.VBox(children=[self.workflows_summary])

~/apps/quantum-espresso/src/aiidalab_qe/app/result/summary_viewer.py in __init__(self, wc_node, **kwargs)
    225 class SummaryView(ipw.VBox):
    226     def __init__(self, wc_node, **kwargs):
--> 227         self.report = generate_report_parameters(wc_node)
    228         self.report_html = _generate_report_html(self.report)
    229 

~/apps/quantum-espresso/src/aiidalab_qe/app/result/summary_viewer.py in generate_report_parameters(qeapp_wc)
    106     energy_cutoff_rho = pw_parameters["SYSTEM"]["ecutrho"]
    107     occupation = pw_parameters["SYSTEM"]["occupations"]
--> 108     scf_kpoints_distance = qeapp_wc.inputs.relax.base.kpoints_distance.value
    109     report.update(
    110         {

/opt/conda/lib/python3.9/site-packages/aiida/orm/nodes/node.py in __getattr__(self, name)
    728             return getattr(self.base.links, new_name)
    729 
--> 730         raise AttributeError(name)

AttributeError: value

I originally checked that in the exported workchain qeapp_wc.inputs.relax.base.kpoints_distance is a Float, 
However then doing the import kpoints_distance becomes a Data 

I checked this in AiiDA 2.5.1

AndresOrtegaGuerrero:
@mikibonacci Is this related to the version of AiiDA ?

AndresOrtegaGuerrero:
The solution i found, for imported files is then to use 

iqeapp_wc.inputs.relax.base.kpoints_distance.base.attributes.get('value')

Issue #549, opened 2023-11-08
Remove Hardcoded Summary from Bands and Pdos


AndresOrtegaGuerrero:
The summary report of the workflow is hardcoded for bands and pdos here
`    if "pdos" in qeapp_wc.inputs:
        report[
            "nscf_kpoints_distance"
        ] = qeapp_wc.inputs.pdos.nscf.kpoints_distance.value`

PR #910, opened 2024-10-31
using ui_parameters to set the report

AndresOrtegaGuerrero:
This solves #909 and makes the app compatible to upload imported nodes from another account using QeApp 

It also solves #549


superstar54 (approved):
LGTM!

PR #917, opened 2024-11-07
Use try/except to catch error

### LINKED ISSUE #916: Failed to install QE when installing the app form the App Store.
In the latest aiidalab-ful-stack image (aiida-core 2.6.2), it failed to install QE when installing the app form the App Store.

Even when I clicked the "install codes", it still failed.

superstar54:
Fix #916 

The error from 916 is because the setup process stopped because the code failed to inspect the qe env. In this PR, we handle this gracefully so that it returns False when failed to inspect qe env.


superstar54:
Very funny thing: the `pre-commit` and `ruff` conflict with each other.

src/aiidalab_qe/setup/codes.py
```diff
@@ -44,18 +45,48 @@ def get_qe_env():
...
+        # Check if 'qe' is listed in the environment
+        for package in info:
+            if package.get("name") == "qe":
+                return True
+        return False  # noqa: TRY300
 
-    return env_exist and "qe" == info["name"]
+    except subprocess.CalledProcessError:
+        # Handle cases where the conda list command fails
+        return False
```
unkcpz, line 86: This may not correct, if the function return from exception, the exception will not propagate up.
The exception was handled by where the function `_install` who is calling `qe_installed()`. 

The change here will shadow docker build to show the exception message when it is failed.

unkcpz: At 2 pm?

superstar54:
As suggested by @unkcpz ,  if the cmd to inspect the env fails, we should raise the error and stop the installation.

src/aiidalab_qe/setup/codes.py
```diff
@@ -44,18 +45,45 @@ def get_qe_env():
...
-    return env_exist and "qe" == info["name"]
+        # Check if 'qe' is listed in the environment
+        for package in info:
+            if package.get("name") == "qe":
+                return True
+        return False  # noqa: TRY300
+    except Exception as error:
+        raise RuntimeError(
+            "Failed to check if Quantum Espresso is installed."
+        ) from error
```
unkcpz, line 86:
```suggestion
    except Exception as exc:
        raise RuntimeError(
            "Failed to check if Quantum Espresso is installed."
        ) from exc
   else:
       return False  # noqa: TRY300
```

Also for the catched exception, I think there are two types of exceptions we know which are `SubprocessCallError` and `JsonParseError`. You can catch those two and for uncatched exceptions, use raise to propagated it up.

unkcpz: For TRY300, you can move it to else block, https://docs.astral.sh/ruff/rules/try-consider-else/

unkcpz:
> Very funny thing: the pre-commit and ruff conflict with each other.

What you mean `pre-commit` and `ruff` are conflict? We use `ruff` in `pre-commit` as linter.

unkcpz:
> I added code required by ruff, in this [commit](https://github.com/aiidalab/aiidalab-qe/pull/917/commits/2e98568a73db876f45ca76a6acdfe0c819c3bcf7), but pre-commit deleted it [later](https://github.com/aiidalab/aiidalab-qe/pull/917/commits/bf06b6d6c5cc9b89a4b4ad1b682f06f9c1285373). Then ruff check failed.

Because the else was supposed to add to try..except. If is not correct to early return with an else in the try block.

unkcpz (changes requested):

edan-bainglass (approved):
I guess LGTM? Maybe check with @unkcpz.

superstar54:
Hi @unkcpz I think the current version works, and we can catch the exception in future work.

superstar54:
An issue is opened for further improvement: https://github.com/aiidalab/aiidalab-qe/issues/1092

unkcpz:
I remember I did discuss with @superstar54 offline. I thought this one is already merged~ So all good.

PR #921, opened 2024-11-09
Store already added node viewer and load it when needed

superstar54:
Reopen PR #560 .
In aiidalab-widgets-base/viewer.py, it creates a new `viewer` every time we select a new AiiDA node, and the viewer is not saved in the app, so we can not reuse it later. This has two drawbacks:
- The performance is back. user needs to wait every time when selecting a new node.
- One can not call the viewer later, e.g., for `reset`.

This PR stores the already added viewer and reuses it later.

This is also needed in the `lazy-loading` PR.


edan-bainglass:
@superstar54 thank you for reopening this! It will greatly improve the usefulness of the lazy-loading work implemented in #802 w.r.t the results section. Will review ASAP!

edan-bainglass (approved):
LGTM! As stated, this will come in handy when `node_view`s become MVCs. Good to not have to recreate them 🙂

PR #924, opened 2024-11-11
Restore update results button functionality

edan-bainglass:
The logic of the update results button was missed in #921. This PR restores the functionality by adding a `refresh` parameter to the `_update_node_view` method to allow bypassing the node view cache.

Note: a follow-up PR is planned to rethink the way we deal with the `WorkChainViewer` tabs which will render the update results button redundant.


superstar54 (approved):
LGTM! thanks for the fix!

PR #940, opened 2024-11-21
set condition for update label if SOC if False

AndresOrtegaGuerrero:
This PR addresses a bug related to the updating of labels for SOC (Spin-Orbit Coupling) calculations. The logic has been refined to ensure that the update occurs only when the SOC calculation parameter is explicitly set to False.
It solves , [#944](https://github.com/aiidalab/aiidalab-qe/issues/944)


edan-bainglass (approved):
Apologies for the delay @AndresOrtegaGuerrero. LGTM! Thanks for the fix 🙂 

One general comment is the use of "band" instead of "bands", which we do this in many places, e.g. `BandPdosPlotly`. I suggest we switch to "bands" across the board, if possible, unless we speak of a single band. TBD 🙂

PR #945, opened 2024-11-26
Add layer of abstraction in `Panel` classes

edan-bainglass:
This PR adds a layer of abstraction for settings panels though the new `ConfigurationSettingsPanel` and `ResourceSettingsPanel` classes and their respective `Model` classes. The rest of the codebase has been renamed for clarity.

PR #939, opened 2024-11-21
Resource setting tabs

superstar54:
- Find all codes in the plugins
- Only list one code for each code kind
- Each plugin has a code panel
- Modification in basic will update all plugins
- Add a override box in each plugin panel
- Warning and blocker message
- Basic panel only show the Base QEcode class, and plugin decide how to update their custom code widget

PR #946, opened 2024-11-26
Add missing features in resource settings tabs

edan-bainglass:
Based on #945

This PR looks to clean up the recently implemented plugin-based tab system in the resources/submission step (#939).
The idea is to leverage the newly developed abstraction in #945 to encapsulate the behavior of ANY resource settings panel, leaving the global resource settings panel to override based on its unique added features.

A few added items addressed in this PR:

1. PW code parallelization widget syncing between tabs
2. Support for syncing newly setup codes
3. Warning notification for overridden plugin resources
4. Syncing back with global resources on un-override

Remaining issues:

The new approach somehow causes three tests, which pass in isolation, to fail when running the full suite.

```python
FAILED tests/test_codes.py::test_check_submission_blockers - assert 0 == 1
 +  where 0 = len([])
 +    where [] = <aiidalab_qe.app.submission.model.SubmissionStepModel object at 0x7fa29d55edc0>.internal_submission_blockers
FAILED tests/test_codes.py::test_qeapp_computational_resources_widget - AssertionError: assert 'none' == 'block'
  - block
  + none
FAILED tests/test_submit_qe_workchain.py::test_create_builder_default - AssertionError: FILES DIFFER:
```

The last one I suspect may simply require a regeneration of the stored references.

But the first two, I suspect here it is a more serious issue that ONLY shows up in testing. I think the issue is that in testing, we generate more than a single instance of certain classes that have `dict` and/or `list` caches. We define these as **class variables**, which is "fine" for the app, but raises issues when various tests create instances of the class. A quick solution (if indeed this is the issue) is to define these as **instance variables**. However, this is not straight forward for the `codes` class variable shared by instances of `ResourceSettingsModel`s.

There is no guarantee that this is the issue. Just a suspicion. Will test further as soon as I am not sick 🤒 

@superstar54 good if you take a look, see if I missed something obvious. We can chat when I'm better.

## Update

My suspicions were correct. For the last test regarding the differing files, the issue is that part of implementing synchronization of newly setup codes required adding code options to the returned dictionary from `CodeModel.get_model_state`. This method, however, is also used when constructing builder parameters. Though it does not harm the builder process, it does affect the regression test part of `test_create_builder_default`. Simply regenerating the cached file does not help, as code options include UUIDs, which change every time the test runs. So I added a `remove_code_options` method to the test to discard the option keys, which resolves the problem.

Regarding the other two failing tests, indeed the issue here was the use of dictionary class variables. Both `plugin_mapping` and `codes` dictionary class variables are now changed to instance variables to avoid sharing state across class instances. Again, this is not an issue for the app, as there is only one instance per class. But this is not true for testing.


edan-bainglass:
@superstar54 ready for review 🙏

edan-bainglass:
> @superstar54 ready for review 🙏

Pleeeeeeeeeeeeeeease @superstar54 🙏

superstar54:
> @superstar54 ready for review 🙏

Hi @edan-bainglass , my apologies for overlooking your comment earlier. I’m reviewing it now. Thanks for your patience! 🙏

superstar54:
One more missing feature: the panel shows a wrong warning message when an HPC is selected. This is probably because it checks the resource before the `model.selected` gets updated, so the app thinks the localhost is used.

In the screenshot, I selected the merlin hpc with 44 cpus for each node, but it give a warning that it only has 28 cpus (which is the number of my localhost).

Screenshot from 2024-12-05 13-04-59

src/aiidalab_qe/common/panel.py
```diff
@@ -209,11 +204,12 @@ def _reset(self):
...
 
 
-class ResourceSettingsModel(SettingsModel):
+class ResourceSettingsModel(SettingsModel, HasModels[CodeModel]):
     """Base model for plugin code setting models."""
 
-    dependencies = ["global.global_codes"]
-    codes = {}  # To be defined by subclasses
+    dependencies = [
+        "global.global_codes",
```
superstar54, line 211: One concern: the base class hardcodes a dependency on "global.global_codes," where "global" is an identifier defined in the subclass or the context where the class is used.

src/aiidalab_qe/common/panel.py
```diff
@@ -228,38 +224,66 @@ def __init__(self, *args, **kwargs):
...
 
     def get_model_state(self):
-        codes = {name: model.get_model_state() for name, model in self.codes.items()}
         return {
-            "codes": codes,
-            "override": self.override,
+            "codes": {
+                identifier: code_model.get_model_state()
+                for identifier, code_model in self.get_models()
+            },
```
superstar54, line 255: we need to save the `override` value so that it shows correctly when loading a previous calculation.
```suggestion
            },
            "override": self.override,
```

tests/test_submit_qe_workchain/test_create_builder_default.yml
```diff
@@ -87,7 +86,6 @@ codes:
         nodes: 1
         ntasks_per_node: 1
         parallelization: {}
-    override: false
```
superstar54, line 90: This is needed. See the comment on the `override` value in the `get_model_state` method.

superstar54 (changes requested):
Hi @edan-bainglass , thanks for the nice work. 

One issue when loading the previous calculations, the code setting and the override box are not loaded correctly.

src/aiidalab_qe/common/panel.py
edan-bainglass: Yeah, this is not great. It appears to be non-breaking, as linking a model's trait to its own same trait does not raise an error and also doesn't do anything. However, it is still a thorn in principal.

Two ways to decouple:
1. Make the global panel unique. I don't like this - regardless of its special treatment, it IS conceptually a resources panel
2. Link the dependency explicitly during plugin discovery (in `_fetch_plugin_resource_settings`) rather than via the `_link_model` method:
```python
global_model = self._model.get_model("global")
... # then later in the for loop...
self._model.add_model(identifier, model)
tl.dlink(
    (global_model, "global_codes"),
    (model, "global_codes"),
)
```
Comments:
- This is the only thing `_link_model` is doing, so by applying the above workaround, we render the method unnecessary
- Other future dependencies between the plugin panels and the global panel may require similar workarounds
---
I'm honestly not too bothered by it. There are more pros than this one conceptual con. But I am open to discussion on the matter.

edan-bainglass:
> One issue when loading the previous calculations, the code setting and the override box are not loaded correctly.

Thanks @superstar54 for the review. Indeed, I discovered this issue yesterday morning but didn't have much time to think about it. I tried first to understand how `set_model_state` works in resource models. I had some questions, but maybe it is best we discuss your implementation in person next week. For now, I'll try to adjust for your other comments, then play with this issue a bit more to see if I can resolve it already 🙂

edan-bainglass:
> One more missing feature: the panel shows a wrong warning message when an HPC is selected. This is probably because it checks the resource before the `model.selected` gets updated, so the app thinks the localhost is used.
> 
> In the screenshot, I selected the merlin hpc with 44 cpus for each node, but it give a warning that it only has 28 cpus (which is the number of my localhost).

Right, indeed, it appears to be so that the check is done prior to the update of `model.selected`. But I'm having difficulty understanding why this is so. In the global settings panel, we do
```python
    def _render_code_widget(
        self,
        code_model: CodeModel,
        code_widget: QEAppComputationalResourcesWidget,
    ):
        super()._render_code_widget(code_model, code_widget)
        if code_model.default_calc_job_plugin == "quantumespresso.pw":
            code_model.observe(
                self._on_pw_code_resource_change,
                [
                    "num_cpus",
                    "num_nodes",
                    "ntasks_per_node",
                    "cpus_per_task",
                    "max_wallclock_seconds",
                ],
            )
```
where in `super()._render_code_widget(...)`, we set up a two-way link between the dropdown and `selected`. This means in the observation order, `selected` should sync prior to triggering the `_on_pw_code_resource_change` handler. But this is not the case. Not sure why at the moment.

edan-bainglass:
@superstar54 could you please have a look at my comments? I think this is closer to a solution, but we need to see if it sufficient. My main concern at the moment is the mistimed firing of `check_resources`. I'm still uncertain as to how this is happening.

Note that some part of the code is only there to handle changes in options occurring on the widget side. This happens when one introduces a new code using the "new code setup" button. However, we discussed removing this button in favor of a dedicated notebook for creating new codes. If and when this gets merged, some code introduced in this PR will be removed.

src/aiidalab_qe/common/panel.py
superstar54: Another solution is to keep the dependencies empty in the `ResourceSettingsModel` and create another `PluginResourceSettingsModel` for the plugin, in which we add this `global.global_codes`; in this way, we follow the oop and avoid making `_fetch_plugin_resource_settings` special.

superstar54:
Hi @edan-bainglass , thank you for the update. Regarding the dependencies issue, I suggest creating a `PluginResourceSettingsModel` class. For the `check_resources` issue, I don't know the reason either, we need more debugging to find the details.

src/aiidalab_qe/common/panel.py
edan-bainglass: Very nice 💪

edan-bainglass: Implemented in latest push!

edan-bainglass:
> Hi @edan-bainglass , thank you for the update. Regarding the dependencies issue, I suggest creating a `PluginResourceSettingsModel` class. For the `check_resources` issue, I don't know the reason either, we need more debugging to find the details.

@superstar54 I found out the issue with `check_resources` misfiring. The observations I set up in the app only come after those set up by the `QEAppComputationalResourcesWidget`. There, the selector's `value` is observed and triggers a change in `num_cpus`, which in turn also syncs up `ntasks_per_node`. As these fire prior to the my links, we see `num_cpus` change before the selector's `value` synchronizes with the model's `selected`. 

I'm afraid there is nothing to be done here short of redesigning `QEAppComputationalResourcesWidget`, which would likely be a minor change.

superstar54 (approved):
Thanks for the nice work! LGTM!

PR #949, opened 2024-11-28
Chaning pdos control labels

AndresOrtegaGuerrero:
image


cpignedoli (approved):
In a separate PR we should add grouping by kinds

PR #956, opened 2024-11-28
modifications to the EditStructureData Widgets

AndresOrtegaGuerrero:
image


cpignedoli (approved):
I would approve this state and leave to additional PRs refinements. Reorganizing the content of the three tabs will require a more in depth action.

PR #933, opened 2024-11-20
changing the labels of periodicity widget

AndresOrtegaGuerrero:
This PR changes the labels of the periodicity options widget as the following 

image

PR #960, opened 2024-11-29
setting periodicity options to the mapping

AndresOrtegaGuerrero:
In #933 , by mistake i also modify the logic in the update method causing that the periodicity was not changed. 

This fix this bug


cpignedoli (approved):
this is correct

PR #965, opened 2024-12-02
upgrade qe-plugin , include nbands_factor to PdosWorkChain

AndresOrtegaGuerrero:
In this PR , we solve the issue of the PDOS and BandStructure not matching the energies. 

newplot (1)

Now the PDOS and BANDS match

### DIFF
--- a/src/aiidalab_qe/plugins/pdos/workchain.py
+++ b/src/aiidalab_qe/plugins/pdos/workchain.py
@@ -4,6 +4,7 @@ from aiida_quantumespresso.common.types import ElectronicType, SpinType
 from aiidalab_qe.plugins.utils import set_component_resources
 
 PdosWorkChain = WorkflowFactory("quantumespresso.pdos")
+PwBandsWorkChain = WorkflowFactory("quantumespresso.pw.bands")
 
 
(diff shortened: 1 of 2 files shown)


cpignedoli (approved):
ok from my side

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -86,6 +87,8 @@ def get_builder(codes, structure, parameters, **kwargs):
             overrides=overrides,
             **kwargs,
         )
+        # include nbands_factor (Same as in BandsWorkChain)
+        pdos["nbands_factor"] = orm.Float(3.0)
```
superstar54, line 91: Suggest to read the value directly from the `PwBandsWorkChain` so that the value always keeps updated with the bands.
```suggestion
        pdos["nbands_factor"] = orm.Float(PwBandsWorkChain.get_protocol_inputs()['nbands_factor'])
```

superstar54 (changes requested):

superstar54 (approved):
Thanks for the work! LGTM!

PR #967, opened 2024-12-09
Fix process-node-related bugs

edan-bainglass:
Based on #946

This PR handles a couple of bugs related to the process node, specifically loading node views and managing the process state.

### DIFF
--- a/src/aiidalab_qe/app/result/__init__.py
+++ b/src/aiidalab_qe/app/result/__init__.py
@@ -189,6 +189,8 @@ class ViewQeAppWorkChainStatusAndResultsStep(ipw.VBox, WizardAppWidgetStep):
             self.node_view_container.children = [self.node_view_loading_message]
             self.node_view = self._create_workchain_viewer(node)
             self.node_views[node.uuid] = self.node_view
+        else:
+            return
 
         self.node_view_container.children = [self.node_view]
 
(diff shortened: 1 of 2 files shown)


edan-bainglass:
@superstar54 rebased. Please review once tests finish 🙏

src/aiidalab_qe/app/result/__init__.py
```diff
@@ -189,6 +189,8 @@ def _update_node_view(self, nodes, refresh=False):
             self.node_view_container.children = [self.node_view_loading_message]
             self.node_view = self._create_workchain_viewer(node)
             self.node_views[node.uuid] = self.node_view
+        else:
```
superstar54, line 192: Could you add a comment on what this case means and why return?

edan-bainglass: This guard was intended to cover cases where you select a non-`QeAppWorkChain` `WorkChainNode`. We decided not to change the rendered view in this case. However, I think it is best instead to render a "No viewer available for this node" message until one is developed (if we decide to do so). This is now implemented.

superstar54: Thanks! Could you add it as coment lines in the code?

edan-bainglass: The current implementation is now self-documenting, no? If you're not registered in the cache AND ARE a `WorkChainNode` BUT ARE NOT the `QeAppWorkChain`, then you don't have a viewer. A comment would be redundant at this point.

superstar54: I didn't notice you pushed a new commit to replace the `return` with a message. Now, it's clear in the code.

superstar54 (approved):
LGTM!

Issue #937, opened 2024-11-21
In step 3 warnings, number of CPUs should be estimated from Computer setup


mikibonacci:
We should rely on the default number of CPUs, as the python os.get_cpus (or similar) will not correctly work on deployments where host and single instance have different number of dedicated cores (which is almost always the case, the single instance will always have only a fraction of the host resources)

edan-bainglass:
Hi @mikibonacci. Is this yet to be implemented? It is something additional to your recently implemented warning system?

mikibonacci:
Hi @edan-bainglass, yes, it needs to be implemented. Is a fixing to the warning system. The issue is that now we estimate the available cores using python, but this detects the total cores of the host system, not the one assigned to the container (this is an issue mainly in kubernetes clusters, e.g. demo server). We need to get the number of available cpus from the code setup parameters, defined when the qe default codes are installed (if I remember correctly, @unkcpz told me that this is done when the container is generated, right?). 

I still need to check this, should happen this week

unkcpz:
> if I remember correctly, @unkcpz told me that this is done when the container is generated, right?).

Yes, it is set after container start bind to the computer node setup. By `verdi computer show <localhost-hq>` you can see the number of CPUs can be used (and by API as well).

Issue #888, opened 2024-10-23
Job remain in the queue if 2 cpus are requested.


superstar54:
If we request 2 cpus the job remains in the queue. Add message (even block this for localhost; check in general if it’s possible to know the # of CPUs and warn if > than the default, for any computer (if default is specified)?

unkcpz:
I believe this should be done from front-end widget. The number of CPUs info is explicitly set for the computer by the logic at 
https://github.com/aiidalab/aiidalab-qe/blob/b34e9971b0ce71003d9e0ef8d3829fe8d74a1bef/before-notebook.d/43_start-hq.sh#L48C1-L59C2

The num_cpu set up widget should read from this info and when the number in exceed a warning should be throw out.

mikibonacci:
Hi @superstar54 and @unkcpz , I fixed the estimation of localhost available cpus [here](https://github.com/aiidalab/aiidalab-qe/pull/971), and I added this also in setting the maximum num_cpus we can set for a code (see [here](https://github.com/aiidalab/aiidalab-qe/pull/971/files#r1885706719)). I think this should be sufficient fix.

PR #971, opened 2024-12-15
Fixing the estimation of available cpus for a machine (localhost)

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/937
This fixes https://github.com/aiidalab/aiidalab-qe/issues/888

Issue https://github.com/aiidalab/aiidalab-qe/issues/937 is fixed [here](https://github.com/aiidalab/aiidalab-qe/pull/971/files#r1885707078),
while https://github.com/aiidalab/aiidalab-qe/issues/888 is fixed [here](https://github.com/aiidalab/aiidalab-qe/pull/971/files#r1885706719)


src/aiidalab_qe/common/widgets.py
```diff
@@ -707,19 +707,19 @@ def update_resources(self, change):
...
-            self.num_cpus.max = os.cpu_count()
+            self.num_cpus.max = 1
             self.num_cpus.value = 1
             self.num_cpus.description = "CPUs"
         else:
             default_mpiprocs = computer.get_default_mpiprocs_per_machine()
             self.num_nodes.disabled = False
             self.num_cpus.max = default_mpiprocs
-            self.num_cpus.value = default_mpiprocs
+            self.num_cpus.value = (
```
mikibonacci, line 720: this can fix the number of max cpus we can choose for a code, avoid the erroneous os.cpu_count

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -125,12 +123,9 @@ def check_resources(self):
...
 
-        try:
-            localhost_cpus = len(os.sched_getaffinity(0))
-        except Exception:
-            # Fallback, in some OS os.sched_getaffinity(0) is not supported
-            # However, not so reliable in containers
-            localhost_cpus = os.cpu_count()
+        machine_cpus = orm.load_node(
+            pw_code_model.selected
+        ).computer.get_default_mpiprocs_per_machine()
```
mikibonacci, line 128: I called machine_cpus as is not more only for localhost_cpus

src/aiidalab_qe/common/widgets.py
```diff
@@ -707,19 +707,19 @@ def update_resources(self, change):
             self.set_resource_defaults(load_code(change["new"]).computer)
 
     def set_resource_defaults(self, computer=None):
-        import os
-
-        if computer is None or computer.hostname == "localhost":
+        if computer is None:
             self.num_nodes.disabled = True
```
superstar54, line 711: I think we still want to disable the `num_nodes` if the computer is "localhost".

src/aiidalab_qe/common/widgets.py
superstar54: Why setting `num_cpus` to 1 if it's localhost?

superstar54 (commented):
Hi @mikibonacci , thanks for the work. I am not quite understand the fix. Maybe @unkcpz know better on this part, please have a look.

src/aiidalab_qe/common/widgets.py
mikibonacci: you are right, I will fix it

src/aiidalab_qe/common/widgets.py
mikibonacci: I guess (was already 1, before this PR) that the reason is not to use, by default, the whole resources of the host, if localhost. Imagine in the demo server, you will be then always suggest as default to use all the resources, maybe is not the ideal case. But I am open to change it

superstar54: Thanks for the explanation. It sounds reasonable to use 1.

src/aiidalab_qe/app/submission/global_settings/model.py
unkcpz, line 128:
```suggestion
        code = orm.load_node(
            pw_code_model.selected
        )
        machine_cpus = code.computer.get_default_mpiprocs_per_machine()
```

Could be more readable.

unkcpz (commented):
Thanks @mikibonacci, for getting the set default CPU number `get_default_mpiprocs_per_machine` is the correct way I was putting for the computer. So this part looks good to me.

edan-bainglass (approved):
Simple enough. LGTM!

PR #978, opened 2024-12-16
Redesign status and results step

### LINKED ISSUE #977: Disable download button when clicked, until the download starts
Since it takes times to create the .zip or the .aiida, people might click multiple times.

Also, I would write "Generating file to be downloaded, please be patient..." rather than "Downloading..."

### LINKED ISSUE #990: Change URL when calc is submitted (and PK is assigned)
Adding ?pk=... so that, if I refresh my page, I stay in the same page (and not go to a fresh submission form)

### LINKED ISSUE #994: Submit button doesn't get immediately disabled and it might take 2-3 seconds to go to next step
Users might then click multiple times. Disable it as the very first thing

edan-bainglass:
This PR redesigns step 4 as follows:
- Split step into three "tabs":
  1. Parameters summary and data download controls
  2. Result panels
  3. Process status views
  - for iii, split into:
    1. "simplified" view (to be implemented in a later PR), and
    2. "advanced" view (the current process tree)
- Smoothly handle transition from submission to results - the app now proceeds to step 4 while the builder is created and submitted. Step 4 is updated once the process is ready (#994)
  - Also resolves the issue where the submit button can be repeatedly pressed
- Discard redundant "Update results" button
- Disable raw-data/archive download buttons when downloading (#977)
- Inject process pk to url on submission (#990)

Resolves #994
Resolves #977 
Resolves #990


superstar54 (approved):
Hi @edan-bainglass , thanks for the nice work!

Issue #981, opened 2024-12-16
Add Spin-orbit coupling as a main button


edan-bainglass:
Make it a main button below the one for magnetism.

* Name it: "Spin-orbit coupling": "Off" and "On" like magnetism, with Off by default

* Make sure this also goes into the default label (add "with spin orbit")

edan-bainglass:
@AndresOrtegaGuerrero do you already have a PR for this? I know you've already handled the label in #954. If you're not already handling moving the widget to basic, I got it.

edan-bainglass:
![image](https://github.com/user-attachments/assets/6ab52437-a1ee-4d48-8f77-7364d1105dd3)

@cpignedoli @giovannipizzi thoughts?

giovannipizzi:
lgtm!

cpignedoli:
Super @edan-bainglass. We will probably have to add in a different PR some warnings/info on the pseudopotentials to be used and on the fact that one should consider doing geoopt  without SOC in "some cases". What do you think?

AndresOrtegaGuerrero:
@edan-bainglass, I suggest addressing @cpignedoli's comment in a follow-up PR. Users might perform geometry optimizations with spin-orbit coupling (SOC), which could cause the application to fail repeatedly if resources are not allocated appropriately.

While enabling such calculations is valuable, it might not be ideal for users to proceed without understanding the implications. We should verify whether the example systems have sufficient resources to complete a geometry optimization with SOC (Demo servers resources).

Additionally, it would be helpful to include a caption warning users about the potential need for extra resources when performing SOC calculations compared to calculations without SOC. This would guide users to allocate resources appropriately and avoid unexpected failures.

edan-bainglass:
@AndresOrtegaGuerrero I agree and counter suggest that you do it 😅 You clearly have a good sense of what should be said there. I'm happy to review the text draft once you have it 😁

AndresOrtegaGuerrero:
I would suggest , there should be a discussion on what is the text it should be included and where. Given the previous experience in the pass about captions, once this is being decided i will do the PR.

PR #984, opened 2024-12-17
Move spin orbit to basic settings

edan-bainglass:
This PR moves the spin orbit control from advanced to basic settings. It also simplifies/uniforms the basic settings UI.

Handles #981


mikibonacci (approved):
Hi @edan-bainglass, thank you for the work, looks nice and indeed makes sense to have the SoC in the basics. 

LGTM!

Issue #988, opened 2024-12-17
The "Start new calculation" button gets blocked by a pop-up blocker


edan-bainglass:
At least on Mac OS with Safari.
Maybe just replacing the button with a `<a target="_blank">` would work? These should not be blocked?

Moreover (as discussed): put it as a fourth gray (not green) button after "Job History". Then call it "New calculation (in new tab)".

Duplicate of #891

PR #995, opened 2024-12-18
Rename job history and convert the button to a link to a new tab

superstar54:
Fix https://github.com/aiidalab/aiidalab-qe/issues/996 and https://github.com/aiidalab/aiidalab-qe/issues/997

- Rename `Job history` to `Calculation history`
- Convert the button to a link to a new tab


src/aiidalab_qe/app/wrapper.py
```diff
@@ -275,3 +249,9 @@ def __init__(self) -> None:
                 footer,
             ],
         )
+
+    def _open_calculation_history(self, _):
+        from IPython.display import Javascript
+
+        url = "./calculation_history.ipynb"
+        display(Javascript(f"window.open('{url}', '_blank')"))
```
edan-bainglass, line 257: See #988. I think this will run into a similar issue.

edan-bainglass (changes requested):
Thanks @superstar54. Looks good and works on my end. A few comments:

- We'll have to figure out another approach for linking to another notebook to avoid browser pop-up blocker (#988)

- Need a space between Quantum and ESPRESSO in the notebook title
- Could we make the table expand to full page width?

image

superstar54:
Hi @edan-bainglass , thanks for the review.

> Could we make the table expand to full page width?

Do you know how to do this? I tried `ipw.Layout(width="100%")`, but it does not work.

edan-bainglass (approved):
Excellent. LGTM! Thanks @superstar54. I'll look into the pop-up blocker issue today. Once I have a fix, I can also apply it to your button.

PR #1004, opened 2024-12-19
Implement missing-info warnings

edan-bainglass:
This PR introduces guards on subsequent steps that inform the user if dependencies have yet to be met, i.e., previous step(s) not confirmed.

### Pros

- Reduced chances of messing up
- Reduced development complexities (due to reduced error potential)
- Self-guiding app

### Cons

- *Reduces user exploration

*Not exactly true. The exploration is still available, but now in a prescribed (correct) order.

@cpignedoli @giovannipizzi thoughts?

image
---
image
---
image

### DIFF
--- a/src/aiidalab_qe/app/main.py
+++ b/src/aiidalab_qe/app/main.py
@@ -18,6 +18,7 @@ from aiidalab_qe.app.structure.model import StructureStepModel
 from aiidalab_qe.app.submission import SubmitQeAppWorkChainStep
 from aiidalab_qe.app.submission.model import SubmissionStepModel
 from aiidalab_qe.common.infobox import InAppGuide
+from aiidalab_qe.common.mixins import DependentStep
 from aiidalab_qe.common.widgets import LoadingWidget
 from aiidalab_widgets_base import WizardAppWidget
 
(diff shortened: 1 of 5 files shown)


giovannipizzi:
Very good!

src/aiidalab_qe/app/main.py
```diff
@@ -152,7 +153,18 @@ def _on_configuration_confirmation_change(self, _):
 
     def _render_step(self, step_index):
         step = self.steps[step_index][1]
-        step.render()
+        if step is self.configure_step and not self.structure_model.confirmed:
+            step.show_missing_information_warning()
+        elif step is self.submit_step and not self.configure_model.confirmed:
+            step.show_missing_information_warning()
+        elif step is self.results_step and not self.submit_model.confirmed:
+            step.show_missing_information_warning()
```
superstar54, line 161: Instead of checking it manually, I think we can do this in an automatic manner: one step depends on its previous step.

edan-bainglass: Yeah, probably. But maybe we can do that in another PR? Low on time at the moment. Let's try to get this in for the release and open an issue to revisit this.

superstar54: Good to me!

superstar54 (approved):
Thanks for the work! Please open an issue after merging the PR.

PR #1007, opened 2024-12-20
Default workdir clean on demo server

### LINKED ISSUE #1006: ONLY IN THE DEMO (not othewise): tick by default the clean workdir
The default behavior on the demo server should be to clean the work directory due to storage limitation.

edan-bainglass:
This PR implement a merging mechanism for custom app defaults configurations. Users can now drop a custom `config.yml` in the `/home/jovyan/.aiidalab/quantumespresso/` directory. The file follows the structure of the `qeapp.yml` default configurations file, though it does not strictly require the inclusion of all fields. The custom `config.yml` file is processed and merged recursively into the object created from processing `qeapp.yml`.

Closes #1006

**Note**: I had to modify the updating procedure of relaxation type options and value (PBC-dependent) to accommodate these changes, though I think perhaps this was actually just a bug in that implementation.

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/model.py
@@ -45,7 +45,7 @@ class AdvancedConfigurationSettingsModel(
     electronic_type = tl.Unicode()
     spin_orbit = tl.Unicode()
 
-    clean_workdir = tl.Bool(False)
+    clean_workdir = tl.Bool(DEFAULT["advanced"]["clean_workdir"])
     override = tl.Bool(False)
     total_charge = tl.Float(DEFAULT["advanced"]["tot_charge"])
     van_der_waals_options = tl.List(
--- a/src/aiidalab_qe/app/parameters/__init__.py
+++ b/src/aiidalab_qe/app/parameters/__init__.py
@@ -1,7 +1,54 @@
+from collections.abc import Mapping
 from importlib import resources
+from pathlib import Path
 
 import yaml
 
 from aiidalab_qe.app import parameters
 
+
+def recursive_merge(d1, d2):
+    """Merge dictionary `d2` into dictionary `d1` recursively.
+
+    - For keys that are in both dictionaries:
+        - If values are dictionaries, merge them recursively.
+        - Otherwise, overwrite the value in `d1` with the value in `d2`.
+    - Keys in `d2` not in `d1` are added to `d1`.
+
+    Parameters
+    ----------
+    `d1` : `dict`
+        The dictionary to merge into.
+    `d2` : `dict`
+        The dictionary to merge from.
+
+    Returns
+    -------
+    `dict`
+        The merged dictionary.
+
+    Examples
+    --------
+    >>> d1 = {'a': 1, 'b': {'c': 2, 'd': 3}}
+    >>> d2 = {'b': {'c': 4, 'e': 5}}
+    >>> recursive_merge(d1, d2)
+    {'a': 1, 'b': {'c': 4, 'd': 3, 'e': 5}}
+    """
+    for key, value in d2.items():
+        if key in d1:
+            if isinstance(d1[key], Mapping) and isinstance(value, Mapping):
+                recursive_merge(d1[key], value)
+            else:
+                d1[key] = value
+        else:
+            d1[key] = value
+    return d1
+
+
 DEFAULT_PARAMETERS = yaml.safe_load(resources.read_text(parameters, "qeapp.yaml"))
+
+
+custom_config_file = Path.home() / ".aiidalab" / "quantumespresso" / "config.yml"
+if custom_config_file.exists():
+    custom_config = yaml.safe_load(custom_config_file.read_text())
+    DEFAULT_PARAMETERS = recursive_merge(DEFAULT_PARAMETERS, custom_config)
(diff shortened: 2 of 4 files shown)


src/aiidalab_qe/config/environment.py
```diff
@@ -0,0 +1,13 @@
...
+
+CONFIG_PATH = Path.home() / ".aiidalab" / "config.yml"
+
+
+def on_demo_server():
+    if CONFIG_PATH.exists():
+        with CONFIG_PATH.open("r") as file:
+            config = yaml.safe_load(file)
+            return config.get("on_demo_server", False)
+    return False
```
superstar54, line 13:
```suggestion
def load_config():
    config = {}
    if CONFIG_PATH.exists():
        with CONFIG_PATH.open("r") as file:
            config = yaml.safe_load(file)
    return config
```

src/aiidalab_qe/app/configuration/advanced/model.py
```diff
@@ -14,6 +14,7 @@
 from aiidalab_qe.app.parameters import DEFAULT_PARAMETERS
 from aiidalab_qe.common.mixins import HasInputStructure, HasModels
 from aiidalab_qe.common.panel import ConfigurationSettingsModel
+from aiidalab_qe.config.environment import on_demo_server
```
superstar54, line 17:
```suggestion
from aiidalab_qe.config.environment import load_config
```

superstar54 (changes requested):
I suggest using a general config file, instead of hard-coding the `demo server` inside the app. See the suggested code.

edan-bainglass:
Thanks @superstar54. Good suggestion! I'll implement it in a bit. 

Just to clarify, I would then do `get_config().get("on_demo_server", False)`, yeah?

superstar54:
> Just to clarify, I would then do `get_config().get("on_demo_server", False)`, yeah?

I would not include any demo server related to the app.
```python
clean_workdir = tl.Bool(load_config().get("clean_workdir", False))
```

edan-bainglass:
> > Just to clarify, I would then do `get_config().get("on_demo_server", False)`, yeah?
> 
> I would not include any demo server related to the app.
> 
> ```python
> clean_workdir = tl.Bool(load_config().get("clean_workdir", False))
> ```

Oh I see. In that case, I wouldn't even use a new config file. I would just use the qeapp.yaml file, no?

edan-bainglass:
> > > Just to clarify, I would then do `get_config().get("on_demo_server", False)`, yeah?
> > 
> > 
> > I would not include any demo server related to the app.
> > ```python
> > clean_workdir = tl.Bool(load_config().get("clean_workdir", False))
> > ```
> 
> Oh I see. In that case, I wouldn't even use a new config file. I would just use the qeapp.yaml file, no?

@giovannipizzi can I get your thoughts on this? I can add `clean_workdir` to qeapp.yaml, which is already being used for various defaults. But this does not provide auto detection for the demo server. It just means that when we deploy on the demo server, we need to modify this file. I'm not sure this is the correct approach, as it means modifying a tracked file in the actual codebase. I think perhaps it would be better to rely on some external config file, maybe again in /home/user/.aiidalab? 

Note that environment variables didn't work out here. Couldn't get Jupyter to detect them.

giovannipizzi:
I agree we should not change a file from he installed package. Probably the app should first load the yaml you mention, then allow for an optional override file in the home folder. If present, those variables take precedence, essentially you load both and do a dictionary update before applying (or something like this, this works for simple values, not sure of you also have dictionaries or lists and need more customized overrides).
If not present, no warning is raised, you just use the defaults from the first file in the code base. 

Then, in the deployment you just drop the overrides file with the fee things you want to change.

Would that work?

edan-bainglass:
> I agree we should not change a file from he installed package. Probably the app should first load the yaml you mention, then allow for an optional override file in the home folder. If present, those variables take precedence, essentially you load both and do a dictionary update before applying (or something like this, this works for simple values, not sure of you also have dictionaries or lists and need more customized overrides). If not present, no warning is raised, you just use the defaults from the first file in the code base.
> 
> Then, in the deployment you just drop the overrides file with the fee things you want to change.
> 
> Would that work?

Possibly. Let me see.

edan-bainglass:
> > I agree we should not change a file from he installed package. Probably the app should first load the yaml you mention, then allow for an optional override file in the home folder. If present, those variables take precedence, essentially you load both and do a dictionary update before applying (or something like this, this works for simple values, not sure of you also have dictionaries or lists and need more customized overrides). If not present, no warning is raised, you just use the defaults from the first file in the code base.
> > Then, in the deployment you just drop the overrides file with the fee things you want to change.
> > Would that work?
> 
> Possibly. Let me see.

Works like a charm. Matter of fact, this is great for development, as I can have an external config with the simplest settings (no relaxation, fast protocol, clean work directory), saving me clicks and storage!

src/aiidalab_qe/app/parameters/__init__.py
```diff
@@ -1,7 +1,57 @@
...
 
 from aiidalab_qe.app import parameters
 
+
+def recursive_merge(d1, d2):
+    """Merge dictionary `d2` into dictionary `d1` recursively.
+
+    - For keys that are in both dictionaries:
+        - If values are dictionaries, merge them recursively.
+        - If values are lists, extend the list in `d1`.
```
superstar54, line 15: I suggest overriding the list instead of extending it, which is better controllable for the following reasons:
- the user can override the list entirely.
- when recursive merge multiple times, it will not add duplicate data.

superstar54 (approved):
Nice!

PR #1011, opened 2024-12-20
Fix safari pop-up issue with new `LinkButton` widget

### LINKED ISSUE #891: Start a new calculation button does not work because the pop-up is prevented by my browser
The “start new calculation” button does not work because the pop-up is prevented by my browser (Safari). Maybe replace button with link and target=”_blank” that should not have the issue I think?

### LINKED ISSUE #988: The "Start new calculation" button gets blocked by a pop-up blocker

edan-bainglass:
This PR introduces a `LinkButton` widget that mimics a `ipywidgets.Button` via an `ipywidgets.HTML` wrapper around an anchor tag (`<a>`).

Resolves #891
Resolves #988

### DIFF
--- a/src/aiidalab_qe/app/static/styles/custom.css
+++ b/src/aiidalab_qe/app/static/styles/custom.css
@@ -20,15 +20,50 @@
   font-style: italic;
 }
 
-.info-toggles {
+.app-controls {
   margin: 0 auto;
+  align-items: center;
 }
-.info-toggles button {
-  width: 100px;
-  margin: 1em 0.5em;
+.app-controls button,
+.app-controls .link-button {
+  width: fit-content;
+  margin: 2px 0.5em;
 }
-.info-toggles button:focus {
+.app-controls button:focus {
   outline: none !important;
+}
+
+/* replicate jupyter button behavior */
+.link-button:hover,
+.link-button:focus {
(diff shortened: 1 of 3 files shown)


superstar54:
Thanks for the work!
Have you tested this on the Safari browser?

edan-bainglass:
> Thanks for the work! Have you tested this on the Safari browser?

Sure haven't. @giovannipizzi could you test the PR, see if it fixes your issue?

edan-bainglass:
> Thanks for the work! Have you tested this on the Safari browser?

Hi @superstar54 for the sake of time, can we approve this one? A quick lookup online suggests that using an anchor tag (via an HTML widget) is standard across browsers and should not trigger a pop-up in Safari (where a button click trigger of JS `window.open` does).

Worst case, it doesn't work (same situation we're in without this PR) and we pursue another solution.

superstar54 (approved):
> Hi @superstar54 for the sake of time, can we approve this one? A quick lookup online suggests that using an anchor tag (via an HTML widget) is standard across browsers and should not trigger a pop-up in Safari (where a button click trigger of JS window.open does).

Sure! Thanks for the work!

PR #1012, opened 2024-12-20
Replace advanced settings override checkbox with a reset button

### LINKED ISSUE #991: Why do we need an 'override' checkbox?
If I activate it, shouldn't all parameters be the defaults?
If not, it's a bit confusing, because maybe I want simply to activate spin-orbit but then some other behavior changes.

If instead they are just the defaults, it's strange that I can't do anything (I was suprised myself that the spin-orbit toggle was not working) until I check this box placed in a special top-right cornet.

I vote to just remove it and keep all buttons, checkboxes, ... active.

**NOTE**: if removed, we should adapt the text below removing at least 2 occurrences of "Tick the box to override the default,"

Rather, replace with a 'Reset defaults' button.

edan-bainglass:
This PR discards the override mechanism in the advanced settings panel. All widgets are enabled and may be overridden by the user. A reset button is added in place of the override checkbox to reset the advanced settings to their defaults w.r.t the basic settings and selected structure.

Closes #991


AndresOrtegaGuerrero:
I would suggest , that you add a message to tell the user that if the calculation is magnetic  "Set the magnetization in advanced settings to select your desired magnetic configuration."

superstar54 (changes requested):
Thanks for the work! I found two small issues when resetting to the default
- the kpoint is reset to 0.1 when protocol is moderate (should be 0.15)
- the Wavefunction cutoff and Charge density cutoff are not reset.

edan-bainglass:
> Thanks for the work! I found two small issues when resetting to the default
> 
> * the kpoint is reset to 0.1 when protocol is moderate (should be 0.15)
> * the Wavefunction cutoff and Charge density cutoff are not reset.

Thanks @superstar54. I'll take a look.

edan-bainglass:
> > Thanks for the work! I found two small issues when resetting to the default
> > 
> > * the kpoint is reset to 0.1 when protocol is moderate (should be 0.15)
> > * the Wavefunction cutoff and Charge density cutoff are not reset.
> 
> Thanks @superstar54. I'll take a look.

@superstar54 I can't reproduce your issues. Everything resets as it should on my end.

edan-bainglass:
> I would suggest , that you add a message to tell the user that if the calculation is magnetic "Set the magnetization in advanced settings to select your desired magnetic configuration." Since the default is 0 , so it wont have magnetization

image

edan-bainglass:
> > > Thanks for the work! I found two small issues when resetting to the default
> > > 
> > > * the kpoint is reset to 0.1 when protocol is moderate (should be 0.15)
> > > * the Wavefunction cutoff and Charge density cutoff are not reset.
> > 
> > 
> > Thanks @superstar54. I'll take a look.
> 
> @superstar54 I can't reproduce your issues. Everything resets as it should on my end.

@AndresOrtegaGuerrero can you check as well?

superstar54:
> the Wavefunction cutoff and Charge density cutoff are not reset.

I still has this.

edan-bainglass:
> > the Wavefunction cutoff and Charge density cutoff are not reset.
> 
> I still have this.

Okay @superstar54. This should be resolved now. Please retest 🙏

AndresOrtegaGuerrero:
If I select a different Pseudos set, like Pseudo Dojo, and adjust the cutoffs, clicking the reset button not only resets the cutoffs but also switches back to SSSP efficiency. Shouldn't it retain the selected Pseudos and only reset the cutoffs?

src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -78,29 +74,21 @@ def render(self):
 
         # clean-up workchain settings
         self.clean_workdir = ipw.Checkbox(
-            description="",
+            description="Tick to clean-up the work directory after the calculation is finished",
```
AndresOrtegaGuerrero, line 77: Tick to delete the work directory after the calculation finishes.

edan-bainglass: It is left empty on purpose, so we can add a label of the same text on the right.

edan-bainglass:
> 
> If I select a different Pseudos set, like Pseudo Dojo, and adjust the cutoffs, clicking the reset button not only resets the cutoffs but also switches back to SSSP efficiency. Shouldn't it retain the selected Pseudos and only reset the cutoffs?

The reset button returns everything back to what it was w.r.t the structure and protocol. For a moderate protocol, the default is SSSP efficiency.

src/aiidalab_qe/app/configuration/advanced/advanced.py
AndresOrtegaGuerrero: I meant that the description text should be maybe be "Tick to delete the work directory after the calculation finishes."

AndresOrtegaGuerrero:
> > If I select a different Pseudos set, like Pseudo Dojo, and adjust the cutoffs, clicking the reset button not only resets the cutoffs but also switches back to SSSP efficiency. Shouldn't it retain the selected Pseudos and only reset the cutoffs?
> 
> The reset button returns everything back to what it was w.r.t the structure and protocol. For a moderate protocol, the default is SSSP efficiency.

Yes, it did feel a bit odd to go there (change the pseudos selection) while making the change, but you're right — it should have that behavior since that's how it has always worked.

src/aiidalab_qe/app/configuration/advanced/advanced.py
edan-bainglass: Oh my bad, the git diff was confusing. Okay, yeah, this has a description and you're suggesting to reword. I agree if that's what actually happens. If the user ticks this, does it delete the workdir, or does it just clean it?

AndresOrtegaGuerrero: It deletes the work directory. I think some users might find the term "clean up" confusing, so it's better to clearly state that it deletes the directory. This clarification will also be useful for the post-processing calculations. The clearer, the better.

AndresOrtegaGuerrero (approved):
LGTM!  thank you @edan-bainglass

superstar54 (approved):
LGTM! Thanks for the nice work!

PR #1014, opened 2024-12-22
Clean up panel classes

edan-bainglass:
This PR addresses preliminary modifications in the panel module required by follow-up PRs.

Note that the cause behind the failing `test_workchainview` test is temporary and is resolved in the follow up PR (#1015).


superstar54 (approved):
LGTM!

PR #1015, opened 2024-12-22
Implement constant but relaxation-dependent structure results panel

### LINKED ISSUE #993: If you don't relax the structure, show a tab "Initial structure"
This is useful because if the user might have to filter by atom index, and it's important to have it close to the results

Also:
- [ ] Rename the tab when some relaxation has been performed to "Relaxed structure" (instead of Final Geometry), note also capitalization here!

edan-bainglass:
This PR implements a structure results panel that is always shown. The tab title and source node are dependent on the nature of the calculation as follows:
- No relaxation - title = "Initial structure"; source = `node.inputs.structure`
- Some relaxation - title = "Relaxed structure"; source = `node.outputs.structure`

Auto rendering is adjusted to ensure the initial structure panel is not dependent on having results (there are none) but instead is always auto rendered.

Closes #993

### DIFF
--- a/src/aiidalab_qe/app/result/components/viewer/structure/__init__.py
+++ b/src/aiidalab_qe/app/result/components/viewer/structure/__init__.py
@@ -1,7 +1,7 @@
 from .model import StructureResultsModel
-from .structure import StructureResults
+from .structure import StructureResultsPanel
 
 __all__ = [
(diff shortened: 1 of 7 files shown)


superstar54 (approved):
LGTM!

PR #1016, opened 2024-12-22
increase fat bands slider max to 5.0 eV

### LINKED ISSUE #899: Increase flat bands max
If one wants to show all bands (including semicore) the max 0.76 eV is too small. Would it be possible to have the max up to 5eV for instance? Or can this be a “relative line thickness”? (Check e.g. gold fast PDOS and show all)

AndresOrtegaGuerrero:
This PR fixes #899 , 
The max limit of the widget is set to 5.0 eV


edan-bainglass (approved):
LGTM!

PR #1019, opened 2024-12-23
Provide categorized example structures.

### LINKED ISSUE #1018: Update the examples tab
1. since we want to first provide only "simple" structures that can be run easily in the demo server, add on top of the dropdown a set of buttons (or tabs), with the following names:
   * Simple crystals
   * XPS examples
   * XAS examples
2. when the button is clicked, only the respective examples are shown (in the future one can think to make these buttons or (sub)tabs extensible by the plugins
3. Move the 3 XPS/XAS structures inside the correct tab
4. For the "Simple crystals" tab, replace the current ones with the following list (IMPORTANT: remove the string "Primitive cell" from everywhere, as done below, as we should always ONLY provide the primitive cell):
   * Bulk silicon
   * Diamond
   * Gallium arsenide
   * Gold
   * LiCoO2
   * Nickel
   * Silicon oxide
5. IMPORTANT NOTE: for Si, C, GaAs, SiO2, OK to keep the current structure. For gold, replace the current one with its primitive cell. For Cobalt, remove it and replace with Nickel FCC.

Note: we are now testing LiCoO2 and we will report back if it's fast enough to be run in Fast mode (then we keep it), or too slow (then we remove it). If we keep it, write it as "Lithium cobalt oxide (LiCoO₂)" (note the use of unicode subscript 2, so it works inside the dropdown)

superstar54:
Fix #1018 

[x] Provide categorized example structures, and only the respective examples are shown when the button is clicked. 
   * Simple crystals
   * XPS examples
   * XAS examples

[x] The example catalog is extensible by the plugins.
[x] For the "Simple crystals" tab, replace the current ones with the following list (IMPORTANT: remove the string 
[x] Remove the "Primitive cell" from everywhere, as done below, as we should always ONLY provide the primitive cell):
   * Bulk silicon
   * Diamond
   * Gallium arsenide
   * Gold
   * LiCoO2
   * Nickel
   * Silicon oxide

[x] For gold, replace the current one with its primitive cell.
[x] For Cobalt, remove it and replace with Nickel FCC.

Note for the `Silicon oxide`: I keep the `alpha quartz` because `Silicon oxide` has many possible structures, and it's important to tell which one we are using.


edan-bainglass (approved):
LGTM @superstar54. Note that the new widget is not MVC compatible and will load on initialization. We're trying to avoid this. We should consider revisiting later (that is, if we don't go the React/Solara way). But for now, the widget is small enough that it should add much of an impact.

superstar54:
> Note that the new widget is not MVC compatible and will load on initialization. 

The widget `CategorizedStructureExamplesWidget` is inherited from `StructureExamplesWidget`, which is from `aiidalab-widget-base`, that's why I didn't use MVC. We need to refactor the `aiidalab-widget-base` using MVC, but in the future.

PR #1023, opened 2024-12-23
Set fat_bands=True when FatBands is required

### LINKED ISSUE #1020: The 'fat bands' checkbox should be selected by default, if the user requested fat bands in the input

AndresOrtegaGuerrero:
This closes #1020


edan-bainglass (approved):
LGTM!

PR #1025, opened 2024-12-23
switch on the `-pd` option for pw code

### LINKED ISSUE #1021: Issue with small systems on large machines
If I run a simple system (e.g. Lithium BCC or similar) with the fast protocol and 4 pools (for the pw.x part), when I get to the dos.x and projwfc.x parts, the calculations crash with 
```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine  fft_type_set (6):
  there are processes with no planes. Use pencil decomposition (-pd .true.)
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```
Could we check if it's really just a matter of using this command line option, if it's safe to do in any case or when it should be used (maybe with an error handler?), or check anyway hot to ensure this crash does not happen?

This was run on Daint @ CSCS, with QE 7.1

superstar54:
fix #1021 
Enable pencil decomposition for all `pw.x` calculations.

**Note**, I still get issue on `projwfc`,
 
```console
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine  blk2cyc_zredist (1):
      nb less than the number of proc
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

### DIFF
--- a/calculation_history.ipynb
+++ b/calculation_history.ipynb
@@ -59,7 +59,7 @@
     "        </p>\n",
     "        <ul>\n",
     "            <li><b>Label Search Field:</b> Enter a job label to find matching jobs.</li>\n",
-    "            <li><b>Job State Dropdown:</b> Filter jobs based on their state (e.g., finished, waiting).</li>\n",
+    "            <li><b>Job State Dropdown:</b> Filter jobs based on their state (e.g., finished, running).</li>\n",
     "            <li><b>Date Range Picker:</b> Select a start and end date to view jobs created within that range.</li>\n",
     "            <li><b>Properties Filter:</b> Select specific properties associated with jobs.</li>\n",
     "        </ul>\n",
--- a/src/aiidalab_qe/plugins/bands/workchain.py
+++ b/src/aiidalab_qe/plugins/bands/workchain.py
@@ -1,7 +1,9 @@
-from aiida import orm
 from aiida.plugins import WorkflowFactory
 from aiida_quantumespresso.common.types import ElectronicType, SpinType
-from aiidalab_qe.plugins.utils import set_component_resources
+from aiidalab_qe.utils import (
+    enable_pencil_decomposition,
+    set_component_resources,
+)
 
 BandsWorkChain = WorkflowFactory("aiidalab_qe.bands_workchain")
 # from .bands_workchain import BandsWorkChain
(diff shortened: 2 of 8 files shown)


calculation_history.ipynb
```diff
@@ -59,7 +59,7 @@
     "        </p>\n",
     "        <ul>\n",
     "            <li><b>Label Search Field:</b> Enter a job label to find matching jobs.</li>\n",
-    "            <li><b>Job State Dropdown:</b> Filter jobs based on their state (e.g., finished, waiting).</li>\n",
+    "            <li><b>Job State Dropdown:</b> Filter jobs based on their state (e.g., finished, running).</li>\n",
```
superstar54, line 62: A small fix that is not related to this PR. But I fix it to avoid creating another PR.

src/aiidalab_qe/plugins/bands/workchain.py
```diff
@@ -33,12 +34,18 @@ def update_resources(builder, codes):
     if "bands" in builder:
         set_component_resources(builder.bands.scf.pw, codes.get("pw"))
         set_component_resources(builder.bands.bands.pw, codes.get("pw"))
+        builder.bands.scf.pw.settings = orm.Dict({"CMDLINE": ["-pd", ".true."]})
```
AndresOrtegaGuerrero, line 37: should it be worth it to create a function in utils like use pencil_decomposition ? like that the code is not repeated , but also is easy to read in the future

superstar54: Good point, I will add a function.

AndresOrtegaGuerrero: perhaps also this should be control in the PwCode model ? so the user can set and unset this ?

superstar54: There are two ways to handle this:
- as you suggested, add a setting in the GUI so that the `advanced` user can control it.
- in the WorkChain, it handles the setting automatically or adds an error_handler to restart with a different number of CPUs.

I suggest discussing which way we want to implement in the future PR.

AndresOrtegaGuerrero: yeah , i think we can leave it here, but in another PR we just leave the freedom plus the explanation to the user

AndresOrtegaGuerrero (approved):
LGTM!, thank you @superstar54 , I would suggest you open an issue, for implementing this in the resource panel in such a way the users can activate or deactivate this. Plus adding explanation to the users about resources and the computers they are using

superstar54:
Thanks @AndresOrtegaGuerrero , I opened an issue: https://github.com/aiidalab/aiidalab-qe/issues/1031

PR #1028, opened 2024-12-23
Install aiida-pseudos from repo to avoid download issue

### LINKED ISSUE #1026: Docker build fails due to missing certificate when downloading the pseudos
This is an issue with `aiida-pseudo` "fixed" in https://github.com/aiidateam/aiida-pseudo/pull/179, but not yet released, so CI is not including the fix.

Tagging @superstar54 who is working on a solution as a reminder to close this issue 🙂

superstar54:
fix #1026 

This is a temporary solution to install aiida-pseudos from repo to avoid downloading issues
Note, since `aiida-pseudo` need `aiida-core` > 2.6, however, we still want to support the previous aiida-core version, so in the image, we use the lastest aiida-pseudo to download the pseudos and install it. After the pseudos are installed, it can be used by the App.


edan-bainglass (approved):
LGTM! Glad this matter is resolved.

PR #1017, opened 2024-12-22
using a center layout for the widget BandsPdosWidget

AndresOrtegaGuerrero:
This PR addresses the issue in #976

Issue #820, opened 2024-09-26
legend is not visible or only partially visible in narrow windows


superstar54:
For narrow windows, especially if bands+PDOS are shown, the legend is not visible or only partially visible.
- Question: is it possible to call a JS function to know the window size, and show selectively a message? so that if the window is < XXX pixels (XX to check but it’s easy to see), an additional message is shown: “NOTE: the window is too narrow - enlarge it to see also the legend and toggle the plot lines.”. This needs to be show just above the plot (below the message discussed above on hovering to show the home button etc.”
- If it’s difficult, communicate back to Carlo & Giovanni who can provide a sentence to show always

AndresOrtegaGuerrero:
@cpignedoli  Could you provide a snapshot to understand a lil better this behaviour ?

cpignedoli:
Could be browser dependent, in my browser (crome) it's less of a problem because tehre are no "intermediate cases" : 
The plot should look like:
<img width="621" alt="image" src="https://github.com/user-attachments/assets/ab438a97-4893-4f87-9c50-91b4bf3f873d">


It is probably obvious that the user should enlarge the window in case the plot looks like:

<img width="494" alt="image" src="https://github.com/user-attachments/assets/4d7ba66e-a112-489c-ac64-06bfb222989f">

**But there could be cases (we have seen it from the pc of @giovannipizzi ) where for example you miss "just the legend" and the user may not know there is a legend...**

The idea is: to know how many pixels we need to visualize correctly the plot, and if the window is smaller we prompt a warning message like: "Please enlarge the browser window, a portion of the plot is missing"

superstar54:
You can debug on devices with different screen sizes inside the browser. Press `F12`, and toggle the device toolbar,



https://github.com/user-attachments/assets/41eb6b41-ffe0-42ab-ad7d-bb8b4b8972ed

PR #1036, opened 2024-12-24
adding a center_layout to get the plot centered

AndresOrtegaGuerrero:
This PR address #1017 , not it just center the plotly widget

### DIFF
--- a/src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
+++ b/src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
@@ -261,12 +261,17 @@ class BandsPdosWidget(ipw.VBox):
         self._model.fetch_data()
         self._model.create_plot()
         self.plot = self._model.plot
+
+        center_layout = ipw.Layout(
+            display="flex", justify_content="center", align_items="center"
+        )
+        centered_plot = ipw.Box([self.plot], layout=center_layout)
         self.proj_bands_width_slider.layout.visibility = (
             "visible" if self._model.project_bands_box else "hidden"
         )
         self.download_button.layout.visibility = "visible"
         self.project_bands_box.layout.visibility = "visible"
-        self.children = (*self.children, self.plot, self.color_selector)
+        self.children = (*self.children, centered_plot, self.color_selector)
 
     def _update_bands_projections(self, _):
         """Update the plot with the selected projection."""
--- a/src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
+++ b/src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
@@ -261,17 +261,19 @@ class BandsPdosWidget(ipw.VBox):
         self._model.fetch_data()
         self._model.create_plot()
         self.plot = self._model.plot
-
-        center_layout = ipw.Layout(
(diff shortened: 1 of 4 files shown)


src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
```diff
@@ -261,12 +261,17 @@ def _initial_plot(self):
...
+            display="flex", justify_content="center", align_items="center"
+        )
+        centered_plot = ipw.Box([self.plot], layout=center_layout)
         self.proj_bands_width_slider.layout.visibility = (
             "visible" if self._model.project_bands_box else "hidden"
         )
         self.download_button.layout.visibility = "visible"
         self.project_bands_box.layout.visibility = "visible"
-        self.children = (*self.children, self.plot, self.color_selector)
+        self.children = (*self.children, centered_plot, self.color_selector)
```
edan-bainglass, line 275: How about just
```python
self.children = [
    *self.children,
    ipw.Box(
        children=[self.plot],
        layout=ipw.Layout(margin="0 auto"),
    ),
    self.color_selector,
]
```
You can keep the variables extracted if you prefer, but I think the use of `flex` is overkill. The plot is centered vertically by default. Setting the horizontal margin to `auto` is enough for our purposes.

src/aiidalab_qe/plugins/bands/result/result.py
```diff
@@ -13,6 +13,6 @@ def _render(self):
         bands_node = self._model.get_bands_node()
         model = BandsPdosModel()
         widget = BandsPdosWidget(model=model, bands=bands_node)
-        widget.layout = ipw.Layout(width="fit-content", margin="0 auto")
+        widget.layout = ipw.Layout(width="1000px")
```
edan-bainglass, line 16: I don't see immediately why these are necessary. I removed all of them and everything seems to work as expected. Can you explain the motivation behind these lines?

edan-bainglass (commented):
I think we can simplify this a bit. Please see my comments.

src/aiidalab_qe/plugins/bands/result/result.py
edan-bainglass: Note that I would simply remove the layout assignment here (and in the other results panels)

AndresOrtegaGuerrero: I just went back to how it was previously

AndresOrtegaGuerrero: we added these because apparently in some cases like in a phono or in a different size of screen , it was showing a different size and behaviour, now i dont remember , i vote for keep it as it was previously and then put an issue to confirm if we can remove it

edan-bainglass: This pertains to issue #820 (PR #839). However, it is no longer necessary. We get the horizontal scroll bar by default.

I removed the layout lines from the result panels and...
image
I still get a horizontal scroll bar when needed (narrow screen, zoomed in, etc.).

I vote we remove these layout lines. I don't think they were an appropriate solution from the beginning.

edan-bainglass (approved):
Excellent! LGTM. Thanks @AndresOrtegaGuerrero

Issue #818, opened 2024-09-26
Combine the 3 tabs: bands, El. Structure and PDOS into a single top-level band, called “Electronic structure”


superstar54:
- In the plugin input checkboxes, call the properties “Electronic structure (bands)” and “Electronic structure (PDOS)” to make it clear it’s going to be in the “Electronic structure” output tab in the end.
- If it’s easy, only show one advanced input tab for both called “Electronic structure”: there are not too many advanced options, so you can just put both the PDOS and bands in the same tab. Otherwise OK to keep as it is now, but give the very same name to the tabs as the checkboxes: “Electronic structure (bands)” and “Electronic structure (PDOS)”.
In the output, only show a single “Electronic structure” results tab.
- As a first simple step, just move three tabs inside this tab, calling them: “Bands only”, “PDOS only”, “Combined bands+PDOS”
In the future, open an issue (to be then rediscussed for the next release) whether we have no internal tabs, we have only the combined plot, and we simply have some checkbox or similar (e.g. a dropdown?) to decide what to show.

edan-bainglass:
Moving comment from duplicate project item:

In Step 2.1 rename to:
Electronic projected density of states (PDOS)

In the Results tabs, only one, the name depends on what is computed:
Electronic PDOS
Electronic bands
Electronic bands + PDOS

PR #1037, opened 2024-12-26
Add `from_nodes` class method to instantiate a `BandsPdosModel` from AiiDA nodes

edan-bainglass:
This PR adjusts the bands-pdos MVC component by removing the required bands/pdos `AttributeDict` nodes from the widget constructor and adding a `from_nodes` class method to the model that takes care of the `AttributeDict` extraction. This is one in several future PRs that aims to both make this MVC component more usable outside of the app (for local plotting of the electronic structure), and prepare the component for the future unification of electronic structure results tabs.

### DIFF
--- a/src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
+++ b/src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
@@ -7,31 +7,9 @@ from .model import BandsPdosModel
 
 
 class BandsPdosWidget(ipw.VBox):
-    """
-    A widget for plotting band structure and projected density of states (PDOS) data.
+    """A widget for plotting band structure and projected density of states (PDOS)."""
 
-    Parameters
-    ----------
-    - bands (optional): A node containing band structure data.
-    - pdos (optional): A node containing PDOS data.
-
-    Attributes
-    ----------
-    - description: HTML description of the widget.
-    - dos_atoms_group: Dropdown widget to select the grouping of atoms for PDOS plotting.
-    - dos_plot_group: Dropdown widget to select the type of PDOS contributions to plot.
-    - selected_atoms: Text widget to select specific atoms for PDOS plotting.
-    - update_plot_button: Button widget to update the plot.
-    - download_button: Button widget to download the data.
-    - project_bands_box: Checkbox widget to choose whether projected bands should be plotted.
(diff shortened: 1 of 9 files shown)


edan-bainglass:
@AndresOrtegaGuerrero a colleague was complaining quite a bit about not being able to use the app's widgets to quickly plot some electronic structure stuff. This PR is trying to address this, while also starting to address #818. At the moment, I am providing the means to easily create the MVC component from workflow nodes. I will probably commit a change in a bit to allow also from the calculation nodes, which is the colleague's case.

src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
```diff
@@ -7,31 +7,9 @@
 
 
 class BandsPdosWidget(ipw.VBox):
-    """
-    A widget for plotting band structure and projected density of states (PDOS) data.
```
AndresOrtegaGuerrero, line 11: If you are removing mos of the Docstring, add it where it should be , so is easy for someone in the future to do changes

edan-bainglass: Agreed, but later, once we understand the design better 👍 This is not done.

edan-bainglass: Same goes for tests. No need to comment on them not passing. I'll adjust w.r.t the final design.

src/aiidalab_qe/common/bands_pdos/model.py
```diff
@@ -79,6 +82,37 @@ def __init__(self, *args, **kwargs):
             lambda _: self._has_pdos or self.needs_projections_controls,
         )
 
+    @classmethod
+    def from_nodes(
+        cls,
```
AndresOrtegaGuerrero, line 87: I think you should include some doc string here, to explain why these decisions, and that the bands conditionals is to guarantee backwards compatibility

edan-bainglass: Sure. Once the design is settled, docstrings 👍

edan-bainglass:
@AndresOrtegaGuerrero I'm closing this one, as it does not work as a partial PR. #1039 picks these changes up and completes the target feature.

PR #1039, opened 2024-12-27
Unify electronic structure result panels

### LINKED ISSUE #818: Combine the 3 tabs: bands, El. Structure and PDOS into a single top-level band, called “Electronic structure”

edan-bainglass:
This PR implements a single unified panel for electronic structure results. The title of the panel's tab is dynamic depending on the selected properties. In addition, if both bands and PDOS are selected, selection controls are added at the top of the panel for the user to choose which property they would like to plot.

Closes #818

### TODOs
- [x] Handle status messages - currently remain even when results have been loaded. This is because in the new electronic structure panel, you may load partial results but would still like to know of the progress of the rest.
- [x] Decide if bands and pdos result panel classes should be discarded


edan-bainglass:
@superstar54 @AndresOrtegaGuerrero can you please test this one locally? No need to review the code quite yet. But good if you can confirm that it does work and provide feedback regarding the UI 🙏

edan-bainglass:
The monitor currently checks the workflow every 2 seconds. Can we at least push this to 5? I know @superstar54 mentioned that we are planning to veer away from interval-based monitoring, but until then, at least 5s, so we don't scan so often?

AndresOrtegaGuerrero:
I think the results tabs title should be fix to Electronic structure
I also think the position of the buttons should be on the right , 
image

Instead of `Select one or more properties to plot:`
what about , 
`Select one or more properties to plot: You can choose to plot only bands, only PDOS, or both together. After making your selection, click "Apply selection" to proceed.`

I think if there is only one properties, bands or pdos , this shoudnt be displayed 
image

edan-bainglass:
> I think the results tabs title should be fix to Electronic structure 

See issue for requirements (last comment)

> I also think the position of the buttons should be on the right 

Right, this was already suggested to me. I chose to do it this way to align the button nicely with the edge of the box, but indeed, it felt a bit off. Okay, will move to right.

> Instead of `Select one or more properties to plot:` what about , `Select one or more properties to plot: You can choose to plot only bands, only PDOS, or both together. After making your selection, click "Apply selection" to proceed.`

I see, maybe indeed it is not so clear that bands + PDOS is actually a single combined plot. Okay, will add info!

> I think if there is only one properties, bands or pdos , this shoudnt be displayed 

Yep, let me add that.

edan-bainglass:
@AndresOrtegaGuerrero please recheck 🙏

superstar54:
> The monitor currently checks the workflow every 2 seconds. Can we at least push this to 5?

Ok to me!

edan-bainglass:
> > The monitor currently checks the workflow every 2 seconds. Can we at least push this to 5?
> 
> Ok to me!

Wow, I was off. It's 0.2 seconds, i.e. 1/5 of a second! That's a lot of monitoring events. Perhaps for now, I move it up to 1s.

edan-bainglass:
@superstar54 what should we do with the now-unused bands/pdos results MVC components? Discard?

AndresOrtegaGuerrero:
I propose removing the bands/PDOS result panels since they are currently unused. However, if we plan to implement the feature for defining boxes for the Local DOS (as described in [INPUT_PROJWFC documentation](https://www.quantum-espresso.org/Doc/INPUT_PROJWFC.html#idm102)) for the PDOS plugin, we could retain them. This feature is particularly relevant for comparing STM measurements. Even so, we could still display these results within the electronic structure tab if implemented in the future.

edan-bainglass:
> I propose removing the bands/PDOS result panels since they are currently unused. However, if we plan to implement the feature for defining boxes for the Local DOS (as described in [INPUT_PROJWFC documentation](https://www.quantum-espresso.org/Doc/INPUT_PROJWFC.html#idm102)) for the PDOS plugin, we could retain them. This feature is particularly relevant for comparing STM measurements. Even so, we could still display these results within the electronic structure tab if implemented in the future.

I will discard for now. If we need them, we can pull them back out from the git history.

edan-bainglass:
@AndresOrtegaGuerrero please review once more 🙏

AndresOrtegaGuerrero (approved):
LGTM!  Thank you!

AndresOrtegaGuerrero:
Just dont forget the conflict

edan-bainglass:
> Just dont forget the conflict 

No need for these reminders 😅 I rebase all my PRs on main before merging.

PR #1040, opened 2024-12-27
Fix sentence case issues

edan-bainglass:


AndresOrtegaGuerrero:
should we also change this editor , 
image


In the history there are quite some 
image

And in the step 2 
image

Metal and Insulator have the first letter with CAP, while in the protocol all are not 

image
The Electron Structure title in results

edan-bainglass:
> should we also change this editor , image

Yes

> In the history there are quite some image

Oh boy, I hadn't looked at all the external notebooks. Let me check.

> And in the step 2 image
> 
> Metal and Insulator have the first letter with CAP, while in the protocol all are not

Well, those are not titles per say, just single keywords. Not sure what we want here. But I agree it should be consistent. Let me try first letter cap in protocol, see how that looks.

> image The Electron Structure title in results

Handled by #1039

edan-bainglass:
> > And in the step 2 image
> > Metal and Insulator have the first letter with CAP, while in the protocol all are not
> 
> Well, those are not titles per say, just single keywords. Not sure what we want here. But I agree it should be consistent. Let me try first letter cap in protocol, see how that looks.

image

@AndresOrtegaGuerrero what do you think?

AndresOrtegaGuerrero:
image
In the PDOS plugin, there is a section  `Recommendation for Molecules and Localized Orbitals:` I would also suggest including some vertical spacing for the text in this PR to make the text less cramped and more visually appealing.

edan-bainglass:
@AndresOrtegaGuerrero I removed the headers from PDOS. Extraneous in my opinion.

image

Please have a look at all plugin panels. Let me know if you spot anything else 🙂

AndresOrtegaGuerrero:
@edan-bainglass thank you everything looks nice. Perhaps i would suggest to highlight the words molecules and localized orbitals (I am not sure, but just in case they are visible to users). From my side the PR is ready

AndresOrtegaGuerrero (approved):
LGTM!

PR #1041, opened 2024-12-28
Use plugin utility to set default starting magnetization

### LINKED ISSUE #982: Set default starting magnetization or move to basic settings
If we run a magnetic calculation with no starting magnetization on any atom, the calculation will never be magnetic. Therefore:

- [ ] Two options:
   * if possible (it should, I think we have default rules in aiida-qe, check with @mbercx) we can leave the setting in the advanced ones, and set some value for magnetic elements. Ideally, use the function from aiida-qe as we might want to change defaults in the future after the results of MC3D
   * If not possible, the magnetization selections should appear when "On" is selected for magnetization.
- [ ] Show a text message near "magnetization" in the basic settings, when it's set to On, saying "Starting values for the initial magnetization have been guessed and set in the advanced settings. Please check them, especially if you are looking for specific magnetic configurations (only a tentative ferromagnetic configuration will be searched)."

I put here @superstar54 because of https://github.com/orgs/aiidalab/projects/17/views/1?pane=issue&itemId=88117658

edan-bainglass:
This PR uses the QE plugin `get_starting_magenetization` utility function to set default initial magnetic moments for the structure when magnetism is turned on. The defaults are set to the default magmom provided by the utility scaled by the Z valence value for the element defined in the selected pseudopotential. Note that for those elements of a zero magmom default, a general default magnetization of 0.1 is returned.

If multiple tags are provided for the same species, a warning is displayed providing some instructions for antiferromagnetic configurations.

Closes #982

---

Screenshot 2025-01-09 103706

Screenshot 2025-01-09 104547

Screenshot 2025-01-09 104615

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
@@ -81,8 +81,7 @@ class MagnetizationConfigurationSettingsModel(
                 content="""
                     If a nonzero ground-state magnetization is expected, you
                     <strong>must</strong> assign a nonzero value to at least one atomic
-                    type (note that the app might already provide tentative initial
-                    values to chemical elements that typically display magnetism).
+                    type (note that the app already provide tentative initial values).
                     To simulate an antiferromagnetic state, first, if you have not
                     done so already, please use the atom tag editor <b>(Select
                     structure -> Edit structure -> Edit atom tags)</b> to mark atoms of
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
@@ -2,7 +2,10 @@ from copy import deepcopy
 
 import traitlets as tl
 
+from aiida.common.exceptions import NotExistent
+from aiida_quantumespresso.workflows.protocols.utils import get_starting_magnetization
 from aiidalab_qe.common.mixins import HasInputStructure
+from aiidalab_qe.utils import fetch_pseudo_family_by_label
 
 from ..subsettings import AdvancedCalculationSubSettingsModel
 
@@ -17,10 +20,12 @@ class MagnetizationConfigurationSettingsModel(
         "input_structure",
         "electronic_type",
         "spin_type",
+        "pseudos.family",
     ]
 
     electronic_type = tl.Unicode()
     spin_type = tl.Unicode()
+    family = tl.Unicode()
 
     type_options = tl.List(
         trait=tl.List(tl.Unicode()),
@@ -41,9 +46,16 @@ class MagnetizationConfigurationSettingsModel(
         if self.spin_type == "none" or not self.has_structure:
             self._defaults["moments"] = {}
         else:
-            self._defaults["moments"] = {
-                kind_name: 0.0 for kind_name in self.input_structure.get_kind_names()
-            }
+            try:
+                self._defaults["moments"] = get_starting_magnetization(
+                    self.input_structure,
+                    fetch_pseudo_family_by_label(self.family),
+                )
+            except NotExistent:
+                self._defaults["moments"] = {
+                    kind_name: 0.0
+                    for kind_name in self.input_structure.get_kind_names()
+                }
         with self.hold_trait_notifications():
             self.moments = self._get_default_moments()
 
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
@@ -69,6 +69,7 @@ class MagnetizationConfigurationSettingsModel(
             self.moments = self._get_default_moments()
 
     def update_type_help(self):
+        """Update the type field help text w.r.t the current model state."""
         if self.electronic_type == "insulator" or self.type == "tot_magnetization":
             self.type_help = self._TYPE_HELP_TEMPLATE.format(
                 content="""
@@ -92,6 +93,9 @@ class MagnetizationConfigurationSettingsModel(
             )
 
     def update_default_starting_magnetization(self):
+        """Update the default starting magnetization based on the structure and
+        pseudopotential family.
+        """
         if not self.has_structure:
             # TODO this guard shouldn't be here! It IS here only because in the present
             # implementation, an update is called on app start. This breaks lazy loading
@@ -115,6 +119,7 @@ class MagnetizationConfigurationSettingsModel(
         }
 
     def _to_moment(self, symbol: str, family: PseudoPotentialFamily) -> float:
+        """Convert the default magnetization to an initial magnetic moment."""
         magnetization = (
             self._default_starting_magnetization.get(symbol, 0.1)
             if self._DEFAULT_MOMENTS.get(symbol, {}).get("magmom")
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 
 import traitlets as tl
+from aiida_pseudo.groups.family import PseudoPotentialFamily
 
 from aiida.common.exceptions import NotExistent
 from aiida_quantumespresso.workflows.protocols.utils import get_starting_magnetization
@@ -91,18 +92,23 @@ class MagnetizationConfigurationSettingsModel(
     def _update_default_moments(self):
         try:
             family = fetch_pseudo_family_by_label(self.family)
-            initial_guess = get_starting_magnetization(self.input_structure, family)
+            guess = get_starting_magnetization(self.input_structure, family)
             self._defaults["moments"] = {
-                kind.name: round(
-                    initial_guess[kind.name] * family.get_pseudo(kind.symbol).z_valence,
-                    3,
-                )
+                kind.name: self._to_moment(kind.symbol, guess[kind.name], family)
                 for kind in self.input_structure.kinds
             }
         except NotExistent:
             self._defaults["moments"] = {
                 kind_name: 0.0 for kind_name in self.input_structure.get_kind_names()
             }
+
+    @staticmethod
+    def _to_moment(
+        symbol: str,
+        magnetization: float,
+        family: PseudoPotentialFamily,
+    ):
+        return round(magnetization * family.get_pseudo(symbol).z_valence, 3)
 
     def _get_default_moments(self):
         return deepcopy(self._defaults.get("moments", {}))
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
@@ -30,11 +30,12 @@ class MagnetizationConfigurationSettingsModel(
     type_options = tl.List(
         trait=tl.List(tl.Unicode()),
         default_value=[
-            ["Magnetic moments", "starting_magnetization"],
+            ["Initial magnetic moments", "starting_magnetization"],
             ["Total magnetization", "tot_magnetization"],
         ],
     )
     type = tl.Unicode("starting_magnetization")
+    type_help = tl.Unicode("")
     total = tl.Float(0.0)
     moments = tl.Dict(
         key_trait=tl.Unicode(),  # kind name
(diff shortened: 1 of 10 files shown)


edan-bainglass:
@giovannipizzi please confirm the screenshots in the description follow what you had in mind 🙏

giovannipizzi:
Looks very good! We can refine in the future if needed, but already now is great

edan-bainglass:
@AndresOrtegaGuerrero haven't worked much with magnetic calculations. Was wondering how to handle total magnetization now. If starting magnetization is provided, does QE computes total magnetization internally, or do we need to explicitly provided it as well? If so, is it simply the sum of the starting magnetic moments? Note that at the moment, we don't have this implemented in the app. Total magnetization does not observe starting magnetic moments.

AndresOrtegaGuerrero:
In the starting_magnetization widget, we're actually passing the initial_magnetic_moments. (Perhaps we should consider renaming the widget for clarity.) These values are then passed to the workchain in the qe-plugin, which generates the starting_magnetization based on the provided magnetic moments.

```python
builder = PwBaseWorkChain.get_builder_from_protocol(
    structure=new_structure,
    code=pw_code,
    protocol='fast',
    spin_type=SpinType.COLLINEAR,
    initial_magnetic_moments=results['magnetic_moments']
)
```

The values you set in the starting_magnetization widget are treated as magnetic_moments, which are then used by the get_starting_magnetization function to calculate the starting_magnetization values in the input. You can verify this behavior with a test.

It’s important to understand the distinction between starting_magnetization and total_magnetization.

starting_magnetization: Defined per atomic kind, QE will calculate and output the final magnetization after convergence.
tot_magnetization: Specifies the imposed difference between spin-up and spin-down electrons.

For more details, refer to the QE documentation: [Quantum ESPRESSO Input Description](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm301).

AndresOrtegaGuerrero:
One possible solution for #982 could be to screen the dictionary from the starting_magnetization widget. If all values are 0.0, we should pass None. This would allow the PwBaseWorkChain to compute the starting_magnetization values (though this should be tested). If any value is different from 0.0, we pass the full dictionary instead. In this case, renaming the widget to initial_magnetic_moment might make the purpose clearer.

edan-bainglass:
@AndresOrtegaGuerrero will you be available on Monday to discuss? I have many questions!

AndresOrtegaGuerrero:
So we should rename the widget to magnetic moment, since that is the information we give the PwBaseWorkChain from Qe-plugin. 

I think we should have a text below the magnetization title that helps the user. 
When magnetic_moment is selected:

`If a nonzero ground-state magnetization is expected, you must assign a nonzero value to at least one atomic type. For antiferromagnetic configurations, specify both positive and negative magnetic moments for the chosen atomic types.`


When Tot_magnetization is selected the help text should say:

`The difference between majority spin charge and minority spin charge. Used to specify the desired total electronic magnetization.`

edan-bainglass:
> So we should rename the widget to magnetic moment, since that is the information we give the PwBaseWorkChain from Qe-plugin.
> 
> I think we should have a text below the magnetization title that helps the user. When magnetic_moment is selected:
> 
> `If a nonzero ground-state magnetization is expected, you must assign a nonzero value to at least one atomic type. For an antiferromagnetic state, define two distinct atomic species representing the sublattices of the same atomic type.`
> 
> When Tot_magnetization is selected the help text should say:
> 
> `The difference between majority spin charge and minority spin charge. Used to specify the desired total electronic magnetization.`

Thanks @AndresOrtegaGuerrero! I'm okay with the text. Maybe @giovannipizzi can review?

Regarding renaming of the widget, one question I had is how to handle remaining references to `starting_magnetization`. We use it in references to the parameters dictionary throughout. I could change this, but I'm concerned at its affect on backwards compatibility. Not sure you'd be able to load older calculations. Need to test.

AndresOrtegaGuerrero:
The units of the widget are Bohr magneton (meaning the number of unpaired electrons). QE plugin takes care of converting this value in the units expected in pw.x input (namely[ starting_magnetization](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm301)) a value between -1 and 1 (in pw.x not in the App) where 1 means that all electrons available for the selected species will be alpha electron. 

Si - Starting magnetization widget = 1

pw.x input
      starting_magnetization(1) =   2.5000000000d-01

edan-bainglass:
@AndresOrtegaGuerrero @cpignedoli in this PR, I'm using the plugin's `get_starting_magnetization` utility function to obtain the starting magnetization, then scale it by the valence provided in the selected pseudo to obtain the moments. Now, Carlo was wondering why 0.4 for Si. This is because `get_starting_magnetization` obtains the `magmom` for a given species using the plugin's `get_magnetization_parameters` utility.

```python
def get_magnetization_parameters() -> dict:
    """Return the mapping of suggested initial magnetic moments for each element.

    :returns: the magnetization parameters.
    """
    with (pathlib.Path(__file__).resolve().parent / 'magnetization.yaml').open() as handle:
        return yaml.safe_load(handle)
```

In "magnetization.yaml", Si's `magmom` is indeed 0. However, `get_starting_magnetization` assigns a general `default_magnetization: 0.1` (also in the yaml file) when the obtain `magmom` is 0. The valence of Si is 4, hence 0.4.

Is this correct logic?

cpignedoli:
Thanks a lot, @edan-bainglass for digging this out. From my use cases this is not ideal. If I am computing Si and for some reason I switch on magnetism I still  want to have 0 as suggested starting magnetization.

Moreover, 0.1 converted to 0.4 is additionally not clear to me:
If we consider Fe, the yaml file has magmom = 5, which means 5 electrons with spin up.
Thus in the QE app widget (Bohr magnetons as units) we should have 5 and this should appear in pw.inp as 5/(number of electrons available in the pseudopotential)

So in the case of Si or C, if I set 1 Bohr magneton in the QE app GUI I should see 0.25 in pw.inp

edan-bainglass:
> Thanks a lot, @edan-bainglass for digging this out. From my use cases this is not ideal. If I am computing Si and for some reason I switch on magnetism I still want to have 0 as suggested starting magnetization.

This is a bit difficult to do. I could apply logic to intercept magnetization of 0.1 and zero them out, but that would be assuming that no magmom divided by its pseudo-dependent Z valence value would somehow also hit 0.1, ending up with zero erroneously! Need to think about this some more.

> Moreover, 0.1 converted to 0.4 is additionally not clear to me: If we consider Fe, the yaml file has magmom = 5, which means 5 electrons with spin up. Thus in the QE app widget (Bohr magnetons as units) we should have 5 and this should appear in pw.inp as 5/(number of electrons available in the pseudopotential)
> 
> So in the case of Si or C, if I set 1 Bohr magneton in the QE app GUI I should see 0.25 in pw.inp

This is all correct and currently implemented. If magmom = 5 in the yaml file, `get_starting_magnetization` will return the scaled value. I "de-scale" it in the frontend, thus re-obtaining 5. When this is submitted, 5 will again go through that logic and will be rescaled to the correct magnetization. Bottom line, Si magmom of 1 in the app = Si 0.25 magnetization in the PW input.

edan-bainglass:
> > Thanks a lot, @edan-bainglass for digging this out. From my use cases this is not ideal. If I am computing Si and for some reason I switch on magnetism I still want to have 0 as suggested starting magnetization.
> 
> This is a bit difficult to do. I could apply logic to intercept magnetization of 0.1 and zero them out, but that would be assuming that no magmom divided by its pseudo-dependent Z valence value would somehow also hit 0.1, ending up with zero erroneously! Need to think about this some more.

Okay, I can solve this. Give me a few minutes!

edan-bainglass:
> > > Thanks a lot, @edan-bainglass for digging this out. From my use cases this is not ideal. If I am computing Si and for some reason I switch on magnetism I still want to have 0 as suggested starting magnetization.
> > 
> > 
> > This is a bit difficult to do. I could apply logic to intercept magnetization of 0.1 and zero them out, but that would be assuming that no magmom divided by its pseudo-dependent Z valence value would somehow also hit 0.1, ending up with zero erroneously! Need to think about this some more.
> 
> Okay, I can solve this. Give me a few minutes 🙏

Screenshot 2025-01-09 121643

Automatic defaults 👍 Note the correct 5 for Co and 0s for the rest. @cpignedoli expected?

edan-bainglass:
> Screenshot 2025-01-09 121643
> 
> Automatic defaults 👍 Note the correct 5 for Co and 0s for the rest. @cpignedoli expected?

But this I think leads to the same issue, no? If I choose bulk silicon and select magnetism, this "solution" will zero out silicon, thus leading to a non-magnetic calculation! @giovannipizzi @cpignedoli what should we do here? This solution is correct perhaps for the expert, but for beginners, this might be confusing.

cpignedoli:
Thanks a lot @edan-bainglass  This reflects what I have in mind. We should warn the user that 
starting magnetizations are set to ... (which convention has ben used the yaml file?) and the user should check the value and adapt to the know-how related to the specific system under investigation

giovannipizzi:
In the function that generates the initial moments, we indeed put 0.1 to ensure we break the "symmetry" and possibly detect any magnetism. If none should be there, the scf should bring it back to zero. This was discussed for quite some time, pinging @mbercx but also @Minotakm @t-reents

For this release, if @cpignedoli is OK, I'd keep the logic in the function (putting some small nonzero also for Si) and we rediscuss. @edan-bainglass this might need a slight change of the explanatory sentence discussed yesterday.

Also as a note, we had a report that setting  too high initial moments might make convergence harder and we'll investigate this in the future, so better to discuss this as part of aiida-qe to avoid having two different approaches

edan-bainglass:
> Thanks a lot @edan-bainglass This reflects what I have in mind. We should warn the user that starting magnetizations are set to ... (which convention has ben used the yaml file?) and the user should check the value and adapt to the know-how related to the specific system under investigation

To be clear, a warning is already in place, pushing users to set the magnetic configuration in the advanced settings. Are you suggesting something in addition to this warning? Note also that once they arrive at the widget in the advanced settings, a help message does instruct the user to set **non-zero* values to reach a nonzero ground-state magnetization. So I think we're covered?

cpignedoli:
sorry @edan-bainglass @giovannipizzi I was not aware of the previous discussions, if the 0.1 was put in place for a good reason, then let's keep it together with the appropriate warnings, I am fine with it. What is important is the consistency with units, the .yaml has the same units as we have in the QE app GUI so we are OK.

edan-bainglass:
> sorry @edan-bainglass @giovannipizzi I was not aware of the previous discussions, if the 0.1 was put in place for a good reason, then let's keep it together with the appropriate warnings, I am fine with it. What is important is the consistency with units, the .yaml has the same units as we have in the QE app GUI so we are OK.

Yes, as long as I apply the rescaling in general, the units are correct. Note that this means for those species of zero magmom, 0.1 magnetization will be rescaled into magmom depending on the Z valence of the element defined in the pseudo. This is how we end up with magmom 0.4 for Si. But it will return back to 0.1 magnetization in the PW input file. As long as we are okay with this, I can implement.

edan-bainglass:
> > sorry @edan-bainglass @giovannipizzi I was not aware of the previous discussions, if the 0.1 was put in place for a good reason, then let's keep it together with the appropriate warnings, I am fine with it. What is important is the consistency with units, the .yaml has the same units as we have in the QE app GUI so we are OK.
> 
> Yes, as long as I apply the rescaling in general, the units are correct. Note that this means for those species of zero magmom, 0.1 magnetization will be rescaled into magmom depending on the Z valence of the element defined in the pseudo. This is how we end up with magmom 0.4 for Si. But it will return back to 0.1 magnetization in the PW input file. As long as we are okay with this, I can implement.

Screenshot 2025-01-09 130831

@giovannipizzi what about the text needs to change?

src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
```diff
@@ -17,41 +23,105 @@ class MagnetizationConfigurationSettingsModel(
...
+            for kind in self.input_structure.kinds
+        }
+
+    def _to_moment(self, symbol: str, family: PseudoPotentialFamily) -> float:
+        magnetization = (
+            self._default_starting_magnetization.get(symbol, 0.1)
+            if self._DEFAULT_MOMENTS.get(symbol, {}).get("magmom")
+            else 0.1
+        )
+        return round(magnetization * family.get_pseudo(symbol).z_valence, 3)
```
AndresOrtegaGuerrero, line 124: I will add some docstring here , or comments about the meaning of the  "3" ,

edan-bainglass: Round to 3 digits? This is a built-in python function. I don't think it merits a comment. Docstring sure. I need to add docstrings in various places. But I don't plan on documenting built-in python functionality.

AndresOrtegaGuerrero: I didnt see the round , never mind! haha

giovannipizzi:
Simplify the first bracket we sentence, eg to: (note that the app already provides tentative initial values) as we set nonzero also for non magnetic species

edan-bainglass:
> Simplify the first bracket we sentence, eg to: (note that the app already provides tentative initial values) as we set nonzero also for non magnetic species

Oh I see. No "might", as we now 100% provide defaults. Got it 👍

AndresOrtegaGuerrero (approved):
LGTM! , thank you @edan-bainglass

Issue #1042, opened 2024-12-28
Improve results tab to make it more user-friendly


edan-bainglass:
- [x] Make the tree view hidden by default, e.g. in a tab "Advanced workflow tree view"
- [x] Replace with simpler tickbox interface showing what is not run yet and waiting for dependencies, what is running (orange emoji?) and what is finished, with a quick overview as well as clickable links to the relevant calculations. See suggestions here: ![Image](https://github.com/user-attachments/assets/390aa91f-0569-4fd3-9dab-15ac322760f1)
- [x] In the "Main settings", add information on the structure (Chemical formula, number of atoms, maybe even space group and cell vectors (a, b, c, alpha, beta, gamma)
- [x] In the title "QE App workflow (pk: 909)" replace instead simply with: `Workflow details` and, among the very first settings (or even better, before "Main settings", a first category "Workflow properties", with the following bullet points: `PK`, `UUID`, `Label`, `Description`, `Creation time: 12 May 2024 12:42 (3 months ago)` (with the suggested format before), and same for `Last modification time`.
- [x] In the advanced settings, write something near "Initial magnetic moments" if not set (e.g. "non-magnetic calculation" or "No initial moments set") - hidden if unset
---
Remember not to use any AiiDA language in the simple view (OK to use "Workflows" or "Calculations" or similar general terms, ideally defining internally what we mean with them so we use them consistently, but no WorkChain or PhononWorkChain or similar)

edan-bainglass:
@giovannipizzi @cpignedoli too simple?

![Image](https://github.com/user-attachments/assets/c96f67e6-0e6f-44bc-9d87-8491c506974a)

giovannipizzi:
I like it! Maybe we can hide the initial magnetic moments if unset, or write something so say that they were not set

edan-bainglass:
@giovannipizzi @cpignedoli regarding the simplified process tree...

https://github.com/user-attachments/assets/1994472e-bfed-4c9c-aa41-26254dec62d8

Feedback please 🙏

Note that one issue already is that the total jobs is currently `len(node.called)`, but that grows dynamically as you can see. I need some way to know in advance what jobs are planned to run. Thinking... input welcomed!

giovannipizzi:
[...]

PR #1044, opened 2024-12-28
Redesign summary view

edan-bainglass:
This PR is one in several that will address #1042

Here, I introduce a bullet list alternative format for the summary report (no actual bullets). The format is defined in `qeapp.yml` as `summary_format: list` and can be overridden by the user in a config file (see #1007 for info regarding file). `summary_format` takes `list` or `table`.

I also adjust the style such that the data download widget is to the right of the summary instead of below. This is true for either selected format.

Lastly, I add "Workflow properties" and "Structure properties" sections to the summary.

---

image

image


edan-bainglass:
If anyone has an older container (with an older AiiDA version, say, 2.5ish), can you please verify that you still get the notification regarding the unavailability of the raw data download feature?

AndresOrtegaGuerrero:
why dont you try with `from IPython.core.display import HTML` and create the tables , like that you have more flexibility in the appearance. 

Also since the tab of picking the settings was Basic settings, you should keep the same instead of using Main settings

edan-bainglass:
> why dont you try with `from IPython.core.display import HTML` and create the tables , like that you have more flexibility in the appearance.

No idea what you're referring to 😅

> Also since the tab of picking the settings was Basic settings, you should keep the same instead of using Main settings

This was a specific request by Giovanni. See issue.

AndresOrtegaGuerrero:
> > why dont you try with `from IPython.core.display import HTML` and create the tables , like that you have more flexibility in the appearance.
> 
> No idea what you're referring to 😅

You can make an HTML object and create a table and to display it , it has better visuals than the ipywidget.HTML. 

The table gave this kind of separation that was nice , plus a highlight , but indeed they were to big the cell size. 

> > Also since the tab of picking the settings was Basic settings, you should keep the same instead of using Main settings
> 
> This was a specific request by Giovanni. See issue.
Ok , then we should keep it consistent in the configuration step , and name it main settings as well (instead of basic)

AndresOrtegaGuerrero:
@edan-bainglass I can confirm you that the notification for aiida 2.5 works 
image

edan-bainglass:
Thanks @AndresOrtegaGuerrero 🙏

edan-bainglass:
@giovannipizzi @cpignedoli alternative...

image

Both `list` and `table` are now valid choices for a new `summary_format` configuration key, which thanks to #1007, is now user-configurable. Currently defaulting to `list`, though the table view is quite clean in my opinion (suggested by @AndresOrtegaGuerrero).

Also, "Main settings" is inconsistent to "Basic settings". Best we uniform this. Do you agree, and if so, preference?

giovannipizzi:
Thanks! Both are good, table is OK as the dark color is not too dark/prominent. But as the list might get longer, maybe the list is OK. I agree with uniforming basic/main,maybe "main" is better but both are are ok.

edan-bainglass:
> Thanks! Both are good, table is OK as the dark color is not too dark/prominent. But as the list might get longer, maybe the list is OK. I agree with uniforming basic/main,maybe "main" is better but both are are ok.

Okay then. Leaving `list` as the default. And I think I'll stick to "basic".

edan-bainglass:
Adding more info requested in #1042

image

and

image

AndresOrtegaGuerrero:
I would suggest omitting the "Workflow Properties" section. If the concern was that the current display is distracting, this version is even more so.

Displaying the creation and modification times, as well as the UUID, seems unnecessary. Also, isn't the PK number already shown at the top?

Instead of adding a separate section, I think we could include the PK, label, and description just above the "Workflow completed successfully" message.

image

AndresOrtegaGuerrero:
And we rather have the Structure "properties" in the results tab Initial Structure 

image

edan-bainglass:
@AndresOrtegaGuerrero could you please review once more? 🙏

AndresOrtegaGuerrero:
@edan-bainglass Thank you! In my test everything works. I did notice an issue I mentioned in #1045, where we sometimes fail to indicate whether the full workflow completed successfully. However, I’m thinking this might not be the right place to address it—what do you think?

AndresOrtegaGuerrero:
Just dont forget to solve the conflict

AndresOrtegaGuerrero (approved):
LGTM! Thank you @edan-bainglass

AndresOrtegaGuerrero:
@edan-bainglass It just crossed my mind now, should we report the properties computed ?

edan-bainglass:
> @edan-bainglass Thank you! In my test everything works. I did notice an issue I mentioned in #1045, where we sometimes fail to indicate whether the full workflow completed successfully. However, I’m thinking this might not be the right place to address it—what do you think?

Indeed, unrelated. Thanks for opening the issue.

edan-bainglass:
> @edan-bainglass It just crossed my mind now, should we report the properties computed ? 

I suppose so? Doesn't hurt. I'll add it 👍

edan-bainglass:
> > @edan-bainglass It just crossed my mind now, should we report the properties computed ?
> 
> I suppose so? Doesn't hurt. I'll add it 👍

@AndresOrtegaGuerrero not sure where and how to best represent the computed properties. For now, I'll merge without it. Let's discuss and PR this later.

PR #1047, opened 2025-01-02
Check for null exit status

edan-bainglass:
Only escaped if exit status was 0, but need to escape if it is `None`

### DIFF
--- a/src/aiidalab_qe/app/result/components/summary/model.py
+++ b/src/aiidalab_qe/app/result/components/summary/model.py
@@ -131,9 +131,8 @@ class WorkChainSummaryModel(ResultsComponentModel):
 
     def generate_failure_report(self):
         """Generate a html for reporting the failure of the `QeAppWorkChain`."""
-        if not (process_node := self.fetch_process_node()):
-            return
-        if process_node.exit_status == 0:
+        process_node = self.fetch_process_node()
+        if not process_node and process_node.exit_status:
             return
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/app/result/components/summary/model.py
```diff
@@ -133,7 +133,7 @@ def generate_failure_report(self):
         """Generate a html for reporting the failure of the `QeAppWorkChain`."""
         if not (process_node := self.fetch_process_node()):
```
AndresOrtegaGuerrero, line 134: suggestion ,

```python
if not process_node or not process_node.exit_status:  
    return
```

edan-bainglass:
@AndresOrtegaGuerrero failing tests unrelated. Something about a deprecation that took place 01.01.2025. Will open an issue.

AndresOrtegaGuerrero:
Could you suggest how to test this ?

edan-bainglass:
> Could you suggest how to test this ?

I'm assuming you refer to the PR, not the failed tests. To test, you could simply remove the check and run a calculation. I encountered the issue when running my typical test workflow (relax positions + bands + pdos).

edan-bainglass:
@AndresOrtegaGuerrero I'll resolve #1053 here. See description.

AndresOrtegaGuerrero (approved):
LGTM!

PR #1051, opened 2025-01-03
Add initial structure pk to summary

### LINKED ISSUE #1049: Summary missing information (Input Structure PK)
When displaying the workflow summary, the PK of the input structure should be shown under Structure Properties. This helps users identify which PK to use for subsequent calculations involving the same structure.

Additionally, it should be clearly defined whether Structure Properties refers to the input structure or the relaxed structure.

Displaying the PK is crucial because, after the workflow finishes, it can be challenging to locate, especially since it is not available in the Input Structure tab.

image

or even if you go to the step 1 to see the structure , 
image

edan-bainglass:
Closes #1049

image

### DIFF
--- a/src/aiidalab_qe/app/result/components/summary/model.py
+++ b/src/aiidalab_qe/app/result/components/summary/model.py
@@ -177,6 +177,7 @@ class WorkChainSummaryModel(ResultsComponentModel):
             "creation_time_relative": relative_time(qeapp_wc.ctime),
             "modification_time": format_time(qeapp_wc.mtime),
             "modification_time_relative": relative_time(qeapp_wc.mtime),
+            "structure_pk": initial_structure.pk,
             "formula": initial_structure.get_formula(),
             "num_atoms": len(initial_structure.sites),
             "space_group": "{} ({})".format(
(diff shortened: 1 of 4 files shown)


edan-bainglass:
@AndresOrtegaGuerrero please review this one next 🙏

AndresOrtegaGuerrero:
I did the test of the pk, it shows the pk  , but now i am experiencing a failure 

image

which is weird since my dft calculation didnt failed, 

image

AndresOrtegaGuerrero:
I have a feeling it was introduce by #1047

edan-bainglass:
> I have a feeling it was introduce by #1047 , this is only occuring with the external plugins/

Indeed. In

```python
if not process_node and process_node.exit_status:
    return
```

it's evaluating `(not A) and B` instead of the intended `not (A and B)`. Should've added parentheses. Will fix in this PR.

edan-bainglass:
@AndresOrtegaGuerrero please retry 🙏

AndresOrtegaGuerrero:
I just tested it is fine now ! thank you Edan

AndresOrtegaGuerrero (approved):
LGTM!

PR #1054, opened 2025-01-05
Generalize summary generation

edan-bainglass:
This PR aims to simplify workflow summary maintenance by generalizing the template and introducing a schema. Adding new fields reduces to adding the field in the report dictionary and schema.


edan-bainglass:
@AndresOrtegaGuerrero this one's next. Let me fix the tests first.

AndresOrtegaGuerrero:
I would suggest to add an extra test, that it doesnt consider the default values,  to check changes are being used. I would say that we should try to merge first #1055 first since that one is causing the app to not work with different periodicity systems.

edan-bainglass:
> I would suggest to add an extra test, that it doesnt consider the default values, to check changes are being used. I would say that we should try to merge first #1055 first since that one is causing the app to not work with different periodicity systems.

I suspect merging #1055 first will cause a massive conflict headache with this one. But I can adjust if you think it is absolutely required.

AndresOrtegaGuerrero:
We can then try this one first , and you add more test in #1055 ?

AndresOrtegaGuerrero (approved):
LGTM!

The test with DFT+U , spin and XYZ periodicity works well. However, we can continue testing in the other PR, focusing on the point-group, and add more tests in that PR to further generalize different options.

PR #1056, opened 2025-01-05
Adding a download image button for different formats

AndresOrtegaGuerrero:
Here we include a button for downloading the image in different formats 

image


src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
```diff
@@ -120,6 +120,22 @@ def render(self):
...
+        self.download_image = ipw.Button(
+            description="Download image",
+            button_style="primary",
+            icon="fa-image",
+        )
+        self.download_image.on_click(self._model.download_image)
+        self.image_format = ipw.Dropdown(
+            description="Format:",
+            disabled=False,
+        )
```
edan-bainglass, line 132: `disabled` is `False` by default. Is there a reason you set it here explicitly? I've seen this done throughout the code in the past and have been wondering the reason.

src/aiidalab_qe/common/bands_pdos/model.py
```diff
@@ -284,6 +290,41 @@ def update_trace_color(self, color):
...
+            var link = document.createElement('a');
+            link.href = 'data:image/{format};base64,{payload}';
+            link.download = "{filename}";
+            document.body.appendChild(link);
+            link.click();
+            document.body.removeChild(link);
+            """.format(
+                payload=payload, filename=filename, format=filename.split(".")[-1]
+            )
+        )
```
edan-bainglass, line 325: I'm wary of injecting JS. Tends to complicate things (note that I also do it at times). Can you say a bit more regarding the need for JS here?

edan-bainglass (commented):
Overall straightforward. Before I approve, can you please respond to my comments?

src/aiidalab_qe/common/bands_pdos/model.py
AndresOrtegaGuerrero: so the file is in the user compute an not in the container

src/aiidalab_qe/common/bands_pdos/bandpdoswidget.py
AndresOrtegaGuerrero: i will remove it , just a sec , i think it just got there while writing in vs code

src/aiidalab_qe/common/bands_pdos/model.py
edan-bainglass: Not sure I understand. Which file? What are you downloading?

AndresOrtegaGuerrero: an image, the pr is to add a button that download an image of the plot in different formats

edan-bainglass: I'm sorry, I still don't understand. Is the image not just the plot, which is in the frontend, which is in the container?

AndresOrtegaGuerrero: The JavaScript is necessary to download the image file directly to my local directory. Without it, saving the file would place it inside the AiiDAlab container, making it inaccessible from my local system.

edan-bainglass (approved):
LGTM! Thanks @AndresOrtegaGuerrero. You can proceed. If possible, could you maybe comment on Plotly's image download API? The plot comes with built-in PNG download. I'm wondering if the modebar could be modified (in a future PR) to include custom buttons for other common formats (likely still using JS).

AndresOrtegaGuerrero:
I used https://plotly.com/python-api-reference/plotly.io.html?highlight=io#module-plotly.io , i dont know if the other option can be modified.

PR #1062, opened 2025-01-07
Adding information optimized structure in results

### LINKED ISSUE #928: RelaxWorkChain Result Tab should display the PK of optimized structure
After running a RelaxWorkChain , the tab doesn't tell the user what is the PK of the optimized structure.

The app should also add a description to the StructureData like "optimized structure" so later it can facilitate the search of such structures, and to let the users know that this is StructureData is from a RelaxWorkChain.

AndresOrtegaGuerrero:
This PR closes #928 
The displayed information includes the PK, label, description, creation time, and number of atoms. 

Structure Creation Details: Knowing when the structure was created and its PK helps users efficiently search for and reuse it in new calculations. The PK also helps to locate the structure within the database.

Interaction with Selection Widget: The number of atoms reminds the user the structure's size (on the tab), enabling them to interact effectively with the selection widget.

Screenshot 2025-01-07 at 21 45 14

### DIFF
--- a/src/aiidalab_qe/app/result/components/viewer/structure/structure.py
+++ b/src/aiidalab_qe/app/result/components/viewer/structure/structure.py
@@ -27,10 +27,7 @@ class StructureResultsPanel(ResultsPanel[StructureResultsModel]):
             """)
             self.atom_coordinates_table = TableWidget()
             self._generate_table(structure.get_ase())
-            self.atom_coordinates_table.observe(self._change_selection, "selected_rows")
-            # Listen for changes in self.widget.displayed_selection and update the table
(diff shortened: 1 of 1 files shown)


edan-bainglass:
@AndresOrtegaGuerrero same comment regarding the header. Redundant in my opinion.

AndresOrtegaGuerrero:
@edan-bainglass thank you , should i just remove it right ? or replace it for other text ?

edan-bainglass:
> @edan-bainglass thank you , should i just remove it right ? or replace it for other text ?

Discard in my opinion

PR #1067, opened 2025-01-08
Synchronize TableWidget row selection

AndresOrtegaGuerrero:
This PR addresses the issue of inconsistent row selection states in the table , in particular when the widget clear a selection in the table , but if you click the table it might retain extra information of the selection. 

An updateSelection function is included to synchronize the selectedIndices array in JavaScript with the selected_rows list from Python

Old

https://github.com/user-attachments/assets/8106b48f-f547-4e41-babc-989715613860

New 

https://github.com/user-attachments/assets/7028ae97-47ea-4903-9158-d4bac43dfd97

### DIFF
--- a/src/aiidalab_qe/common/widgets.py
+++ b/src/aiidalab_qe/common/widgets.py
@@ -1235,68 +1235,70 @@ class QeDependentWizardStep(QeWizardStep[QWSM]):
 class TableWidget(anywidget.AnyWidget):
     _esm = """
     function render({ model, el }) {
-      let domElement = document.createElement("div");
-      el.classList.add("custom-table");
-      let selectedIndices = [];
+        let domElement = document.createElement("div");
+        el.classList.add("custom-table");
+        let selectedIndices = [];
 
-      function drawTable() {
-          const data = model.get("data");
-          domElement.innerHTML = "";
-          let innerHTML = '<table><tr>' + data[0].map(header => `<th>${header}</th>`).join('') + '</tr>';
+        function drawTable() {
+            const data = model.get("data");
+            domElement.innerHTML = "";
+            let innerHTML = '<table><tr>' + data[0].map(header => `<th>${header}</th>`).join('') + '</tr>';
 
-          for (let i = 1; i < data.length; i++) {
-              innerHTML += '<tr>' + data[i].map(cell => `<td>${cell}</td>`).join('') + '</tr>';
-          }
+            for (let i = 1; i < data.length; i++) {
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/common/widgets.py
```diff
@@ -1235,68 +1235,70 @@ def _hide_missing_information_warning(self):
...
+                        row.classList.remove('selected-row');
+                    }
+                }
+            });
+        }
+
+        drawTable();
+        model.on("change:data", drawTable);
+        model.on("change:selected_rows", updateSelection);
+        el.appendChild(domElement);
```
edan-bainglass, line 1301: I don't understand, why is it marking the entire code as a change? I can't tell what you changed 🤷🏻‍♂️ That said, from the behavior in the clip, seems like you solved it. But maybe you can say a bit what happened?

edan-bainglass (commented):
Can't tell what changed 🤷🏻‍♂️

src/aiidalab_qe/common/widgets.py
AndresOrtegaGuerrero: i was doing a test in a jupyter notebook , then i guess in the copy paste it moves the spaces , so the issue is that I added 
```JS
function updateSelection() {
    const newSelection = model.get("selected_rows");
    selectedIndices = [...newSelection]; 
    const rows = domElement.querySelectorAll('tr');
    rows.forEach((row, index) => {
        if (index > 0) {
            if (selectedIndices.includes(index - 1)) {
                row.classList.add('selected-row');
            } else {
                row.classList.remove('selected-row');
            }
        }
    });
}
```
So when i assigned selected_row in python it is reflected in the JS, because before the selection was keeping the previous state

AndresOrtegaGuerrero: and to listen this changes i added , so is i set to selected_rows in Python it will triggers the function updateSelection

``
model.on("change:selected_rows", updateSelection);
``

edan-bainglass (approved):
LGTM!

superstar54:
Hi @AndresOrtegaGuerrero and @edan-bainglass , for the `calculation history` page, I created a new table widget (https://github.com/superstar54/table-widget) today, it supports interactive features (editing, filtering, search), and I tried to make the table as general as possible. Will make a PR for the calculation history soon. 

@AndresOrtegaGuerrero, please have a look to see if it can also be used here.

edan-bainglass:
Very nice! Performance okay?

superstar54:
> Performance okay?

React, and [Datagrid](https://www.datagrid.com/) libraries are used, making it smooth, even for large data.

PR #1075, opened 2025-01-12
Spruce up advanced panel styling

edan-bainglass:
This PR is mostly cosmetic, applying the following styles to the advanced settings panel:

- Vertically align widgets
- Add more sub-section headings and emphasize style (`h2` & `h4` tags)
- Make widget labels clear and concise
- Add more help text (thanks @AndresOrtegaGuerrero 🙏)
- Reduce help text overflow where appropriate
- Move units from label to right of input widgets (using new `HBoxWithUnits` widget)
- Move Hubbard U check box down
- Covert functional selector to toggle buttons
- Reduce verbosity in pseudopotential widgets by moving some description up into help text
- Add link for more info on pseudopotential cutoffs (thanks @AndresOrtegaGuerrero 🙏)

---

image

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced/advanced.py
@@ -73,9 +73,16 @@ class AdvancedConfigurationSettingsPanel(
         if self.rendered:
             return
 
-        # clean-up workchain settings
+        self.reset_to_defaults_button = ipw.Button(
+            description="Reset to defaults",
+            button_style="primary",
+            icon="undo",
+            layout=ipw.Layout(width="fit-content"),
+        )
+        self.reset_to_defaults_button.on_click(self._on_reset_to_defaults_button_click)
+
         self.clean_workdir = ipw.Checkbox(
-            description="Tick to delete the work directory after the calculation is finished",
+            description="Delete the work directory after the calculation",
             indent=False,
             layout=ipw.Layout(width="fit-content"),
         )
@@ -83,13 +90,6 @@ class AdvancedConfigurationSettingsPanel(
             (self._model, "clean_workdir"),
             (self.clean_workdir, "value"),
         )
-        self.reset_to_defaults_button = ipw.Button(
-            description="Reset to defaults",
-            button_style="warning",
-            icon="undo",
-            layout=ipw.Layout(width="fit-content"),
-        )
-        self.reset_to_defaults_button.on_click(self._on_reset_to_defaults_button_click)
 
         # Smearing setting widget
         self.smearing.render()
@@ -98,8 +98,8 @@ class AdvancedConfigurationSettingsPanel(
         self.kpoints_distance = ipw.BoundedFloatText(
             min=0.0,
             step=0.05,
-            description="K-points distance (1/Å):",
-            style={"description_width": "initial"},
+            description="K-points distance:",
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "kpoints_distance"),
@@ -110,7 +110,7 @@ class AdvancedConfigurationSettingsPanel(
             (self.kpoints_distance, "disabled"),
             lambda _: not self._model.has_pbc,
         )
-        self.mesh_grid = ipw.HTML()
+        self.mesh_grid = ipw.HTML(layout=ipw.Layout(margin="0 0 0 10px"))
         ipw.dlink(
             (self._model, "mesh_grid"),
             (self.mesh_grid, "value"),
@@ -125,7 +125,7 @@ class AdvancedConfigurationSettingsPanel(
             max=3,
             step=0.01,
             description="Total charge:",
-            style={"description_width": "initial"},
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "total_charge"),
@@ -135,7 +135,7 @@ class AdvancedConfigurationSettingsPanel(
         # Van der Waals setting widget
         self.van_der_waals = ipw.Dropdown(
             description="Van der Waals correction:",
-            style={"description_width": "initial"},
+            style={"description_width": "150px"},
         )
         ipw.dlink(
             (self._model, "van_der_waals_options"),
@@ -153,8 +153,9 @@ class AdvancedConfigurationSettingsPanel(
         self.scf_conv_thr = ipw.BoundedFloatText(
             min=1e-15,
             max=1.0,
-            description="SCF conv.:",
-            style={"description_width": "initial"},
+            format="0.0e",
+            description="SCF:",
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "scf_conv_thr"),
@@ -167,8 +168,9 @@ class AdvancedConfigurationSettingsPanel(
         self.forc_conv_thr = ipw.BoundedFloatText(
             min=1e-15,
             max=1.0,
-            description="Force conv.:",
-            style={"description_width": "initial"},
+            format="0.0e",
+            description="Force:",
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "forc_conv_thr"),
@@ -181,8 +183,9 @@ class AdvancedConfigurationSettingsPanel(
         self.etot_conv_thr = ipw.BoundedFloatText(
             min=1e-15,
             max=1.0,
-            description="Energy conv.:",
-            style={"description_width": "initial"},
+            format="0.0e",
+            description="Energy:",
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "etot_conv_thr"),
@@ -196,8 +199,8 @@ class AdvancedConfigurationSettingsPanel(
             min=20,
             max=1000,
             step=1,
-            description="Max. electron steps:",
-            style={"description_width": "initial"},
+            description="Electronic:",
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "electron_maxstep"),
@@ -208,8 +211,8 @@ class AdvancedConfigurationSettingsPanel(
             min=50,
             max=1000,
             step=1,
-            description="Max. optimization steps:",
-            style={"description_width": "initial"},
+            description="Ionic:",
+            style={"description_width": "150px"},
         )
         ipw.link(
             (self._model, "optimization_maxsteps"),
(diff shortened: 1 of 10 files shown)


src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -208,8 +211,8 @@ def render(self):
             min=50,
             max=1000,
             step=1,
-            description="Max. optimization steps:",
-            style={"description_width": "initial"},
+            description="Ionic:",
```
AndresOrtegaGuerrero, line 214: I would be a lil keep optimization , or relaxation since for people with no experience with DFT wont understand right away.

src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -196,8 +199,8 @@ def render(self):
             min=20,
             max=1000,
             step=1,
-            description="Max. electron steps:",
-            style={"description_width": "initial"},
+            description="Electronic:",
```
AndresOrtegaGuerrero, line 202: I would also keep the word steps, like that is clear that is an integer.

AndresOrtegaGuerrero: and i guess we should add the units ,  (a.u) ,

src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -167,8 +168,9 @@ def render(self):
         self.forc_conv_thr = ipw.BoundedFloatText(
             min=1e-15,
             max=1.0,
-            description="Force conv.:",
-            style={"description_width": "initial"},
+            format="0.0e",
+            description="Force:",
```
AndresOrtegaGuerrero, line 172: here i would add the units , (a.u)

AndresOrtegaGuerrero:
thank you @edan-bainglass , I would suggest that we group the "Convergence thresholds" and "Maximum cycles steps" .or  
It could also help a text explaning that these parameter control the accuracy of your simulation (In the case of the convergence thresholds"  , and some text about the cycles steps, since most people dont know DFT so they wont understand what these cycles mean

AndresOrtegaGuerrero:
maybe something below smearing "smearing: we make the occupations continuous near the Fermi level as if the temperature is finite. This is important for metallic systems" ? since the k-point has a mini definition

src/aiidalab_qe/app/configuration/advanced/pseudos/pseudos.py
```diff
@@ -152,40 +137,62 @@ def render(self):
...
+                </div>
+            """),  # noqa: RUF001
             self.setter_widget,
-            self.cutoff_helper,
+            ipw.HTML("<b>Cutoffs</b>"),
+            ipw.HTML("""
+                <div style="line-height: 1.4;">
+                    The cutoffs used for the calculation are the maximum of the
+                    default cutoffs from all pseudopotentials.
+                    You can override them here.
```
AndresOrtegaGuerrero, line 180: What about we add a link like , for more information about the cutoffs , and put a link to , https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm814:~:text=to%20Top%5D-,ecutrho,-REAL , ?

src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -153,8 +153,9 @@ def render(self):
         self.scf_conv_thr = ipw.BoundedFloatText(
             min=1e-15,
             max=1.0,
-            description="SCF conv.:",
-            style={"description_width": "initial"},
+            format="0.0e",
+            description="SCF:",
```
AndresOrtegaGuerrero, line 157: what about , Self-consistent field (SCF)

src/aiidalab_qe/app/configuration/advanced/advanced.py
edan-bainglass: I don't know what "I would be a lil keep optimization" means

src/aiidalab_qe/app/configuration/advanced/advanced.py
edan-bainglass: It was SCF before. Let's keep it as is for now.

edan-bainglass:
> thank you @edan-bainglass , I would suggest that we group the "Convergence thresholds" and "Maximum cycles steps" .or It could also help a text explaning that these parameter control the accuracy of your simulation (In the case of the convergence thresholds" , and some text about the cycles steps, since most people dont know DFT so they wont understand what these cycles mean

I agree with help text (please provide a draft), but disagree with grouping. Please explain why.

edan-bainglass:
> maybe something below smearing "smearing: we make the occupations continuous near the Fermi level as if the temperature is finite. This is important for metallic systems" ? since the k-point has a mini definition

Not sure I understand

src/aiidalab_qe/app/configuration/advanced/advanced.py
AndresOrtegaGuerrero: Yes, it was like that before, but I wonder if it's clear enough for users who are not familiar with DFT.

AndresOrtegaGuerrero:
> > thank you @edan-bainglass , I would suggest that we group the "Convergence thresholds" and "Maximum cycles steps" .or It could also help a text explaning that these parameter control the accuracy of your simulation (In the case of the convergence thresholds" , and some text about the cycles steps, since most people dont know DFT so they wont understand what these cycles mean
> 
> I agree with help text (please provide a draft), but disagree with grouping. Please explain why.

I just feel there's too much space, and these variables also control the convergence of the calculation.

AndresOrtegaGuerrero:
> > maybe something below smearing "smearing: we make the occupations continuous near the Fermi level as if the temperature is finite. This is important for metallic systems" ? since the k-point has a mini definition
> 
> Not sure I understand

We should add a description for smearing , something like "Smearing: Makes occupations near the Fermi level continuous, simulating finite temperature effects, essential for accurate calculations in metallic systems"

The k-point widget has a description so i feel the smearing should have something as well

src/aiidalab_qe/app/configuration/advanced/advanced.py
AndresOrtegaGuerrero: I would keep the term "optimization" or "relaxation" in the description rather than "ionic," as not all users may be familiar with.

AndresOrtegaGuerrero:
> > thank you @edan-bainglass , I would suggest that we group the "Convergence thresholds" and "Maximum cycles steps" .or It could also help a text explaning that these parameter control the accuracy of your simulation (In the case of the convergence thresholds" , and some text about the cycles steps, since most people dont know DFT so they wont understand what these cycles mean
> 
> I agree with help text (please provide a draft), but disagree with grouping. Please explain why.

what about , Setting convergence thresholds for energy, forces, and self-consistency ensures calculation accuracy and stability by determining when the self-consistent cycle is complete.

AndresOrtegaGuerrero:
For the eigenvalues (in hubbard widget), text i would add 

For transition metals and lanthanoids, the starting eigenvalues can be defined (Magnetic calculation).
"This is useful to suggest the desired orbital occupations when the default choice takes another path."

AndresOrtegaGuerrero:
Why separate the smearing widget vertically (type and width) instead of keeping them horizontal? Since they relate to the same property, a horizontal layout might be more intuitive. Additionally, it would align with the aesthetic you proposed, as functional and pseudopotentials are also organized horizontally.

edan-bainglass:
> Why separate the smearing widget vertically (type and width) instead of keeping them horizontal? Since they relate to the same property, a horizontal layout might be more intuitive. Additionally, it would align with the aesthetic you proposed, as functional and pseudopotentials are also organized horizontally.

Are they?

edan-bainglass:
@AndresOrtegaGuerrero the reason for this PR is mostly due to what I perceive to be a hectic panel. Everything was cramped, nothing aligned. I'm trying here to provide some uniform sense to the page, which does help psychologically.

Regarding the abundance of white space. I agree. However, in small screens, it is not there. Also, in small screens, the present state of the app is unreadable due to the horizontal layout. With this PR, at least this panel is readable. That said, the ideal solution would be a standard use of CSS breakpoints to switch between clean horizontal layout in wide screens and clean vertical layout in small screens.

This PR does not handle this yet. A future PR could easily build on this to do just that.

edan-bainglass:
@AndresOrtegaGuerrero also, regarding your text suggestions, that is not the purpose of this PR. You can open a separate follow-up PR to add more help text.

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero also, regarding your text suggestions, that is not the purpose of this PR. You can open a separate follow-up PR to add more help text.

Sure, take the ones you agree and discards what it is not related. For me was pertinent since it affect the visuals and could be addresses here, but lets leave it as you said for another PR.

AndresOrtegaGuerrero:
> > Why separate the smearing widget vertically (type and width) instead of keeping them horizontal? Since they relate to the same property, a horizontal layout might be more intuitive. Additionally, it would align with the aesthetic you proposed, as functional and pseudopotentials are also organized horizontally.
> 
> Are they?

Yeah, you apply Smearing width (Ry) value , you just select the function (type)

edan-bainglass:
> > > Why separate the smearing widget vertically (type and width) instead of keeping them horizontal? Since they relate to the same property, a horizontal layout might be more intuitive. Additionally, it would align with the aesthetic you proposed, as functional and pseudopotentials are also organized horizontally.
> > 
> > 
> > Are they?
> 
> Yeah, you apply Smearing width (Ry) value , you just select the function (type)

Sorry, I meant are the functional and pseudopotentials really organized horizontally? They've also been vertically aligned.

AndresOrtegaGuerrero:
> > > > Why separate the smearing widget vertically (type and width) instead of keeping them horizontal? Since they relate to the same property, a horizontal layout might be more intuitive. Additionally, it would align with the aesthetic you proposed, as functional and pseudopotentials are also organized horizontally.
> > > 
> > > 
> > > Are they?
> > 
> > 
> > Yeah, you apply Smearing width (Ry) value , you just select the function (type)
> 
> Sorry, I meant are the functional and pseudopotentials really organized horizontally? They've also been vertically aligned.

Ah, in the figure, the options are arranged horizontally.

AndresOrtegaGuerrero (approved):
LGTM!

PR #1076, opened 2025-01-12
Shuffle results panel top-level tabs and update default selection

edan-bainglass:
This PR does the following:
1. Reorder results tabs as Summary | Status | Results
2. Sets the default tab dynamically (Status if running, Results if complete)

### DIFF
--- a/src/aiidalab_qe/app/result/__init__.py
+++ b/src/aiidalab_qe/app/result/__init__.py
@@ -81,8 +81,8 @@ class ViewQeAppWorkChainStatusAndResultsStep(QeDependentWizardStep[ResultsStepMo
 
         self.panels = {
             "Summary": self.summary_panel,
-            "Results": self.results_panel,
             "Status": self.status_panel,
+            "Results": self.results_panel,
         }
 
         self.toggle_controls = ipw.ToggleButtons(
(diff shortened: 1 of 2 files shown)


src/aiidalab_qe/app/result/__init__.py
```diff
@@ -118,7 +118,11 @@ def _post_render(self):
         self._update_kill_button_layout()
         self._update_clean_scratch_button_layout()
 
-        self.toggle_controls.value = "Summary"
+        self.toggle_controls.value = (
+            "Results"
+            if (process := self._model.fetch_process_node()) and process.is_finished_ok
```
AndresOrtegaGuerrero, line 123: If a process has multiple properties and one fails, would this prevent the visualization of the results for the properties that completed successfully?

edan-bainglass: I don't see why it would. This is just defining the default selected tab. All three tabs are still there.

AndresOrtegaGuerrero: because if one fail, process.is_finished_ok will be false , right ?

edan-bainglass: The default in this case is Status and the app will direct the user there. But the other panels are still accessible.

AndresOrtegaGuerrero (approved):
LGTM!

PR #1078, opened 2025-01-13
Group bands by angular momentum by default

edan-bainglass:
Requested by Giovanni

### DIFF
--- a/src/aiidalab_qe/common/bands_pdos/model.py
+++ b/src/aiidalab_qe/common/bands_pdos/model.py
@@ -47,7 +47,7 @@ class BandsPdosModel(Model):
             ("No grouping (each orbital separately)", "orbital"),
         ],
     )
-    dos_plot_group = tl.Unicode("total")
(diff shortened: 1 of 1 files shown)


AndresOrtegaGuerrero (approved):
I wonder why one test fail ...

edan-bainglass:
Need to check more closely, but I think it may be an issue with an assumption in the test.

edan-bainglass:
> I wonder why one test fail ...

I guess a backwards compatibility issue. Python 3.9 + AiiDA v2.3 tests failing (2.6 passing).

```python
if plot_tag != "total":
        if not pdos.nscf.output_parameters.get("spin_orbit_calculation"):
            data_dict = _update_pdos_labels(data_dict)
```

Something about the use of `pdos.nscf.output_parameters.get` is an issue. Unrelated to this PR (though the change appears to have found it, suggesting that the plotter is broken on older AiiDA instances). Proceeding. Will open an issue.

PR #1082, opened 2025-01-13
Fix rendering logic of results panel controls

edan-bainglass:
Thie PR fixes incorrect logic in deciding which widgets to add to a results panel that was incorrectly rendering status notification for completed workflows.

### DIFF
--- a/src/aiidalab_qe/common/panel.py
+++ b/src/aiidalab_qe/common/panel.py
@@ -635,11 +635,14 @@ class ResultsPanel(Panel[RM]):
             ]
             self._load_results()
         else:
-            self.children = [
-                self.guide,
-                self._get_controls_section(),
-                self.results_container,
-            ]
+            children = [self.guide]
+            if (
+                self._model.identifier != "structure"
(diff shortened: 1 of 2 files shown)


AndresOrtegaGuerrero:
could you suggeset a test for this behaviour?, or this is while the simulation runs ?

edan-bainglass:
> could you suggeset a test for this behaviour?, or this is while the simulation runs ?

For starters, a completed job should not have a status notification. So we could test for this. Not sure how to test while running. Our workchain generation fixture is not so flexible I think. I'll add the first though.

AndresOrtegaGuerrero:
> > could you suggeset a test for this behaviour?, or this is while the simulation runs ?
> 
> For starters, a completed job should not have a status notification. So we could test for this. Not sure how to test while running. Our workchain generation fixture is not so flexible I think. I'll add the first though.

I meant to get this behaviour in qe app (error) , to check the logic and the response from the app you want

AndresOrtegaGuerrero (approved):
LGTM!

PR #1087, opened 2025-01-14
Fix process tree issues

edan-bainglass:
This PR handles a styling issue (casuing some buttons to have odd height) and a process label issue (now handling "vc-" labels). It also adds a link to the docs to learn more about how the restart feature works.


superstar54 (approved):
LGTM!

PR #1094, opened 2025-01-14
Handle cpu and memory check gracefully

superstar54:
This old script assumes a cgroupv2 environment, but the issue arises because some system doesn't have the required cgroupv2 files (/sys/fs/cgroup/cpu.max and /sys/fs/cgroup/memory.max). 

- If cpu.max and memory.max are missing, fall back to cgroupv1
- If neither cpu.max nor cpu.cfs_quota_us exists, use default values.


superstar54:
I tested the image on my local computer, and it workeded without error.

edan-bainglass (approved):
LGTM!

PR #1098, opened 2025-01-14
Adding widget to control aspect ratio of plot 

AndresOrtegaGuerrero:
As requested by Carlo, this PR we add widget to change the width of the plots of the bandspdosplot widget. 

For instance , in a 1D GNR the original 
image

can be modified to 
image

The same applied for the bands+pdos plot
image

The user can do modifcations to obtain 
image


edan-bainglass (approved):
Looks good!

PR #1101, opened 2025-01-16
Support building and publishing to PYPI

superstar54:
This PR allows us to build and publish the app into PYPI, so that the user can install the app by:
```
pip install aidialab-qe
```
In order to achieve this, this PR 

- uses relative path for the assets (e.g., log image)
- creates a `QeApp` class as the main entry point for the app
- adds github workflow to publish the app to pypi automatically when a release is made

One can use the app:

```python
from aiidalab_qe.app.main import QeApp
app = QeApp()
app.load()
```

Note that this does not mean the user can use the app directly in a normal Jupyter notebook with all features, because the app has other pages (calculation history etc.). However, the main calculation app itself is already beneficial for the advanced user who wants to use the app (or its functions) in the Pythonn script.


edan-bainglass (approved):
Very nice @superstar54. LGTM!

edan-bainglass:
@superstar54 can you merge this? 🙏 Would be very useful for testing against Solara.

superstar54:
Merged!

PR #1102, opened 2025-01-16
Add guards on null active eigenvalues

edan-bainglass:
There were no guards on the active eigenvalues. So if someone selected to define eigenvalues, but didn't (leaving all as -1), it raised an error.


AndresOrtegaGuerrero:
really , but i thought , https://www.quantum-espresso.org/Doc/INPUT_PW.html , -1 means not set by fault

AndresOrtegaGuerrero:
could you show the error we get ?

src/aiidalab_qe/app/configuration/advanced/hubbard/model.py
```diff
@@ -66,16 +66,22 @@ def update(self, specific=""):  # noqa: ARG002
...
+        ):
+            return []
         eigenvalues_array = np.array(active_eigenvalues, dtype=object)
-        new_shape = (np.prod(eigenvalues_array.shape[:-1]), 4)
-        return eigenvalues_array.reshape(new_shape).tolist()
+        new_shape = (int(np.prod(eigenvalues_array.shape[:-1])), 4)
+        return [
+            tuple(eigenvalue)
+            for eigenvalue in eigenvalues_array.reshape(new_shape).tolist()
+        ]
```
edan-bainglass, line 85: @AndresOrtegaGuerrero without the guard, `active_eigenvalues` might be empty (if all were left -1), in which case, `new_shape` is `(0, 4)`.

Two errors were raised here:
1. The shape was actually `(0.0, 4)` due to `np.prod` returning a `float`
2. Can't reshape an array of size 0 - `eigenvalues_array` has a null shape due to `active_eigenvalues` being empty

AndresOrtegaGuerrero: i see

AndresOrtegaGuerrero (approved):
LGTM! Thank you @edan-bainglass

PR #1104, opened 2025-01-17
Synchronize `spin_orbit` on process load

edan-bainglass:
#984 neglected to address this. The hack ensures that when a process is loaded, the unpacked parameters applied to `spin_orbit` in the advanced model are synchronized with the basic model (and UI by observation).


AndresOrtegaGuerrero (approved):
LGTM!

PR #1116, opened 2025-01-30
Clarify raw data / archive download message

### LINKED ISSUE #1109: Say "Creating file to download" rather than "Downloading file" when clicking on the two download buttons
From OP @giovannipizzi: As the current sentence "Downloading file" gives the feeling that the file is ready and the time taken is for download. Instead, it's for file creation.

edan-bainglass:
Change message to clarify that the waiting time is due to AiiDA creating the file, not the user's network.

Closes #1109

### DIFF
--- a/src/aiidalab_qe/app/result/components/summary/download_data.py
+++ b/src/aiidalab_qe/app/result/components/summary/download_data.py
@@ -83,7 +83,6 @@ class DownloadDataWidget(ipw.VBox):
         Args:
             button_instance (ipywidgets.Button): The button instance that was clicked.
         """
-        self._disable_buttons()
         if "archive" in button_instance.description:
             what = "archive"
             filename = f"export_qeapp_calculation_pk_{self.node.pk}.aiida"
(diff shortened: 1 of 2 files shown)


src/aiidalab_qe/app/result/components/summary/download_data.py
```diff
@@ -99,7 +99,12 @@ def _download_data(self, button_instance):
         self._enable_buttons()
 
     def _show_downloading_message(self, what):
-        self._downloading_message.value = f"Downloading {what} data..."
+        self._downloading_message.value = f"""
```
AndresOrtegaGuerrero, line 102: what about adding to the function more flexibility to control the message 
```python
def _show_downloading_message(self, what, message="Creating"):
```
```python
   {message} {what} data to download
```

or even flexibility to the size  or icon, leave it up to what you consider, i will approve in anycase

AndresOrtegaGuerrero (approved):
LGTM

PR #1121, opened 2025-01-30
Correct observation order of code selection

edan-bainglass:
Moved observation of code selector above its initial setting to ensure that the resource update, which also sets max CPUs, is triggered.

### DIFF
--- a/src/aiidalab_qe/common/panel.py
+++ b/src/aiidalab_qe/common/panel.py
@@ -282,11 +282,11 @@ class ResourceSettingsPanel(SettingsPanel[RSM]):
         else:
             code_widget = self.code_widgets[code_model.name]
         if not code_model.is_rendered:
-            self._render_code_widget(code_model, code_widget)
             code_widget.observe(
                 code_widget.update_resources,
                 "value",
             )
+            self._render_code_widget(code_model, code_widget)
 
     def _render_code_widget(
         self,


src/aiidalab_qe/common/widgets.py
```diff
@@ -706,10 +706,10 @@ def __init__(self, **kwargs):
         self.code_selection.layout.width = "80%"
 
         self.num_nodes = ipw.BoundedIntText(
-            value=1, step=1, min=1, max=1000, description="Nodes", width="10%"
+            value=1, step=1, min=1, max=10000, description="Nodes", width="10%"
         )
         self.num_cpus = ipw.BoundedIntText(
-            value=1, step=1, min=1, description="CPUs", width="10%"
+            value=1, step=1, min=1, max=10000, description="CPUs", width="10%"
```
AndresOrtegaGuerrero, line 712: Shouldnt this be set by the selected computer ?

mikibonacci: Something like this: `computer.get_default_mpiprocs_per_machine()`, but then I don't know if we can get a computer at rendering time, i.e. if there is already a computer/code from which we can take this information. For some code, we may need to first set it up (e.g. phonopy, but it can be true in general). 

So maybe this is fine, but I am not 100% sure.

edan-bainglass: Hmm, I think we should be able to set the max values for both nodes and CPUs from the computer as @AndresOrtegaGuerrero suggests, no? We can do it dynamically? When a code is selected, we load it, thus gaining access to the computer, and set the traits. I guess this is what we already do to set the default CPUs. We just also extend it to setting the max values.

mikibonacci: If we have a computer (or better, a code) it is surely possible, but what if a plugin defines a code which is not yet set up (let's say, the first time the user enable the plugin to run)? My concern is about the first time the resources general tab renders and, for a new unset code... however, we can still fallback to a default 1 and, as @AndresOrtegaGuerrero suggested, to set it up when a code is available

edan-bainglass: Well... looks like it is already implemented this way

```python
    def set_resource_defaults(self, computer=None):
        if computer is None:
            self.num_nodes.disabled = True
            self.num_nodes.value = 1
            self.num_cpus.max = 1
            self.num_cpus.value = 1
            self.num_cpus.description = "CPUs"
        else:
            default_mpiprocs = computer.get_default_mpiprocs_per_machine()
            self.num_nodes.disabled = (
                True if computer.hostname == "localhost" else False
            )
            self.num_cpus.max = default_mpiprocs <-- already implemented
            self.num_cpus.value = (
                1 if computer.hostname == "localhost" else default_mpiprocs
            )
            self.num_cpus.description = "CPUs"
```

So if there's a computer, max CPU is set to `default_mpiprocs`. This explains how it was able to go to 128 in @mikibonacci's case. But it doesn't explain why when the global settings were set to a code with 128 CPUs, the same code in the band structure panel was unable to sync, only going to 100 (the default max set by `ipywidget`). It could possibly be that somehow the order of events is off. Need to test.

edan-bainglass: Indeed, it was a wrong order of events. We were observing the code selector after its initial setting, so the first render did not reflect the correct max CPUs. It does now!

AndresOrtegaGuerrero (approved):

PR #1123, opened 2025-01-31
Fix loader margins

edan-bainglass:
Fixing an annoying CSS issue where a scroll bar keeps (dis)appearing due to the way the `fa-spin` class is implemented.


mikibonacci (approved):
LGTM! Thanks for the work

Issue #1122, opened 2025-01-30
Adding pencil decomposition (`-pd .true.`) also for `dos.x` and `projwfc.x`


mikibonacci:
This is needed as, if we run on large clusters, the default `num_cpus` will be large (e.g. for eiger is 128), so the calculation could except. For safety (the user may not know this a priori), we should require the pencil decomposition option.

superstar54:
Do `dos.x` and `projwfc.x` support `-pd .ture.`? I can not find any docs on these codes.

mikibonacci:
Yeah, they both want to FFT the charge density (or some related quantity), so they support it. For both cases I indeed get this error:

```
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
     Error in routine  fft_type_set (6):
  there are processes with no planes. Use pencil decomposition (-pd .true.)
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
```

mikibonacci:
So, for `dos.x` it works, for `projwfc.x` does not work (even if the error message is the same, suggesting to use it). 
I will put then the pd in the dos calculation, but not in projwfc. Thinking if we should put a warning, but I am not sure what to say then (we cannot know a priori when this decomposition is needed - i.e. when to use few CPUs).

PR #1125, opened 2025-01-31
Adding pencil decomposition for dos.x by default

mikibonacci:
This fixes https://github.com/aiidalab/aiidalab-qe/issues/1122

This because we risk to have too many processors and few plane waves. The problem is also present for projwfc.x, but there is not -pd option.

For projwfc.x, we can simply add something in the documentation of the app.


superstar54 (approved):
LGTM!

PR #1130, opened 2025-02-03
Fix nglview not show

superstar54:
I am testing the app inside normal notebooks, and this PR fixes:

- nglview not show in step 1 and step 4
- cutoffs are wrong in pseudo


edan-bainglass (approved):
Thanks @superstar54. LGTM! Good catch on the cutoffs issue 💪

PR #1133, opened 2025-02-03
Implement logging widget (mostly for devs)

edan-bainglass:
This PR adds a flag, which if True, adds an `Output` widget at the bottom of the app, accompanied by a reset button. The widget can then be passed down to any part of the app that needs to be tested.
The PR hardcodes one use-case - monitoring. The `ProcessMonitor` was recently changed (https://github.com/aiidalab/aiidalab-widgets-base/pull/669) to accept an output widget and use it to log exceptions in monitored callbacks. This PR passes the output widget (if included) down to the `ProcessMonitor` instance, exposing its exceptions. This is useful for debugging issues in the results step.

### DIFF
--- a/src/aiidalab_qe/app/main.py
+++ b/src/aiidalab_qe/app/main.py
@@ -5,6 +5,7 @@ Authors: AiiDAlab team
 
 from pathlib import Path
 
+import ipywidgets as ipw
 from IPython.display import display
 
 from aiidalab_qe.app.static import styles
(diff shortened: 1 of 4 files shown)


src/aiidalab_qe/app/main.py
```diff
@@ -34,6 +40,30 @@ def __init__(
...
+                    children=[
+                        reset_button,
+                        self.log_widget,
+                    ],
+                )
+            )
+
         # Set up bug report handling (if a URL is provided)
         if bug_report_url:
             install_create_github_issue_exception_handler(
```
superstar54, line 69: Still, show the output on top?

edan-bainglass: For now, this is fine. At the moment, my only use case is intercepting monitor-raised exceptions. These are silent, so they will not trigger the report widget. We can consider modifying this later though.

edan-bainglass: Hmm, doesn't hurt. I'll just capture these as well. Will update code.

superstar54 (approved):
Very useful! thanks!

PR #1137, opened 2025-02-04
Bump QE plugin to 4.9

### LINKED ISSUE #959: Check why PDOS sometimes skips the first semicore, and why PDOS and DOS are different
Update November 26

- We should set the default deltaE of DOS (and PDOS) to 0.01. This will reduce the probability to encounter the problem and "fix" the case of Au as tested by Michail Minotakis.

- We should expose in advanced settings deltaE in case a user would prefer a different default.

- We should still report the issue to QE developers

Image

BELOW info from November 19

Two notes:

- [ ] the output of QE also has the integral of the DOS/PDOS: its derivative does not miss the first state. Image

- [ ] For the difference between DOS and PDOS, of course they are different in the high conduction bands, but can we check why in the valence they are different? Different method, just different smearing, ...? Image

Some additional comments from @cpignedoli and @giovannipizzi investigation:

Issue to better understand in QE:
* PDOS not matching total dos; fast protocol, maybe still PDOS k-mesh is too coarse?
* Dos.x bug, missing narrow semicore states. E.g. when doing Au with the app, the state at -106 eV wrt Fermi level (~-85.1 with QE zero) shows no peak at all in DOS and PDOS with fast protocol, and small but not “full” peak for moderate:
* Au, moderate protocol, default QE app energy range:
  * Restarted by hand, changing Emin in the energy range to start at lower E: it cures a little bit, but still bug:

  * Note from above: the integral actually “sees” something of the peak (even if the DOS and the derivative of the integral do not match exactly, which is more accurate? Should we show by default the derivative of the integral instead? Is it possible also for the PDOS?)
* Note that in the stdout, the code “sees” the contribution. E.g. (for ref, E fermi = 20.7534 eV and with these orbitals:
   ```
     state #   1: atom   1 (Au ), wfc  1 (l=0 m= 1)
     state #   2: atom   1 (Au ), wfc  2 (l=1 m= 1)
     state #   3: atom   1 (Au ), wfc  2 (l=1 m= 2)
     state #   4: atom   1 (Au ), wfc  2 (l=1 m= 3)
     state #   5: atom   1 (Au ), wfc  3 (l=0 m= 1)
     state #   6: atom   1 (Au ), wfc  4 (l=2 m= 1)
     state #   7: atom   1 (Au ), wfc  4 (l=2 m= 2)
     state #   8: atom   1 (Au ), wfc  4 (l=2 m= 3)
     state #   9: atom   1 (Au ), wfc  4 (l=2 m= 4)
     state #  10: atom   1 (Au ), wfc  4 (l=2 m= 5)
   The lowest energy state at a couple of energy points chosen arbitrarily shows:
	 k =   0.0000000000  0.0000000000  0.0000000000
   ==== e(   1) =   -85.15582 eV ==== 
     psi = 0.973*[#   1]+0.027*[#   5]
    |psi|^2 = 1.000


    k =  -0.7071067812  0.2357022604  0.0000000000
   ==== e(   1) =   -85.13980 eV ==== 
     psi = 0.992*[#   1]+0.007*[#   5]
    |psi|^2 = 1.000
   ```
   Therefore, it knows it’s almost 100% the #1 projectors (lowest s state).

AndresOrtegaGuerrero:
The new QE plugin introduces the use of "tetrahedra_opt" in the NSCF calculation of the PDOS Workflow. 
This will close #959 , and It will allow the correct calculation of pdos for molecules as well.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ packages = find:
 install_requires =
     aiida-core~=2.2,<3
     Jinja2~=3.0
-    aiida-quantumespresso~=4.8
+    aiida-quantumespresso~=4.9
     aiidalab-widgets-base[optimade]==2.3.0a3
     aiida-pseudo~=1.4
     filelock~=3.8


edan-bainglass (approved):
Glad we got this through! Thanks again to @Minotakm for the PR, @mbercx for the release, and @AndresOrtegaGuerrero for the bump 💪

AndresOrtegaGuerrero:
@Minotakm , I just did a test with a molecule using the qe plugin 4.9, and I noticed that i am obtaining projections as 0.0 , and this is not the case when using 4.8 , Would you mind testing on a molecule 

For your reference I am adding  the calculation using 4.8 and 4.9. 

4,9
newplot. 

4.8
newplot (1)

Minotakm:
Hey @AndresOrtegaGuerrero do you mind sending me the structure and the pseudos you used?

AndresOrtegaGuerrero:
[Uploading export_35343_raw.zip…]()

Minotakm:
For some reason I cannot download them. Would you mind sending them over to my mail ?

AndresOrtegaGuerrero:
try again

cpignedoli:
We tested with Andres,  we cannot have tetraedra_opt with 0D systems. I also tried having 2 kpoints (nonsense) still does not work. We should have either occupations fixed or gaussian smearing (also tetrahedra was working in the past, but in my opinion it does not make sense)

Minotakm:
@cpignedoli Thanks for letting me know. I will do some tests and update you when I have the results.

PR #1140, opened 2025-02-05
Tetrahedra occupation for molecules 

AndresOrtegaGuerrero:
Since #1137 introduces an issue with PDOS calculations for molecules due to the use of "tetrahedra_opt" in the NSCF step, projections are missing in the results.

To fix this, we apply the tetrahedra method only for molecules, ensuring that projections are correctly included in PDOS calculations.

Additionally, we should discuss the following for molecular systems:

Allowing smearing (or fixed) in NSCF calculations (currently not supported in PdosWorkChain)? 
Assessing whether an NSCF step is necessary for PDOS calculations in molecules?


AndresOrtegaGuerrero:
@cpignedoli @giovannipizzi @Minotakm

cpignedoli:
@AndresOrtegaGuerrero  @giovannipizzi @Minotakm can't we just directly do all the changes needed to have the correct way of calculating molecules? If this requires more than a couple of days, I will install the branch proposed by @AndresOrtegaGuerrero for Empa users

Minotakm:
Hey @cpignedoli, Why it doesn't work for molecules needs to be investigated, and to be honest, I’m not sure why the optimized tetrahedra method isn’t working without trying it for molecules. Of course, as you mentioned earlier, a method that works for solids might not yield the same level of results for molecules and vice versa. If it’s urgent, I’d suggest sticking to the method that works best for your use case for now.

AndresOrtegaGuerrero:
I agree. There may be a need to define an appropriate protocol for molecules. For now, I’m leaving the PR here in case Empa wants to use it temporarily, but a proper discussion is still needed.

cpignedoli:
Dear all, I agree, and what I wrote before  is misleading: the fact that we have a correct DOS means that I cannot point the finger at tetraedra_opt as wrong for 0D. The problem could just be an issue of PDOS, great @Minotakm that you are investigating this and that we can use, in the meanwhile, the workaround proposed by @AndresOrtegaGuerrero

giovannipizzi:
Mmm that's strange. I was looking at the files you provided [here](https://github.com/aiidalab/aiidalab-qe/pull/1137#issuecomment-2636066103).

I see a few things that would need double checking @AndresOrtegaGuerrero:
- you are now using QE 7.4. Can you try with QE 7.2 (the one I have in the demo server), to see if it's an issue of QE? Quite a few things happened between 7.2 and 7.4 (including parallelization of the projwfc routines). Let's see if it's a bug of QE rather than the plugin
- With respect what happens to me if in the current demo server (still using tetrahedra instead of tetrahedra_opt in the NSCF), your input for PROJWFC also specifies a degauss, which means the code will not use tetrahedra, but just a smearing. Is this expected? Why is this new input now added? (maybe you selected a "custom degauss" in the advanced PDOS parameters? What happens if you don't do it?
- Your projwfc output does not have any file that would normally contain the PDOS, such as files named `aiida.pdos_atm#1(Si)_wfc#1(s)`. This might be the reason why there is no PDOS. Is the fact that the files are missing expected?

giovannipizzi:
(BTW: for molecules, instead of using a grid `AUTOMATIC 1 1 1 0 0 0`, we should use `GAMMA` and check if all features that we need work (and are not buggy). If this is the case, `GAMMA` will take half the time, because it assumes wave functions are real and therefore does 2 FFT in one single operation

AndresOrtegaGuerrero:
So we are testing the daint alps (Thats why the QE 7.4) , but @cpignedoli  did the test using QE 7.2 (right ? let me know if i am wrong on this one) but i also did a test with the QE 7.2 in my localhost, 

a few things

1) We use the point [1 1 1 0 0 0] instead , because it might bring incompatibility when using aiida vibroscopy , when applying electric field for molecules (but perhaps we can adapt the external plugin @mikibonacci ) 

> With respect what happens to me if in the current demo server (still using tetrahedra instead of tetrahedra_opt in the NSCF), your input for PROJWFC also specifies a degauss, which means the code will not use tetrahedra, but just a smearing. Is this expected? Why is this new input now added? (maybe you selected a "custom degauss" in the advanced PDOS parameters? What happens if you don't do it? [EDIT: I see this is enforced for 0D systems - why?]

If you do this, (not use degauss and have tetrahedra in the NSCF) , you will not get a contribution from DOS and PROJWFC.

> Your projwfc output does not have any file that would normally contain the PDOS, such as files named aiida.pdos_atm#1(Si)_wfc#1(s). This might be the reason why there is no PDOS. Is the fact that the files are missing expected?

My files are empty, even the pdos_tot. However when we use degauss(PROJWFC) and tetrahedra(NSCF) this is not the case. So indeed the issue is from the projwfc.x side .

cpignedoli:
Correct.
- I used 7.2
- But DOS looks correct to me.
- I have an apparently correct aiida.out from projwfc with non zero coefficients on the different atomic waves
- I have atom....(s)... files but with zero values

edan-bainglass:
@Minotakm will try to test this today if he has time, tomorrow at the latest. @AndresOrtegaGuerrero I haven't been following closely, but in your assessment (and others), do we hold off on a release until this is in?

giovannipizzi:
> Correct.
> 
> * I used 7.2
> * But DOS from dos.x looks correct to me.
> * I have an apparently correct aiida.out from projwfc with non zero coefficients on the different atomic waves
> * I have atom....(s)... files but with zero values
> * pdos_tot zero values as well

@cpignedoli can you do a download of the raw files (verdi process dump, or the zip download from AiiDAlab), to test this? (great if you could check that it includes the atom...(s)... files - if they are not there, let us know if it's because they are not there in the first place, or they are left on the remote computer scratch).

Since you have access to the scratch, it might also be useful to play a bit (directly on the supercomputer, no AiiDA) with the projwfc parameters to see if they make a difference (e.g. with tetrahedra_opt, would not putting a smearing and using tetrahedra work also for molecules, as it does for solids?).

Anyways, do I understand that this is not an issue of the aiida-qe plugin but of QE?

giovannipizzi:
One more thing - if really the only change is that `tetrahedra` worked and `tetrahedra_opt` does not (gives zero in PDOS): can you try also `tetrahedra_lin` please?

The issue of the different DOS vs PDOS is that projwfc, if it finds tetrahedra, actually uses tetrahedra_lin. 
I'd then prefer to have, as a workaround, using tetrahedra_lin for molecules (if _opt does not work), and clearly saying this somewhere.

Moreover, in any case: does it even matter which tetrahedron method you use, if in the end you put a degauss? @AndresOrtegaGuerrero @cpignedoli

AndresOrtegaGuerrero:
> One more thing - if really the only change is that `tetrahedra` worked and `tetrahedra_opt` does not (gives zero in PDOS): can you try also `tetrahedra_lin` please?
> 
> The issue of the different DOS vs PDOS is that projwfc, if it finds tetrahedra, actually uses tetrahedra_lin. I'd then prefer to have, as a workaround, using tetrahedra_lin for molecules (if _opt does not work), and clearly saying this somewhere.
> 
> Moreover, in any case: does it even matter which tetrahedron method you use, if in the end you put a degauss? @AndresOrtegaGuerrero @cpignedoli

I did the test with `tetrahedra_lin` , for the molecules ,and it works , i just push the change

AndresOrtegaGuerrero:
@edan-bainglass I changed the test to 

```python
def test_reminder_info():
    """Dynamic add/remove the reminder text based on the workchain settings."""
    model = ConfigurationStepModel()
    config = ConfigureQeAppWorkChainStep(model=model)
    config.render()
    bands_info = config.property_children[2].children[1] 
```
Because it seems that in #1139, by making the XAS and XPS plugin external , when is installed it changes the position of properties placing bands as third.

edan-bainglass:
> @edan-bainglass I changed the test to
> 
> ```python
> def test_reminder_info():
>     """Dynamic add/remove the reminder text based on the workchain settings."""
>     model = ConfigurationStepModel()
>     config = ConfigureQeAppWorkChainStep(model=model)
>     config.render()
>     bands_info = config.property_children[2].children[1] 
> ```
> 
> Because it seems that in #1139, by making the XAS and XPS plugin external , when is installed it changes the position of properties placing bands as third.

Maybe this is just too strict. We could instead search for the bands property by looking for a child with "Electronic Band Structure" info. What do you think? Can you test this?

```pyhon
bands_info = next("Bands" in child.children[1] for child in config.property_children, None)
```

Or something like that. We can assert that indeed there is a `bands_info`, then continue with the regular tests.

src/aiidalab_qe/plugins/pdos/workchain.py
```diff
@@ -87,6 +87,8 @@ def get_builder(codes, structure, parameters, **kwargs):
 
     # Update the nscf kpoints distance from the setting panel
     nscf_overrides["kpoints_distance"] = parameters["pdos"]["nscf_kpoints_distance"]
+    if structure.pbc == (False, False, False):
```
edan-bainglass, line 90: I've been using `not any(structure.pbc)` to check for molecules. Feel free to adopt or ignore :)

tests/test_configure.py
```diff
@@ -64,7 +64,14 @@ def test_reminder_info():
...
     config.render()
-    bands_info = config.property_children[0].children[1]
+    bands_info = next(
+        (
+            child.children[1]
+            for child in config.property_children
+            if "Electronic band structure" in child.children[0].title
+        ),
+        None,
+    )
```
edan-bainglass, line 74: This works okay?

edan-bainglass (approved):
LGTM! Glad resolution to the DOS matter was not too difficult.

tests/test_configure.py
AndresOrtegaGuerrero: yes

edan-bainglass:
> One more thing - if really the only change is that `tetrahedra` worked and `tetrahedra_opt` does not (gives zero in PDOS): can you try also `tetrahedra_lin` please?
> 
> The issue of the different DOS vs PDOS is that projwfc, if it finds tetrahedra, actually uses tetrahedra_lin. I'd then prefer to have, as a workaround, using tetrahedra_lin for molecules (if _opt does not work), and clearly saying this somewhere.
> 
> Moreover, in any case: does it even matter which tetrahedron method you use, if in the end you put a degauss? @AndresOrtegaGuerrero @cpignedoli

@giovannipizzi this seems to work for @AndresOrtegaGuerrero. Would you care to test, or shall we proceed with a merge?

giovannipizzi:
I can't test, I'm OK to go ahead. Just 2 comments: @AndresOrtegaGuerrero do you confirm then that with molecules tetrahedra_opt does not work and tetrahedra_lin does?

If so, please explain this very clearly at least in the code where this is done, also with a comment of why we do this.

Then, any idea of what happens in 1d and 2d?

I guess we can go ahead, and if we don't have an answer for 1d and 2d, I would open an issue and check it before the next release (April?)

cpignedoli:
I checked a 1D system, and to me tetrahedra_opt looks OK
I confirm that tetrahedra_opt does NOT work for molecules, while tetrahedra_lin seems to work

AndresOrtegaGuerrero:
@giovannipizzi Yes, if we use tetrahedra_opt, the projwfc.x doesnt produce the projections . I have added a comment in the code for future reference that we are doing this only for molecules.

giovannipizzi:
Ok, could one of you report this to qe, to see if this is known/expected?

PR #1143, opened 2025-02-06
Plot multiple bands and add wannier90 plugin

superstar54:
This PR supports plotting multiple bands in one plot with an additional plot setting.

For example, in the [wannier90 plugin](https://github.com/superstar54/aiidalab-qe-wannier90), we need to plot both pw bands and the wannier bands.

newplot

In addition, I added the Wannier90 plugin to the plugin list.


src/aiidalab_qe/common/bands_pdos/bandpdosplotly.py
```diff
@@ -58,6 +63,10 @@ def plot_type(self):
     def bandspdosfigure(self):
         return self.figure_object
 
+    @property
+    def first_band(self):
```
AndresOrtegaGuerrero, line 67: Could we try a different naming?

AndresOrtegaGuerrero: I would say that to avoid compatibility you , keep the naming of bands, and instead add a new naming for additional bands

AndresOrtegaGuerrero:
@superstar54 Than you , what about the labels? Did you check the other controls of color and so on work fine ? I put a comment regarding the new naming.

superstar54:
> what about the labels? 

I update the code to only draw the `labels` for the first bands.

> Did you check the other controls of color and so on work fine ? 

I checked other controls, and they work fine.

src/aiidalab_qe/common/bands_pdos/bandpdosplotly.py
superstar54: > and instead add a new naming for additional bands

Do you mean adding another variable, e.g., self.wannier_bands? The idea is that I don't want to hard coded the available bands.

superstar54:
Hi @AndresOrtegaGuerrero , I use `external_bands` to hold all other band data, and kept the `bands_data` untouched. 

Please review.

src/aiidalab_qe/common/bands_pdos/model.py
```diff
@@ -29,6 +29,7 @@
 
 class BandsPdosModel(Model):
     bands = tl.Instance(AttributeDict, allow_none=True)
+    external_bands = tl.Dict(value_trait=tl.Instance(AttributeDict), allow_none=True)
```
edan-bainglass, line 32: @superstar54 you designed this as a dictionary. What do keys represent?

edan-bainglass (commented):
@superstar54 overall it makes sense. I'm just not sure about the design and its general use. Let's discuss.

src/aiidalab_qe/common/bands_pdos/bandpdosplotly.py
```diff
@@ -28,8 +28,15 @@ class BandsPdosPlotly:
         "horizontal_range_pdos": [-10, 10],
     }
 
-    def __init__(self, bands_data=None, pdos_data=None, bands_projections_data=None):
+    def __init__(
+        self,
+        bands_data=None,
+        external_bands_data=None,
```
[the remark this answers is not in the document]
superstar54: >  I was wondering if an alternative approach would be to add a dimension to bands_data

I did this, but as suggested by @AndresOrtegaGuerrero , we need to keep the backward compatibility, so we kept the `bands_data` untouched, and add the `external_bands_data` to hold other bands data.

The spin is handled automatically inside the original code.

src/aiidalab_qe/common/bands_pdos/model.py
superstar54: The keys represent the names of the bands. They will appear as options in the dropdown menu, allowing the user to select and modify the appearance. 

```python
external_bands = {"wannier90": outputs.band_data,
                 "GW": ouputs.bands_data"}
```

edan-bainglass (approved):
LGTM!

superstar54:
Ping @qiaojunfeng; the Wannier90 plugin is registered in QEapp and will be available in the next release.

PR #1147, opened 2025-02-10
Support nested dependency

superstar54:
This PR allows us to use nested traits as dependencies, e.g.,

```python
dependencies = [
        'input_structure',
        'advanced.pseudos.functional'   # <===   here is the example
    ]
```


edan-bainglass (approved):
LGTM!

PR #1156, opened 2025-02-13
Pin plotly version

superstar54:
The latest plotly version (v6.0) needs Jupyter Notebook 7.0.

This PR pins the version 5.24, which I tested and work in our docker environment.


cpignedoli:
Thanks a lot @superstar54 I will test it as soon as posisble

PR #1157, opened 2025-02-13
Add information about magnetism feature limitations for insulators

edan-bainglass:
This PR adds a note in the magnetization section and one next to the electronic type selector informing users about the limitations in combining the insulator and magnetism options. It also styles info messages in the basic settings as inline info alerts to grab the users attention.

### DIFF
--- a/src/aiidalab_qe/app/configuration/basic/basic.py
+++ b/src/aiidalab_qe/app/configuration/basic/basic.py
@@ -34,6 +34,10 @@ class BasicConfigurationSettingsPanel(
             (self._model, "electronic_type"),
             (self.electronic_type, "value"),
         )
+        self.electronic_type.observe(
+            self._on_spin_type_change,
+            "value",
+        )
 
         # SpinType: magnetic properties of material
         self.spin_type = ipw.ToggleButtons(style={"description_width": "initial"})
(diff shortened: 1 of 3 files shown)


src/aiidalab_qe/app/configuration/basic/basic.py
```diff
@@ -50,9 +54,18 @@ def render(self):
             "value",
         )
 
+        self.starting_magmom_warning = ipw.HTML(
+            value="""
+                <div class="alert alert-inline alert-info" style="margin-left: 10px;">
+                    Can only set total magnetization for insulating systems
```
mikibonacci, line 60: Note: only total magnetization can be set for insulators
```suggestion
                    Can only set total magnetization for insulating systems
```

mikibonacci (approved):
Hi @edan-bainglass , LGTM! I just added a suggestion for a sentence, but not really important (decide what you think is better. Thanks!

src/aiidalab_qe/app/configuration/basic/basic.py
edan-bainglass: Yeah, I wasn't all too happy with my phrasing. Okay, in that case, I will make messages in the format of "Note: ..." 👍

PR #1159, opened 2025-02-14
Lift app setup above wizard initialization

edan-bainglass:


edan-bainglass:
Took a different route. See #1171

PR #1161, opened 2025-02-14
Use AWB's `AiidaNodeViewWidget` to handle node views

edan-bainglass:
This PR reverts to using the AWB node view widget, which now supports caching (https://github.com/aiidalab/aiidalab-widgets-base/pull/686).


mikibonacci (approved):
Hi @edan-bainglass, LGTM!

PR #1162, opened 2025-02-15
Rename reference to local config file

edan-bainglass:
No need to folderize this. Enough to just have `qe-config.yml` under ~/.aiidalab


superstar54:
Hi @edan-bainglass , why this change is needed? I would prefer to have one config folder for the app.

edan-bainglass:
Can you comment on what we should put under this folder other than the config file? If indeed there's more things to add, then we keep it a folder. Otherwise, a file would do, no?

superstar54:
Each QE app plugin could have its config file under that folder.

edan-bainglass:
> Each QE app plugin could have its config file under that folder.

I see. I guess we can pick these plugin-specific config files via the entry point system? This could work nicely by simply adding a path to the file in the `__init__` files. The path can be flexible, but a good place would be .aiidalab/quantum-espresso. I'm okay with this. Thoughts?

superstar54:
In this PR, the core app tries to load all plugins config and override the core app's config, this may result an issue: several plugins could override the same parameters, and the order of overridden is not guaranteed.

I would suggest that, the plugin handle their config files. We keep the config folder, and suggest the developers to use the folder, but it depends on the plugin developer whether they want to put the config file in the same folder as the core app or not.

edan-bainglass:
@superstar54 it is true that one file can override the other, but is this really our concern? I imagine plugin configs are to configure plugin-specific parameters. Since plugins were designed to be independent of one another, there should not be any shared parameters. Now of course this does not prevent one file from setting parameters of another, but that's just erroneous, no? It is unclear to me what you are proposing. Let's discuss later.

edan-bainglass:
@superstar54 following our discussion, the planned mechanism is not required and is removed from this PR. This PR is then simply a change of reference name to `.aiidalab/quantum-espresso/qe-config.yml`, with plugin configs left to plugin developers to author and use. We only recommend to use `.aiidalab/quantum-espresso/` as config file placement, but not strict.

superstar54 (changes requested):
Could you also provide a help function to return the path of the config folder, so that the plugin developer can use.

superstar54 (approved):
LGTM!

edan-bainglass:
Changing the reference to `.aiidalab/quantum-espresso/qe-config.yml`

PR #1163, opened 2025-02-16
Pop relax port if not relaxation

superstar54:
Currently, the `relax` port always exists in the builder even if the `relax` is not required, and the summary report also depends on the `relax` input.

This PR fixes this by:
- If no relaxation is required, the `relax` port will be popped out.
- The summary parameters are read from `ui_parameters` instead of `relax` port.

This fixes the wrong number of jobs in the status tree.
image

### DIFF
--- a/src/aiidalab_qe/app/result/components/summary/model.py
+++ b/src/aiidalab_qe/app/result/components/summary/model.py
@@ -226,17 +226,10 @@ class WorkChainSummaryModel(ResultsComponentModel):
             },
         }
 
-        # Extract the pw calculation parameters from the workchain's inputs
-        # energy_cutoff is same for all pw calculations when pseudopotentials are fixed
-        # as well as the smearing settings (smearing and degauss) and scf kpoints distance
-        # read from the first pw calculation of relax workflow.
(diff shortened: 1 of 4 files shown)


superstar54:
@edan-bainglass , this PR is ready for review.

edan-bainglass (approved):
Hi @superstar54. Code LGTM!. As long as you've tested and verified summaries look as they should in various cases, we can proceed with this one.

edan-bainglass:
@superstar54 I suppose with this in, I can remove the popping of non-property keys in #1166. Though I suppose it doesn't hurt and is a negligible performance hit. What do you think? Remove? Keep?

superstar54:
> I can remove the popping of non-property keys in https://github.com/aiidalab/aiidalab-qe/pull/1166. Though I suppose it doesn't hurt and is a negligible performance hit. What do you think? Remove? Keep?

I think we can remove it.

superstar54:
Hi @AndresOrtegaGuerrero , how to reproduce this error?

 This PR did not touch the code in `structure/model.py`. But indeed, there is a bug in the `self.properties`, which is an AiiDA data node. I am curious why we didn't get this error before?

PR #1166, opened 2025-02-17
Improve status panel behavior

edan-bainglass:
Based on #1162

This PR updates the status panel as follows:
1. The root node of the process tree is force selected to avoid inconsistent selection on process node loading
2. `metadata_inputs` keys not found in root workflow `properties` are ignored
3. * marking propagates upwards (legend updated to better reflect meaning)
4. Top-level relaxation workflow is no longer replaced by its children and is also tagged with † referencing a potential convergence testing runs


edan-bainglass:
@superstar54 @mikibonacci I tried to reproduce the duplicate-node issue on the new laptop. No luck!

Tried:
- Si2 unrelaxed + pdos
- Si2 fully relaxed + pdos + fat bands

I'll check in person later what you guys are running and we can discuss.

superstar54:
@edan-bainglass thanks for hte work. It looks like this PR is not related to #1162 . Could you remove the unrelated code? Keeping the PR focused will make the review process smoother.

edan-bainglass:
@superstar54 please review #1162 first, then I'll rebase 🙏 If you wish to review this one first, simply ignore the first file 🙂 The rest of the changes are relevant to this PR.

edan-bainglass:
@superstar54 please give it another look 🙏

src/aiidalab_qe/app/result/components/status/tree.py
```diff
@@ -79,23 +80,18 @@ def _render(self):
...
+                        <div style="font-style: italic;">
+                            *dynamic workflow may submit more jobs than expected
+                            <br>
+                            <b>(advanced)</b> click
+                            <a
+                                href="https://aiida.readthedocs.io/projects/aiida-core/en/stable/howto/workchains_restart.html"
+                                target="_blank"
+                            >here</a> to learn about AiiDA's dynamic error handling
+                        </div>
+                    """,
```
superstar54, line 94: The link to the `How to write error-resistant workflows` page is too advanced, even for experienced AiiDA users. I suggest not including it here. Since the calculation of the number of jobs is not strictly accurate, due to error handling, while/if logic, and the way how metadata is interpreted, I recommend adding a simple clarification:

`The number of jobs is an estimate. The actual number may vary due to error handling, dynamic sub-workflows, and other runtime adjustments.`

superstar54 (changes requested):
Please see the added comment.

superstar54 (approved):
LGTM!

PR #1167, opened 2025-02-17
Remove aiida-qe-xspec

superstar54:
XPS/XAS plugins are migrated out, so the `aiida-qe-xspec` should not be a dependency.


edan-bainglass (approved):
LGTM!

PR #1170, opened 2025-02-18
Fix `LinkButton` styles

edan-bainglass:
Styles were applied to the `LinkButton`'s inner anchor tag. Here, a `link-button` CSS class is added directly to the `LinkButton` class (inherits `ipywidgets.HTML`).


mikibonacci (approved):
Hi @edan-bainglass, LGTM!

PR #1174, opened 2025-02-20
Improve simplified process tree

edan-bainglass:
This PR does the following:
- Refactor tree classes to aiidalab_qe/common/process module
- Add missing "not rendered" guards for monitor-triggered events
- Improve tree styles, particularly height (max 500px)
- Improve performance of counting finished jobs by replacing recursion with single DB query
- Use a flag to signal when adding branches to avoid race conditions with monitoring thread

### Future plan
- [ ] Replace more iterative/recursive tree components with single DB query
- [ ] Fetch more info from plugins to avoid on-the-fly processing


edan-bainglass:
@mikibonacci @superstar54 please test for duplicate nodes. I added a test, but not sure how reliable it is.

superstar54:
Hi @edan-bainglass , thanks for the work. The test seems not reliable. I think the test needs to mock the `_add_branches_recursive` with some time duration (~ 5 seconds), then you start another thread to trigger the `_add_branches_recursive` simultaneously. The time duration is important to test whether the lock indeed works.. If this is not easy to implement, we can skip the test for the moment.

edan-bainglass:
> Hi @edan-bainglass , thanks for the work. The test seems not reliable. I think the test needs to mock the `_add_branches_recursive` with some time duration (~ 5 seconds), then you start another thread to trigger the `_add_branches_recursive` simultaneously. The time duration is important to test whether the lock indeed works.. If this is not easy to implement, we can skip the test for the moment.

I'll give it a try. I'm more concerned in the meantime not being able to resolve the compatibility issue with v2.3

edan-bainglass:
@superstar54 @mikibonacci test improved and indeed shows that the flag is necessary and effective.

tests/test_status.py
```diff
@@ -216,6 +218,44 @@ def test_calcjob_node(self):
...
+            # Without the `_adding_branches` flag, the monitor thread will
+            # attempt to add branches at the same time as the main thread,
+            # which will result in duplicate branches
+            assert len(self.tree.trunk.branches) > 1
+            self.tree.trunk._add_branches = add_branches  # restore original method
+        else:
+            # With the flag in place, the monitor thread bails from adding
+            # branches when the parent branch is presently doing so, leading
+            # to only the adding of the single intended branch
+            assert len(self.tree.trunk.branches) == 1
```
superstar54, line 257: The code starts the threading first, then click in main. Thus, it can not guarantee that the main will run first.

edan-bainglass: The monitor thread cannot add new branches to collapsed branches. Only after the main branch expands a parent branch and begins adding sub-branches does the monitor thread attempt to add sub-branches.

This was the root of the problem. Duplications happened when a monitor event happened shortly (and while) the user was expanding a parent node.

superstar54 (approved):
LGTM

PR #1175, opened 2025-02-20
Allow setting bands name and color

superstar54:
This PR allows us to set the name and line style of the bands from Python, as shown below.

image

This will be used in the wannier90 plugin, we want to fix the name `DFT Bands` and `Wannier-interpolated bands`.


AndresOrtegaGuerrero:
Hi @superstar54 , thank you , did you test that is compatible , with the bands and pdos (previou) calculations ?

superstar54:
> did you test that is compatible , with the bands and pdos (previou) calculations ?

Yes, I did the calculations on bands and pdos, and no issues. Since this PR only adds new features and does not modify the previous behavior, it should be compatible.

AndresOrtegaGuerrero (approved):
LGTM!

PR #1178, opened 2025-02-20
Hubbard: change int values of eigenvalues to float 

### LINKED ISSUE #1176: DFT+U Fails
Image

Try to run a DFT+U calculation fails (daint alps , QE 7.4)

AndresOrtegaGuerrero:
The values for the `starting_ns_eigenvalue `should be floats and not integers
At the moment the app was doing two errors 
1) it was sending integers
2) it was setting spin up as 0 and spin down as 1 (However this is not the convention of QE)

The test was testing spin down as 1.

This PR changes the values to floats and corrects the tests. Closes #1176


src/aiidalab_qe/app/configuration/advanced/hubbard/model.py
```diff
@@ -143,7 +143,7 @@ def _define_default_eigenvalues(self):
                     [state + 1, spin, kind_name, -1]  # default eigenvalue
                     for state in range(num_states)
                 ]
-                for spin in range(2)  # spin up and down
+                for spin in [2, 1]  # spin up and down
```
edan-bainglass, line 146: Why reversed? Is this a standard?

src/aiidalab_qe/app/configuration/advanced/hubbard/hubbard.py
```diff
@@ -211,17 +211,17 @@ def update(index, spin, state, symbol, value):
...
                         kind_index=kind_index,
                         state_index=state_index,
                         kind_name=kind_name: update(
                             kind_index,
-                            1,
+                            2,
                             state_index,
                             kind_name,
-                            int(value),
+                            float(value),
```
edan-bainglass, line 224: Ah, I see [here](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm503) that it is indeed expecting `REAL`. Good catch

edan-bainglass (approved):
LGTM! Great if you can answer my comment, for further clarification 🙏

src/aiidalab_qe/app/configuration/advanced/hubbard/model.py
AndresOrtegaGuerrero: i will reverse it ,

PR #1180, opened 2025-02-20
Fix properties in `HasProcess`

superstar54:
The input `properties` is a orm.List, and we return the raw list value for safety.

### DIFF
--- a/src/aiidalab_qe/common/mixins.py
+++ b/src/aiidalab_qe/common/mixins.py
@@ -94,7 +94,7 @@ class HasProcess(tl.HasTraits):
     @property
     def properties(self):
         process_node = self.fetch_process_node()
-        return process_node.inputs.properties if process_node else []
+        return process_node.inputs.properties.get_list() if process_node else []
 
     @property
     def outputs(self):


superstar54:
Hi @AndresOrtegaGuerrero , could you plase test importing the data again?

AndresOrtegaGuerrero:
So the issue @superstar54 is that we tested with @edan-bainglass , and we realize that this is occuring in one particular account . I produce a node and when i tried to imported in the user it didnt work , but in my container and edan it works. 
Edan and I have this behaviour for the node , 

```python
In [4]: node.inputs.properties
Out[4]: <List: uuid: 2fd7b49d-90d4-4a29-aeda-16fa3a3dbdcc (pk: 7738) value: ['vibronic']>
```

while in his account 

```python
In [2]: node.inputs.properties
Out[2]: <Data: uuid: 30f203c9-c12b-425f-a893-ee93f565cd4d (pk: 13234)>
```

src/aiidalab_qe/common/mixins.py
```diff
@@ -94,7 +94,13 @@ def inputs(self):
...
     def properties(self):
         process_node = self.fetch_process_node()
-        return process_node.inputs.properties if process_node else []
+        # read the attributes directly instead of using the `get_list` method
+        # to avoid error in case of the orm.List object being converted to a orm.Data object
+        return (
+            process_node.inputs.properties.base.attributes.get("list")
+            if process_node
+            else []
+        )
```
edan-bainglass, line 103: How/when/why does this conversion occur?

AndresOrtegaGuerrero: In the past, @superstar54 and I observed this behavior. I don’t fully understand why, but when importing an archive, it sometimes changes data types to Data. In this case, the user has the same aiida-core version as the exporting environment, so I have no idea why this happens. Should we reach out to someone from aiida-core for clarification?

superstar54: Hi @AndresOrtegaGuerrero , yes, please open an issue in the `Aida-core.`

PR #1183, opened 2025-02-21
Cache node per thread to reduce DB calls

edan-bainglass:
As the current implementation makes significant use of AiiDA nodes, good to cache them. Here I use a cache dict to store the node instance per thread id, to avoid DB session issues.


superstar54 (approved):
LGTM!

PR #1186, opened 2025-02-24
Codify state icons in a separate module

edan-bainglass:
This PR defines process state icons in a separate common module and utilizes them throughout the app.


superstar54 (approved):
Nice, thank for the work!

PR #1189, opened 2025-02-27
Total Magnetization in Advanced

AndresOrtegaGuerrero:
This PR , set the total_magnetization value in the advanced tab when loading a previous calculation


edan-bainglass (approved):
LGTM! Easy!

PR #1200, opened 2025-03-06
changing phonon and muon versions in plugin_list.

mikibonacci:
I am now using tags/releases from github.
This helps in updating the release to just be able to update the plugin to latests versions without having to change the app.

This is not ideal, but I will be fine for the time being, and for minor modifications which do not break backward comp.


superstar54 (approved):
LGTM!

PR #1207, opened 2025-03-12
Include Kaleido package

AndresOrtegaGuerrero:
Plotly requires the Kaleido package to export images in SVG and PDF formats. This dependency affects the electronic structure plugin, causing a bug when users attempt to download SVG files.


edan-bainglass:
@AndresOrtegaGuerrero can you describe the bug or reference an issue? 🙏

AndresOrtegaGuerrero:
in plotly for the svg images kaleido is required 
https://plotly.com/python/static-image-export/

Without the library , the app will report the bug that kaleido is missing

PR #1195, opened 2025-03-03
Change UI for energy thresholds to per atom

edan-bainglass:
This PR changes the UI for energy thresholds to be per atom by updating the logic and correcting the units.

This fixes https://github.com/aiidalab/aiidalab-qe/issues/1194

PR #1219, opened 2025-03-20
Removing num_atom from thresholds

AndresOrtegaGuerrero:
The values of forc_conv_thr and conv_thr were updated based on the number of atoms, following the protocol in AiiDA-QuantumESPRESSO. However, this approach contradicts the intent of manually setting convergence thresholds. This becomes problematic for large systems, as the thresholds become too loose, potentially affecting the accuracy of the calculations. This solve #1220

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/model.py
@@ -93,7 +93,6 @@ class AdvancedConfigurationSettingsModel(
             self._update_kpoints_mesh()
 
     def get_model_state(self):
-        num_atoms = len(self.input_structure.sites) if self.input_structure else 1
         parameters = {
             "initial_magnetic_moments": None,
             "pw": {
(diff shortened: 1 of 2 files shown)


AndresOrtegaGuerrero:
@superstar54 I’m not sure why the previous behavior was changed, especially since we’ve discussed before that for larger systems, the app’s protocols can lead to loose parameters. The purpose of these thresholds is to allow users to set the values they need. Since this was the previous behavior, changing it now can cause simulations to fail or not converge properly.

superstar54:
Hi @AndresOrtegaGuerrero , this is changed in this [PR](https://github.com/aiidalab/aiidalab-qe/pull/1195), I agree that this is a break change. 

The PR changed to `per atom` so that these parameters will not change in the GUI for different systems. The most important thing to let the user know the unit, either `per atom` or `extensive quality`. I don't remember why we switched to `per atom` instead of adding description (e.g., extensive quality) , maybe @mikibonacci  and @edan-bainglass  can comment.


> especially since we’ve discussed before that for larger systems, the app’s protocols can lead to loose parameters.

I think we need to change the protocol in aiida-qe to handle this.

AndresOrtegaGuerrero:
I close the PR since it will break also the logic of the protocol, the text and so on , I will just warn the users that the behaviour is different to avoid the calculations to crash.

PR #1227, opened 2025-03-26
Include PP plugin in manager

AndresOrtegaGuerrero:


AndresOrtegaGuerrero:
Ok then , let me see if i can update the cli part,

AndresOrtegaGuerrero:
also maybe you already have a python@localhost ? is it  possible ?

AndresOrtegaGuerrero:
@superstar54 you are right , I just update the CLI (the pythonjob part was not there). I did a new release of the plugin. Could you test again ?

PR #1230, opened 2025-03-28
Update app w.r.t new protocols

edan-bainglass:
This PR changes all instances of "moderate" and "precise" to "balanced" and "stringent" following the changes in the new `aiida-quantumespresso` release (v4.10). It also bumps the dependency of the plugin.


edan-bainglass:
@t-reents ready to test 🙏

t-reents:
@edan-bainglass perfect, I'll give it a try

t-reents (approved):
@edan-bainglass I ran 3 simple SCF calculations and the correct values are picked (I also validated the k-points when creating them manually).

edan-bainglass:
> @edan-bainglass I ran 3 simple SCF calculations and the correct values are picked (I also validated the k-points when creating them manually).

Thanks @t-reents! @superstar54 good to get your review on this. Also w.r.t the failed tests. Looks like maybe a new dependency issue. Will check later.

tests/configuration/test_advanced.py
```diff
@@ -21,8 +21,8 @@ def test_advanced_default():
     smearing_model.reset()
 
     assert smearing_model.type == "cold"
-    assert smearing_model.degauss == 0.01
-    assert model.kpoints_distance == 0.5
+    assert smearing_model.degauss == 0.0275
+    assert model.kpoints_distance == 0.3
```
superstar54, line 25: Instead of hardcoding the value and manually updating it when the protocol changes, it would be better to read the value from the protocol directly.
```python
parameters = PwBaseWorkChain.get_protocol_inputs(protocol)
# then you can use this value
parameters["kpoints_distance"]
```

superstar54 (changes requested):
Thanks for the update!  Only a small change request, please check the comment.

tests/configuration/test_advanced.py
edan-bainglass: Great idea. Let's do this in another PR. The protocols change seldom, and the testing suite needs all sorts of updates anyhow.

superstar54: Fine to me.

superstar54 (approved):
LGTM!

PR #1231, opened 2025-03-28
Clarify use of thresholds

### LINKED ISSUE #1220: input file convergence thresholds values different from that specified
The values of etot_conv_thr and conv_thr in the input file generated are different from that specified in the advanced settings panel of the configure workflow.
I am using QE app v25.03.0.

edan-bainglass:
This PR does the following:
- Separate and reorder thresholds UI
- Add documentation w.r.t number of atoms scaling
- Add labels showing the actual values (Ry) used in the calculation

image

Resolves #1220

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced/advanced.py
@@ -165,25 +165,21 @@ class AdvancedConfigurationSettingsPanel(
             (self._model, "scf_conv_thr_step"),
             (self.scf_conv_thr, "step"),
         )
-        self.forc_conv_thr = ipw.BoundedFloatText(
-            min=1e-15,
-            max=1.0,
-            format="0.0e",
-            description="Force:",
-            style={"description_width": "150px"},
-        )
-        ipw.link(
-            (self._model, "forc_conv_thr"),
-            (self.forc_conv_thr, "value"),
+        scf_conv_thr_abs = ipw.Label(
+            layout=ipw.Layout(
+                width="150px",
+                text_align="center",
+            )
         )
         ipw.dlink(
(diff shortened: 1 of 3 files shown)


edan-bainglass:
@giovannipizzi let me know if the text should be different 🙏

giovannipizzi:
I would split differently. On top: title should br "Threshold for SCF cycle" and you only have what is currently called SCF (should be "energy threshold"). Then a category "Thresholds for ionic convergence", and inside you have an energy threshold and a force threshold

giovannipizzi:
I would write the expression abstractly as `threshold x num_atoms` followed by "(where the number of atoms is `num_atoms = 2`)"

AndresOrtegaGuerrero:
I would suggest we include test to check the builder and the final values that go in the calculation.

edan-bainglass:
@giovannipizzi

image

edan-bainglass:
> I would suggest we include test to check the builder and the final values that go in the calculation.

@AndresOrtegaGuerrero do you mind handling this? 🙏 You can either commit to this PR, or we merge this and you can do a follow up PR.

giovannipizzi:
Very good @edan-bainglass! Maybe I'd just remove the first comma, before "and", that I probably suggested. Then it's OK.

There's only another thing, everywhere we should not speak of accuracy that it's usually used to refer to correctness wrt experiment, but of precision, i.e. wrt a fully converged calculation with correct pseudos etc (ie, getting to the "correct" dft results). This should be checked a bit everywhere.

I let you decide if it's something to do here, or open an issue to do later even after release (not super urgent, but good not to forget)

AndresOrtegaGuerrero (approved):
LGTM!

PR #1233, opened 2025-03-31
Add bader, wannier90 and pythonjob codes

superstar54:
This PR tries to pre-set up some necessary codes/pseudos for some core plugins that we maintain.

- **Bader**: install Bader in the same conda env as the QE
- **Wannier90**: because we need to use the latest source code from GitHub, we compile Wannier90 from scratch. 
- **AiiDA codes**: Bader, Wannier90, PythonJob
- **Pseudos**: Bader needs PAW, and Wannier90 needs SSSP 1.1.

With all this added, the size compared to the previous one:

| Image       | Size   |
|-------------|--------|
| New image   | 5.5 GB |
| Old image   | 4.92 GB |

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ ARG QE_VER
 ARG QE_DIR
 
 USER ${NB_USER}
-RUN mamba create -p ${QE_DIR} --yes qe=${QE_VER} && \
+RUN mamba create -p ${QE_DIR} --yes qe=${QE_VER} bader && \
     mamba clean --all -f -y
 
 # STAGE 2
@@ -59,11 +59,10 @@ RUN wget -c -O hq.tar.gz https://github.com/It4innovations/hyperqueue/releases/d
     tar xf hq.tar.gz -C /opt/conda/
 
 ENV PSEUDO_FOLDER=/tmp/pseudo
-# temporary solution to install aiida-pseudos from repo to avoid downloading issues
+# Install plugin
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
-     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} \
-     "aiida-pseudo@git+https://github.com/aiidateam/aiida-pseudo"
+     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader
 
 RUN mkdir -p ${PSEUDO_FOLDER} && \
     python -m aiidalab_qe download-pseudos --dest ${PSEUDO_FOLDER}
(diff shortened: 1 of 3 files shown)


superstar54:
For your reference, to test the image, you can run:

```
docker run -p 8889:8888 ghcr.io/aiidalab/qe:pr-1233
```
then open the terminal, and 
```
verdi code list
```
you will see `bader@localhost`, `wannier90@localhost`, `python@localhost`
and run 
```
verdi code test pk
```

Dockerfile
```diff
@@ -59,11 +59,10 @@ RUN wget -c -O hq.tar.gz https://github.com/It4innovations/hyperqueue/releases/d
     tar xf hq.tar.gz -C /opt/conda/
 
 ENV PSEUDO_FOLDER=/tmp/pseudo
-# temporary solution to install aiida-pseudos from repo to avoid downloading issues
+# Install plugin
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
-     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} \
-     "aiida-pseudo@git+https://github.com/aiidateam/aiida-pseudo"
+     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader
```
edan-bainglass, line 66: Added `aiida-bader`, but removed `aiida-pseudo`?

setup.cfg
```diff
@@ -26,7 +26,7 @@ packages = find:
 install_requires =
     aiida-core~=2.5,<3
     Jinja2~=3.0
-    aiida-quantumespresso~=4.9
+    aiida-quantumespresso==4.9.0
```
edan-bainglass, line 29: Not sure why the forced pin, but FYI, I'm doing `aiida-quantumespresso~=4.10.0` in #1230.

edan-bainglass (approved):
Thank @superstar54. Overall looks good! If you can respond to my comments 🙏 If no issues, go ahead 🙂

Dockerfile
superstar54: Yes, we used `aiida-pseudo` from github previously to avoid a bug in aiida-pseudo. Now the new version of aiida-pseudo is released, so no needed anymore.

`aiida-bader` is added, so that we can use it to run the post-install script to install pseudos.

setup.cfg
edan-bainglass: I see. No worries. As I mentioned, this is done in my PR.

edan-bainglass: Wait, why does it fail? Since #1230 bumps it to ~=4.10.0, will it cause issues for your setup?

superstar54: Update to 4.10.0 now will fail the tests, because your PR is not merged yet. That's why I pin the version.

PR #1236, opened 2025-03-31
Always refresh NGLView to ensure it is indeed rendered

edan-bainglass:
If one opens step 1 then quickly closes it, the NGLView widget is rendered while its parent is closed, leading to a computed width of 0 for the NGLView widget. This is a persistent issue not only with the NGLView, but also other similar viewers (e.g. WEAS).

This PR ensures that opening step 1 refreshes the NGLView **every time**!


superstar54 (approved):
LGTM!

PR #1243, opened 2025-04-01
Update/guides and sorting plugin order

mikibonacci:
- updating the basic in-app guide
- enforcing order in plugins list (first the built-in ones, then the rest, alphabetically)


edan-bainglass (approved):
LGTM!

PR #1245, opened 2025-04-01
Docker image pre-install all stable plugins

superstar54:
- Docker image pre-install all stable plugins: vibroscopy, muon and bader, and run the required `post-install` script by the plugins.
- I marked XAS/XPS as `beta` because there are not enough pseudopotentials yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,10 +59,12 @@ RUN wget -c -O hq.tar.gz https://github.com/It4innovations/hyperqueue/releases/d
     tar xf hq.tar.gz -C /opt/conda/
 
 ENV PSEUDO_FOLDER=/tmp/pseudo
-# Install plugin
+# Install plugin for post-install
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
-     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader
+     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader \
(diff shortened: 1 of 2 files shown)


superstar54:
Hi @mikibonacci, I imported the vibroscopy and muon examples, and the image works fine in the new image. But to make sure everything is correct, could you please run some test calculations for vibroscopy and muon? In principle, everything should already be set up (e.g., the codes), so you can start the calculation directly.

```
docker run -p 8888:8888 ghcr.io/aiidalab/qe:pr-1245
```

mikibonacci (approved):
LGTM! I tested and it works as expected

Dockerfile
```diff
@@ -128,6 +132,13 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${QE_APP_SRC},target=${QE_APP_SRC},rw \
     uv pip install --strict --system --compile-bytecode --cache-dir=${UV_CACHE_DIR} ${QE_APP_SRC} "aiida-hyperqueue@git+https://github.com/aiidateam/aiida-hyperqueue"
 
+# Install plugin in the final image
+RUN --mount=from=uv,source=/uv,target=/bin/uv \
+--mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
+ uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader \
+ git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
+ git+https://github.com/mikibonacci/aiidalab-qe-muon@v1.0.0
```
edan-bainglass, line 140: @superstar54 why is this necessary? You already install these into the tarball in an earlier stage. These dependencies will be included in the untarred home, no?

PR #1250, opened 2025-04-02
Update local install docs

edan-bainglass:
This PR adjusts the local installation instruction for clarity as follows:
- Link to general documentation for accessing AiiDAlab now opens in a new tab, so the user can easily return and continue
- The two steps (following general instructions and launching the app) are now numbered steps
- In "launching the app" page, the image has been updated to the current version of the home app


mikibonacci (approved):
LGTM!

docs/source/installation/access/container.rst
```diff
@@ -6,10 +6,18 @@ Local Docker Instance
 
    If you used another method to access AiiDAlab, you may proceed to :doc:`install the app </installation/install>`.
```
superstar54, line 7: Can we remove this line?

edan-bainglass: If they used another method, that's where they need to go. What's your concern?

superstar54 (approved):

docs/source/installation/access/container.rst
superstar54: Not sure if this still needed, because the user visits this page by choosing the `Local Docker Instance`. I feel that the user needs to read a lot of text before the installation.

but fine for me to keep.

PR #1251, opened 2025-04-02
Fix selection of indices in `Edit atom tags`

t-reents:
Discussed with @mikibonacci and @giovannipizzi 

It is possible to generate slices in the edit atom tags panel, e.g., `1..5` to select the sites with indices `[1, 2, 3, 4, 5]`. However, in case of a structure with 5 sites, `1..10` would fail and the table wouldn't even appear. In our discussion we agreed that we would expect a behavior similar to lists in Python, i.e., we would expect all indices to be selected/printed, as the maximum of the slice is larger than the number of sites.


edan-bainglass (approved):
LGTM!

PR #1252, opened 2025-04-02
Fix default magnetic moments for atoms with tags

### LINKED ISSUE #1248: `Bug`: Setting initial magnetic moments in antiferromagnetic systems.
**Describe the bug**
The magnetic moments are not initialised correctly when dealing with antiferromagnetic systems or systems with multiple kinds of a given element. The initial magnetic moments are taken from the AiiDA-QE plugin. However, we only provide initial magnetic moments for certain chemical elements there (based on the chemical element as key) and set the default to `0.1`. In the current version of the app, we simply use the `kind_names` to get the corresponding initial magnetic moments. In case the `kind_name` contains a tag, the default of `0.1` is used, as this is not an expected key in AiiDA-QE. This leads to the behavior that ferromagnetic `Ni` gets an initial magnetic moment of `5 mu_B` while antiferromagnetic `Ni1` and `Ni2` would receive `0.18 mu_B` (18 valence electrons times 0.1). See the attached screenshots. 

**To Reproduce**
Check default magnetic moments for the `Ni` example structure with the default structure and after setting tags.

**Screenshots**

Image

Image

t-reents:
Pinging @edan-bainglass who was refactoring this, but also @mikibonacci and @AndresOrtegaGuerrero as discussed.

This PR will address #1248.

The previous logic was a bit convoluted and resulted in the problem that it doesn't work with tags, see the issue for an example. 
Before describing in more detail, I believe that the fix can be very simple. We simply remove any tags by using `kind.name` and then access the default magnetic moments from the AiiDA-QE plugin.

------------------
Why the current version doesn't work.

Taking `Ni2` with kinds `Ni1` and `Ni2` as an example. First, we call https://github.com/aiidalab/aiidalab-qe/blob/9a7cc724c18413cecb81f8bdc4dfb29638e068bb/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py#L96-L107
which sets `self. _default_starting_magnetization` to `{'Ni1': 0.1, 'Ni2': 0.1}`, as the AiiDA-QE methods do only work with the chemical elements without any sites. Next, we call https://github.com/aiidalab/aiidalab-qe/blob/9a7cc724c18413cecb81f8bdc4dfb29638e068bb/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py#L115-L129
While we use the correct logic of accessing the magnetic moments based on the chemical elements, i.e., `kind.symbol` (`Ni`) in this example, we try to access `self. _default_starting_magnetization` instead of `self._DEFAULT_MOMENTS`. However, as mentioned above, `self. _default_starting_magnetization` only contains the keys with the tags and hence we return again the default of `0.1`. Since we don't use the pseudo family explicitly, and we only want to reproduce the behavior of the AiiDA-QE plugin, I propose the solution mentioned in the beginning: we simply access the default magnetic moments by removing any tag information. Please let me know if I'm missing something.

---

Closes #1248

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/magnetization.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/magnetization.py
@@ -41,10 +41,6 @@ class MagnetizationConfigurationSettingsPanel(
             self._on_magnetization_type_change,
             "type",
         )
-        self._model.observe(
-            self._on_family_change,
-            "family",
-        )
 
     def render(self):
         if self.rendered:
@@ -126,9 +122,6 @@ class MagnetizationConfigurationSettingsPanel(
     def _on_magnetization_type_change(self, _):
         self._toggle_widgets()
         self._model.update_type_help()
-
-    def _on_family_change(self, _):
-        self._model.update_default_starting_magnetization()
 
     def _update(self, specific=""):
         if self.updated:
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/magnetization.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/magnetization.py
@@ -41,6 +41,10 @@ class MagnetizationConfigurationSettingsPanel(
             self._on_magnetization_type_change,
             "type",
         )
+        self._model.observe(
+            self._on_family_change,
+            "family",
+        )
 
     def render(self):
         if self.rendered:
@@ -122,6 +126,9 @@ class MagnetizationConfigurationSettingsPanel(
     def _on_magnetization_type_change(self, _):
         self._toggle_widgets()
         self._model.update_type_help()
+
+    def _on_family_change(self, _):
+        self._model._update_default_moments()
 
     def _update(self, specific=""):
         if self.updated:
--- a/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/magnetization/model.py
@@ -1,11 +1,13 @@
 from copy import deepcopy
 
 import traitlets as tl
+from aiida_pseudo.groups.family import PseudoPotentialFamily
 
 from aiida_quantumespresso.workflows.protocols.utils import (
     get_magnetization_parameters,
 )
 from aiidalab_qe.common.mixins import HasInputStructure
+from aiidalab_qe.utils import fetch_pseudo_family_by_label
 
 from ..subsettings import AdvancedCalculationSubSettingsModel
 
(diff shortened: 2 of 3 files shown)


t-reents:
I'll add tests once we agreed that this is the correct way to proceed

edan-bainglass:
@t-reents you remove the entire pseudo-dependency component here. However, as I recall, we need to scale between magnetization (what we fetch) and moments (what we display), which depends on the pseudopotential used.

t-reents:
The dictionary that is stored `self._DEFAULT_MOMENTS`, is just the yaml file taken from AiiDA-QE which actually contains the magnetic moments, not magnetization. In the current version, we first convert it to magnetization and convert it back afterwards. At least this is my understanding/impression.

Regarding the tags, yes, but the problem is that we try to get the magnetic moments from `self. _default_starting_magnetization` by using `kind.symbol`. However, this dictionary only contains the keys with tags.

I'll prepare an example to better show what is going on. We can discuss in the office if you're around

edan-bainglass:
> The dictionary that is stored `self._DEFAULT_MOMENTS`, is just the yaml file taken from AiiDA-QE which actually contains the magnetic moments, not magnetization. In the current version, we first convert it to magnetization and convert it back afterwards. At least this is my understanding/impression.
> 
> Regarding the tags, yes, but the problem is that we try to get the magnetic moments from `self. _default_starting_magnetization` by using `kind.symbol`. However, this dictionary only contains the keys with tags.
> 
> I'll prepare an example to better show what is going on. We can discuss in the office if you're around

Right, sorry, I was updating my comment as you were typing 😅

AndresOrtegaGuerrero:
Thank you @t-reents , would it be possible to update (or add a test for the magnetization with a tag) if you are busy, lets open an issue. I would suggests you test, uploading an old calculation to check the advanced tab was properly upload. I will try to test it.

AndresOrtegaGuerrero:
@t-reents I did a test on your PR, everything works , the summary is correct ,and in the qe. input. I would still suggest you add a test for the tags.

AndresOrtegaGuerrero:
Update: I tested with GaAs, and in this case, it correctly does not set an initial magnetic moment. However, when using a structure containing elements like Co or Fe—elements that do have a magnetic moment—it correctly assigns values from aiida-qe.

That said, there's an interesting question here: If a system contains elements that are not typically magnetic, the initial moments are set to zero—even though that structure might still exhibit a ferromagnetic configuration.

We've had this discussion before, and Giovanni and Carlo agreed that if a magnetic calculation is selected, the app should attempt to set some initial magnetic moments to encourage a ferromagnetic result (if one exists). The idea is that the initial guess should actively suggest a reasonable magnetic configuration rather than defaulting to zero.

t-reents:
Ah well, I'll check again after lunch. It thought it was working correctly. Of course, it should always initialize a non 0 value, which should in principle always be the case in AiiDA-QE

AndresOrtegaGuerrero:
I think so, and in the tests we can include the default magnetic where a starting magnetization should be use, and another when using tags.

edan-bainglass:
> @AndresOrtegaGuerrero You're right, sorry!
> 
> So the idea would then be to use 0.1 as the magnetization in case the magnetic moments are set to 0, right (and to convert it to the magnetic moments using the pseudo)? In this case, the previous logic makes sense. So combining that logic with the fixed version should then cover everything. Do you agree? If yes, I'll adjust it later.

@t-reents can you re-state what you mean by "fixed version"? 🙏

t-reents:
My initially proposed version where we directly accessed the magnetic moments in AiiDA-QE. However, I ignored, that this yaml also contains cases where we set 0 as the magnetic moment and use 0.1 as the default magnetization (and it also included another small bug):
https://github.com/aiidateam/aiida-quantumespresso/blob/5d51932b6fb1e964e0753ccce32d40a35fb8ba89/src/aiida_quantumespresso/workflows/protocols/utils.py#L158-L165

This should now be addressed in the new version, which is more like a modification of your version instead of fully replacing it. Quickly summarized:

1. Get the default magnetic moments of AiiDA-QE. If non-zero, just use it.
2. If 0, set `0.1 * number of valence electrons` as magnetic moment.

I think it should now reproduce the correct behavior.

AndresOrtegaGuerrero:
@t-reents I tested it, and everything appears to be working correctly. I also verified it with tags, and it seems fine as well. It might be a good idea to open an issue to add a test for the QEApp builder to specifically check this part and ensure future reliability.

edan-bainglass (approved):
Very nice @t-reents! LGTM!

AndresOrtegaGuerrero (approved):
LGTM! , thank you @t-reents

PR #1253, opened 2025-04-02
Advanced in-app guides

t-reents:


t-reents:
@edan-bainglass ready for review (not sure if you did another review for the others, but you get the point 😄 )

src/aiidalab_qe/app/structure/__init__.py
```diff
@@ -161,6 +162,7 @@ def _render(self):
                 </p>
             """),
             self.manager,
+            # InAppGuide(identifier="structure-edit"),
```
edan-bainglass, line 165: ?

src/aiidalab_qe/guides/1_Advanced settings.html
```diff
@@ -0,0 +1,175 @@
+<div id="guide-header">
```
edan-bainglass, line 1: General comment here regarding the name of the file. I believe we now have another way to name the tutorial not dependent on the file name. In that case, for simplicity, maybe we name the basic tutorial just 0_basic.html and this one 1_advanced.html?

edan-bainglass (approved):
LGTM! Just maybe address my comments 🙏 As for the text, I trust you gave it a good read 🙂

src/aiidalab_qe/app/structure/__init__.py
t-reents: Cheers, this was just an initial test. However, this reminds me of a discussion with @giovannipizzi. We should consider, not necessarily now, but for the future, to add at least a place holder for in-app guides also in the `Edit structure` panel. For example, this would simplify the description of the setup steps for antiferromagnetic configurations (doubling the cell, assigning tags etc.)

src/aiidalab_qe/guides/1_Advanced settings.html
t-reents: Sure, wasn't aware of this, thanks!

t-reents: @mikibonacci Edan mentioned you worked on it, how does it work now?

PR #1257, opened 2025-04-03
Push image to docker hub

### LINKED ISSUE #1255: Docker supports arm and also push to docker hub
Wjem users install the app locally using `aiidalab-launch` or Docker Desktop, there are two issues:

- The qeapp image is not pushed to https://hub.docker.com/r/aiidalab/qe, so the user needs to open a terminal and pull the image from ghcr.
- There is no image for the Arm system.

superstar54:
Fix #1255 

Push the image to Docker Hub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,9 +62,9 @@ ENV PSEUDO_FOLDER=/tmp/pseudo
 # Install plugin for post-install
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
-     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader \
-     git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
-     git+https://github.com/mikibonacci/aiidalab-qe-muon@v1.0.0
+     uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader
+    #  git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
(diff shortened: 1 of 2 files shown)


superstar54:
@edan-bainglass , you can check here: https://hub.docker.com/r/aiidalab/qe/tags
you will find that it indeed push a image from this PR ([pr-1257](https://hub.docker.com/layers/aiidalab/qe/pr-1257/images/sha256-90e3289ead2e86fbcb2e3780526f31bad06a4eff489da50b454f01e87a9523dc))

Dockerfile
```diff
@@ -63,7 +63,7 @@ ENV PSEUDO_FOLDER=/tmp/pseudo
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=from=build_deps,source=${UV_CACHE_DIR},target=${UV_CACHE_DIR},rw \
      uv pip install --system --strict --cache-dir=${UV_CACHE_DIR} aiida-bader \
-     git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
+    #  git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
```
edan-bainglass, line 66: Why commented out? Same question for the rest of the file.

edan-bainglass (approved):
I trust you tested this. Go ahead. Maybe just address my comment regarding the Dockerfile. Also, if this works, and I can pull the QE image from Docker Hub, maybe we adjust the documentation to suggest this.

superstar54:
Yes, this makes the installation inside the Docker Desktop simple. We can wait for another fix for the arm64 machine, then we can update the docs together.

There is an issue with vibroscopy #1265, so I skipped it for the moment. @mikibonacci  and I are investigating it. Will add it back when we find the solution.

PR #1260, opened 2025-04-03
Update tutorial documentations

edan-bainglass:
This PR adds an in-app guide how-to section and points to it from the tutorials page, which is now reduced to screenshots/captions of key features from the in-app guides.


superstar54 (changes requested):
Could you please add the Bader charge? It is also in the Google Docs.

superstar54 (approved):
LGTM!

Issue #1214, opened 2025-03-14
The properties `List` is converted to a `Data` again!


superstar54:
I ran the app using the aiidalab-launch and the image of the latest version ghcr.io/aiidalab/qe:v25.03.0 QE. 
Try: 
- silicon
- full-geometry
- bands+pdos

Then the WorkChain is expected. Here is the error message

```
(base) jovyan@691b0b7dc5ef:~$ verdi process report 734
2025-03-14 13:01:38 [1 | REPORT]: [734|QeAppWorkChain|on_except]: Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/site-packages/plumpy/process_states.py", line 228, in execute
    result = self.run_fn(*self.args, **self.kwargs)
  File "/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/workchains/workchain.py", line 302, in run
    return self._do_step()
  File "/opt/conda/lib/python3.9/site-packages/aiida/engine/processes/workchains/workchain.py", line 319, in _do_step
    finished, stepper_result = self._stepper.step()
  File "/opt/conda/lib/python3.9/site-packages/plumpy/workchains.py", line 295, in step
    finished, result = self._child_stepper.step()
  File "/opt/conda/lib/python3.9/site-packages/plumpy/workchains.py", line 246, in step
    return True, self._fn(self._workchain)
  File "/opt/conda/lib/python3.9/site-packages/aiidalab_qe/workflows/__init__.py", line 245, in setup
    self.ctx.run_relax = "relax" in self.inputs.properties
TypeError: argument of type 'Data' is not iterable
```

The `Properties` input should be a `List`, but it seems to be a `Data` now.

superstar54:
Ran the calculation using the same parameters in the second time, no error anymore, more confused 😕

PR #1262, opened 2025-04-03
Fix `Data` node issues in tree

edan-bainglass:
Applying workaround to migrated `Data` node issue - see #1214


superstar54 (approved):
LGTM!

The failed test is not related.

PR #1263, opened 2025-04-03
Fix/1255/build arm image

superstar54:
This PR builds the image for ARM64. Updates the Dockerfile that uses conditional logic (based on TARGETARCH) to install or compile architecture-specific dependencies.


edan-bainglass (approved):
Looks great!

PR #1264, opened 2025-04-04
Adjust panel outline width to fit content

edan-bainglass:
We were getting ... for long descriptions, even though there is plenty of space 🙂


superstar54 (approved):
I trust your local test.

But it would be good if you could provide a screenshot next time?

Issue #1271, opened 2025-04-04
Make the arm image build optional.


superstar54:
because it takes two hours to build the arm image, one needs to wait 2 hours for the test to be finished. Better have the option to trigger the arm build manually.

PR #1272, opened 2025-04-04
Split the action into: build-upload and build-test

superstar54:
Fix https://github.com/aiidalab/aiidalab-qe/issues/1271

Previously, there was only one action to:
- build the image and upload (amd64 ~8mins, arm64 ~2h)
- test the app using the built image

Therefore, one needs to wait 2 h to see the result of the test.

In this PR, we separate the action into two:
- build-test: build only the amd64 (only ~ 8mins), and test the app
- build both amd64 and arm64 and upload ( ~ 2h)

For a normal PR from the fork, the `build-upload` will not triggered.

### DIFF
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -1,0 +1,111 @@
+---
+name: Build Docker image and test
+
+on:
+    pull_request:
+    push:
+        branches:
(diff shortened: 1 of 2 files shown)


edan-bainglass (approved):
Fantastic!

PR #1273, opened 2025-04-04
Fix guide bug

edan-bainglass:
The guides were not being packaged due to the removal of the __init__.py file. This PR restores it, but makes sure to only pick up the HTML documents as guides, ignoring anything else in the folder. This also allows guide developers to safely add images, figures, and other guide-related files along with their guides.


superstar54 (approved):
LGTM!

PR #1277, opened 2025-04-04
build amd and arm separately and then merge into one

superstar54:
Split Docker builds by arch and add multi-arch manifest creation

- Build amd64 and arm64 separately to reduce runner disk usage
- Use `imagetools create` to merge into multi-arch image


superstar54:
Replaced by #1278

PR #1278, opened 2025-04-05
build amd and arm separately and then merge into one

superstar54:
Split Docker builds by arch and add multi-arch manifest creation

- Build amd64 and arm64 separately to reduce runner disk usage
- Build arm64 on the github arm runner
- Use `imagetools create` to merge into multi-arch image


mikibonacci:
Hi @superstar54 , LGTM!

mikibonacci (approved):

PR #1282, opened 2025-04-05
Update Dockerfile

mikibonacci:
Hi @superstar54 , just putting the aiida-bader post install at the end of the RUN, because fails, I think then nothing will be executed (like the other post install). See https://github.com/aiidalab/aiidalab-qe/actions/runs/14280666542/job/40029931956#step:8:1714

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,7 +129,6 @@ RUN --mount=from=qe_conda_env,source=${QE_DIR},target=${QE_DIR} \
     git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
     git+https://github.com/mikibonacci/aiidalab-qe-muon@v1.0.0 && \
     # run post_install for plugin
-    python -m aiida_bader post-install && \
     python -m aiidalab_qe_vibroscopy setup-phonopy && \
     python -m aiidalab_qe_muon setup-python3 && \
     # wannier90 plugin need SSSP 1.1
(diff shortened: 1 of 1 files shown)


superstar54:
Hi @mikibonacci , thanks for pointting out this, the error related to bader plugin seems to be a general problem for all qeapp plugins.

I did a quick test locally, for wannier90
```
>>> from aiidalab_qe_wannier90 import wannier90
Failed to load plugin entry point wannier90.
This may be due to compatibility issues with the current QEApp version.
Please contact the plugin author for further assistance.
Thus, the plugin will not be available. However, you can still use the rest of the app.
Error message: partially initialized module 'aiidalab_qe_wannier90.wannier90' has no attribute 'wannier90' (most likely due to a circular import)
```

also for vibroscopy
```
>>> from aiidalab_qe_vibroscopy.app import property
Failed to load plugin entry point vibronic.
This may be due to compatibility issues with the current QEApp version.
Please contact the plugin author for further assistance.
Thus, the plugin will not be available. However, you can still use the rest of the app.
Error message: partially initialized module 'aiidalab_qe_vibroscopy.app' has no attribute 'property' (most likely due to a circular import)
```

The `post-install` of bader plugin also installs additional pseudos. Even though there is a error message, but the post-install script works as expected, so we need to keep it.

Still need more time to investigate the reason, but it does not affect the functionality of the app in the final image. So I suggest that we leave it for the future.

superstar54:
Make a note for the future, to reproduce the error
```
docker run -it ghcr.io/aiidalab/full-stack:2025.1026 /bin/bash
```
then install `aiidalab-qe`, then run this:
```python
>>> from aiidalab_qe.plugins.pdos import pdos
Failed to load plugin entry point pdos.
This may be due to compatibility issues with the current QEApp version.
Please contact the plugin author for further assistance.
Thus, the plugin will not be available. However, you can still use the rest of the app.
Error message: partially initialized module 'aiidalab_qe.plugins.pdos' has no attribute 'pdos' (most likely due to a circular import)


```

superstar54 (approved):
LGTM!

PR #1293, opened 2025-04-05
Bump version v25.04.0 -> v25.04.1a2

superstar54:


superstar54:
Closed this, because all these PRs are only used to test the build of the image with `latest` tag.
Same for #1291 , #1289 , #1287 , #1285

PR #1301, opened 2025-04-07
Add post-install script for the `aiida-qe-xspec` plugin

superstar54:
`aiida-qe-xspec` introduces `post-install` script to download and import the core-hole pseudopotentials.


mikibonacci (approved):
LGTM!

PR #1318, opened 2025-05-19
WIP: Test with aiida 2.7

danielhollas:


danielhollas:
@superstar54 thanks! Basically I just need release a new docker image with the new version and this should be good to know. Will do later today.

PR #1336, opened 2025-06-05
Decrease QeApp images size

danielhollas:
I noticed a clear issue where the image size is bloated by installing the compilation and lapack toolchains. These need to be installed in the same step as they are removed, otherwise they stay in the Docker layer.

Also I would strongly recommend to put all this compilation logic into a script and call it from Dockerfile, as the Dockerfile has become very messy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -174,7 +174,7 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
 RUN pip install aiida-bader \
     git+https://github.com/mikibonacci/aiidalab-qe-vibroscopy@v1.2.0 \
     git+https://github.com/mikibonacci/aiidalab-qe-muon@v1.0.0 && \
-    conda install scipy==1.13.1 --y
+    mamba install scipy==1.13.1 --y
 
 # copy hq binary
 COPY --from=home_build /opt/conda/hq /usr/local/bin/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -182,12 +182,11 @@ COPY --from=home_build /opt/conda/hq /usr/local/bin/
...
 
 USER root
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-      gfortran libblas-dev liblapack-dev git openmpi-bin
 
 # Build wannier90 for all arches, and build bader from source ONLY on arm64
 RUN set -ex; \
+    apt-get update && apt-get install -y --no-install-recommends \
+    gfortran libblas-dev liblapack-dev git openmpi-bin liblapack3; \
```
danielhollas, line 189: Hmm, why are you installing git here? It should be in the image already? Is openmpi-bin really needed here? Isn't it installed in the qe conda environment?

superstar54:
@danielhollas Thanks for investigating this. 

What's the size difference before and after this PR?

> Also I would strongly recommend to put all this compilation logic into a script and call it from Dockerfile, as the Dockerfile has become very messy.

Agree.

superstar54 (approved):
Nice!

Dockerfile
danielhollas: `openmpi-bin` doesn't seem to be needed for compilation, but probably is for runtime? I've put it back for now to reduce the change of breakage.

But I think all these dependencies would be better installed via conda packages so that we end up with consistent environment. But that's for another PR.

PR #1340, opened 2025-06-29
Fix width in `fat bands`

t-reents:
Currently, we determine the line width to display the projectability in a `fat bands` plot simply by shifting the bands up and down by `+- band_width / 2`. 
However, this is not correct as one would need to perform the shift perpendicular to the band. As an example, a vertical band would have 0 width in the current version, independent of the projectability.

I discussed with @edan-bainglass @eimrek at some point that it might be possible to use the gradients to plot the correct `fat bands`. While the general idea would work, I realized during the last days that this approach doesn't work that smoothly and produces visual artefacts, e.g., in areas with a strong curvature. Hence, I move to a simple scatter plot in this version to visualize the projections, which is also used in other tools and much simpler than the actual filled line plots with varying width, as outlined above. Moreover, it avoids recalculating the projections and so may improve performance. 

Also pinging @AndresOrtegaGuerrero 

Happy to discuss this further, just wanted to open it already so that it can be used for the paper.

### new version

image

### old version

image

### DIFF
--- a/src/aiidalab_qe/common/bands_pdos/bandpdosplotly.py
+++ b/src/aiidalab_qe/common/bands_pdos/bandpdosplotly.py
@@ -414,13 +414,13 @@ class BandsPdosPlotly:
                 mode="markers",
                 line={"width": 0, "color": data["color"]},
                 name=data["label"],
-                marker=dict(
-                    size=np.array(data["marker_sizes"]) * projections_width,
-                    sizemode="diameter",
-                    color=data["color"],
-                    opacity=0.7,
-                    line=dict(width=0),
-                ),
+                marker={
+                    "size": np.array(data["marker_sizes"]) * projections_width,
+                    "sizemode": "diameter",
+                    "color": data["color"],
+                    "opacity": 0.7,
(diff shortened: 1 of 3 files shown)


AndresOrtegaGuerrero:
Hi @t-reents,
I was wondering if there's a way to improve the appearance of the plot or to smooth the data. When comparing the figures, I do see some projections along the vertical lines (Old version). However, the way the plot currently displays  (New) what look like "gaps" is a bit confusing—it gives the impression that there are no projections in certain parts of the bands. I think it would be helpful to clarify or resolve this before moving forward.

t-reents:
Yes, either we simply interpolate the projections to always have a certain density of points, which is implicitly done when plotting the projections as filled bands in the previous version, or we adjust the minimum size of the individual points.

There is anyway one additional aspect: When moving to a scatter plot, the slider should no longer have explicit units, as we only change the size of the markers which is defined in pixels. We don't have an easy way to express this in data units.

AndresOrtegaGuerrero:
Then i guess before that is solve , the best is to set this one as a draft

PR #1343, opened 2025-07-17
Save and load result panel state

superstar54:
Implement the feature to save and load the result panel state.
- Add save and load state button for the result panel
- The state parameters are saved as "result" in the `node.base.extras`.

Take the electronic structure as an example (only a draft)

### DIFF
--- a/src/aiidalab_qe/plugins/electronic_structure/result/model.py
+++ b/src/aiidalab_qe/plugins/electronic_structure/result/model.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import traitlets as tl
+
 from aiidalab_qe.common.panel import ResultsModel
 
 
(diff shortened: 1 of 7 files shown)


src/aiidalab_qe/plugins/electronic_structure/result/model.py
```diff
@@ -20,6 +22,27 @@ class ElectronicStructureResultsModel(ResultsModel):
...
+            ("Group all orbitals per atom", "total"),
+            ("Group by angular momentum ", "angular_momentum"),
+            ("No grouping (each orbital separately)", "orbital"),
+        ],
+    )
+    dos_plot_group = tl.Unicode("angular_momentum")
+    selected_atoms = tl.Unicode("")
+    project_bands_box = tl.Bool(False)
+    proj_bands_width = tl.Float(0.5)
+
```
edan-bainglass, line 45: Why is all this here? It is duplicated from the bands pdos widget.

edan-bainglass (commented):
The approach is a bit unclear. Let's discuss.

superstar54:
Hi @edan-bainglass , thanks for the comments!

> steps 2 and 3, we do have calculation settings and resource settings, respectively. So it is fair to also extend this to step 4.

This is what I thought when I added this feature: it makes the app consistent throughout the steps. However, we can discuss and refactor to make this more clear.

PR #1344, opened 2025-07-18
Reduce the number of decimal points to displayed

AndresOrtegaGuerrero:
Here we reduce the number of decimals points to be displayed ,

image

This way, we avoid displaying numbers like 1.9999999999999 without exponential notation.
For example , 

image

I also lowered the minimum allowed SCF threshold to 1e-18. For systems with more than 120 atoms, we sometimes need thresholds around 1e-14 for vibrational calculations, so this small adjustment makes that possible.


edan-bainglass (approved):
Very good!

PR #1360, opened 2025-08-10
Discard local `LoadingWidget` in favor of AWB

edan-bainglass:
Based on #1362 

The `LoadingWidget` moved to AWB in https://github.com/aiidalab/aiidalab-widgets-base/pull/685. This PR discards the widget in this repo and points all imports to AWB.

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/subsettings.py
+++ b/src/aiidalab_qe/app/configuration/advanced/subsettings.py
@@ -6,6 +6,7 @@ import traitlets as tl
 
 from aiidalab_qe.common.mixins import HasBlockers
 from aiidalab_qe.common.mvc import Model
+from aiidalab_widgets_base import LoadingWidget
 
 
 class AdvancedCalculationSubSettingsModel(
@@ -41,8 +42,6 @@ M = t.TypeVar("M", bound=AdvancedCalculationSubSettingsModel)
 
 class AdvancedConfigurationSubSettingsPanel(ipw.VBox, t.Generic[M]):
     def __init__(self, model: M, **kwargs):
-        from aiidalab_qe.common.widgets import LoadingWidget
-
         self.loading_message = LoadingWidget(f"Loading {model.identifier} settings")
 
         super().__init__(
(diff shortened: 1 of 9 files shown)


edan-bainglass:
@AndresOrtegaGuerrero the tests are failing already on `main`. Somehow the pseudo tests that passed before are now broken, even though no commits since they last passed. They seem to be quite fragile w.r.t order of events. I'll open an issue/PR later to fix this matter. Doesn't affect this PR, nor #1361.

AndresOrtegaGuerrero:
dont we need first a release of AWB first , before merging this ?

edan-bainglass:
> dont we need first a release of AWB first , before merging this ?

https://github.com/aiidalab/aiidalab-widgets-base/pull/685 is already in the current version of AWB (2.4.0)

AndresOrtegaGuerrero:
All the external plugins need a PR otherwise it will break the app , perhaps in this PR , you can leave (temporarily) LoadingWidgets ( but still use the AWB one in the app)  with a warning that it will be removed, while all the external plugins update and adapt the AWB

edan-bainglass:
> All the external plugins need a PR otherwise it will break the app , perhaps in this PR , you can leave (temporarily) LoadingWidgets ( but still use the AWB one in the app) with a warning that it will be removed, while all the external plugins update and adapt the AWB

Good point. Will do. Thanks!

AndresOrtegaGuerrero:
> @AndresOrtegaGuerrero have a look. I now do
> 
> ```python
> class LoadingWidget(_LoadingWidget):
>     def __init__(self, *args, **kwargs):
>         warnings.warn(
>             "`LoadingWidget` has moved to `aiidalab-widgets-base` and will be "
>             "deprecated here in the next major release. Please update your imports.",
>             DeprecationWarning,
>             stacklevel=2,
>         )
>         super().__init__(*args, **kwargs)
> ```
> 
> where `_LoadingWidget` is coming from AWB. This supports our plugins, but note that they will not see this other than maybe in their test output.
> 
> Did you have something else in mind?

I think is enough (Thanks!), I will try to do the modifications in vibro and pp plugins !

PR #1362, opened 2025-08-10
Fix pseudopotential fixtures and tests

edan-bainglass:
This PR simplifies the pseudopotential session test fixtures. It also updates the tests, which now reflect correct pseudopotentials w.r.t the codebase.


edan-bainglass:
@unkcpz I've requested you for review as I believe you had implemented these original fixtures. Would appreciate your eyes on it 🙏

PR #1361, opened 2025-08-10
Refactor convergence settings to separate sub-advanced-settings module

edan-bainglass:
Based on #1360

This PR refactors the convergence threshold section out to its own advanced sub-settings module.

Follow-up PR will go further in simplifying the advanced settings panel/model by delegating model tasks to the sub-settings.

PR #1363, opened 2025-08-10
Split advanced settings into tabs

edan-bainglass:
Based on #1361

Recent work updating/extending the advanced settings (e.g., #1355, #1358) has motivated switching from a vertical scroll to a horizontal one (i.e. tabs).

image

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/subsettings.py
+++ b/src/aiidalab_qe/app/configuration/advanced/subsettings.py
@@ -36,6 +36,9 @@ class AdvancedCalculationSubSettingsModel(
         """Resets the model to present defaults."""
         raise NotImplementedError()
 
+    def _get_default(self, trait):
+        return self._defaults.get(trait, self.traits()[trait].default_value)
+
 
 M = t.TypeVar("M", bound=AdvancedCalculationSubSettingsModel)
 
--- a/tests/test_configure.py
+++ b/tests/test_configure.py
@@ -3,16 +3,6 @@ from aiidalab_qe.app.configuration.model import ConfigurationStepModel
 from aiidalab_qe.setup.pseudos import PSEUDODOJO_VERSION, SSSP_VERSION
 
 
-def test_protocol():
-    model = ConfigurationStepModel()
-    _ = ConfigureQeAppWorkChainStep(model=model)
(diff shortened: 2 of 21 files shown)


AndresOrtegaGuerrero:
From a technical perspective, k-point distance impacts both accuracy and convergence, but I’d lean more toward classifying it under accuracy.
On the design side, I’m not sure if using accordions would actually work better.

edan-bainglass:
> From a technical perspective, k-point distance impacts both accuracy and convergence, but I’d lean more toward classifying it under accuracy. 

Sure. I kept it top level for now, but happy to reevaluate 🙂 

> On the design side, I’m not sure if using accordions would actually work better.

Thought about it but decided to proceed with tabs, as I believe they are a faster UI for though who want a quick overview. There was a comment at some point that people want to see the whole thing together. I think this argument is a bit optimistic w.r.t the capacity of the human mind, but at least fast tab switching gets the user close to that state. But again, happy to reevaluate, or at least to be proven wrong about this 🙂

edan-bainglass:
@superstar54 @mikibonacci can I get your comments on this one? 🙏

superstar54:
> To keep tab titles short, I would suggest "Convergence/Accuracy". 

Agree!

src/aiidalab_qe/app/configuration/advanced/subsettings.py
```diff
@@ -36,6 +36,9 @@ def reset(self):
         """Resets the model to present defaults."""
         raise NotImplementedError()
 
+    def _get_default(self, trait):
+        return self._defaults.get(trait, self.traits()[trait].default_value)
+
```
edan-bainglass, line 41: Lifted up to parent class

tests/test_configure.py
```diff
@@ -3,16 +3,6 @@
 from aiidalab_qe.setup.pseudos import PSEUDODOJO_VERSION, SSSP_VERSION
 
 
-def test_protocol():
```
superstar54, line 6: Why was this test deleted?

superstar54: I see. Then, no need to reinstate.

superstar54 (approved):
LGTM!

PR #1341, opened 2025-07-02
Use shallow copy to avoid creating a new AiiDA node every time

superstar54:
The parameters collected from the GUI are passed to the builder of the QeAppWorkChain, and then passed to each plugin's WorkChain builder. To avoid each plugin muting original parameters, we used deepcopy before. However, deepcopy will create new AiiDA nodes, breaking the provenance. In this PR, we replace the deepcopy with a shallow copy, so the plugin can not mutate the original dict.

Here are the inputs of a new calculation, which shows all the pseudos are consistent now.

```
Inputs                               PK    Type
-----------------------------------  ----  -------------
bands
    bands
        scf
            pw
                pseudos
                    Si               134   UpfData
                code                 1216  InstalledCode
                parameters           4532  Dict
                parallelization      4533  Dict
                settings             4534  Dict
            kpoints_distance         4535  Float
            kpoints_force_parity     4536  Bool
            max_iterations           4537  Int
        bands
            pw
                pseudos
                    Si               134   UpfData
                code                 1216  InstalledCode
                parameters           4538  Dict
                parallelization      4539  Dict
                settings             4540  Dict
            max_iterations           4541  Int
        clean_workdir                4542  Bool
        nbands_factor                4543  Float
        bands_kpoints_distance       4544  Float
    structure                        4522  StructureData
pdos
    scf
        pw
            pseudos
                Si                   134   UpfData
            code                     1216  InstalledCode
            parameters               4545  Dict
            parallelization          4546  Dict
            settings                 4547  Dict
        kpoints_distance             4548  Float
        kpoints_force_parity         4549  Bool
        max_iterations               4550  Int
    nscf
        pw
            pseudos
                Si                   134   UpfData
            code                     1216  InstalledCode
            parameters               4551  Dict
            parallelization          4552  Dict
            settings                 4553  Dict
        kpoints_distance             4554  Float
        kpoints_force_parity         4555  Bool
        max_iterations               4556  Int
    dos
        code                         3     InstalledCode
        parameters                   4557  Dict
        settings                     4558  Dict
    projwfc
        code                         2     InstalledCode
        parameters                   4559  Dict
    nbands_factor                    4560  Float
    clean_workdir                    4561  Bool
relax
    base
        pw
            pseudos
                Si                   134   UpfData
            code                     1216  InstalledCode
            parameters               4523  Dict
            settings                 4524  Dict
            parallelization          4525  Dict
        kpoints_distance             4526  Float
        kpoints_force_parity         4527  Bool
[...]

PR #1368, opened 2025-08-12
Inject default codes into `ResourceSettingsModel`

edan-bainglass:
This feature was unintentially removed in the MVC redesign. Restored here.


superstar54 (approved):
LGTM!

src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -106,7 +111,6 @@ def _render(self):
             description="Plugin store",
             link="plugin_manager.ipynb",
             icon="puzzle-piece",  # More intuitive icon
-            button_style="primary",  # Keeps it prominent
```
edan-bainglass, line 109: This didn't actually do anythin

src/aiidalab_qe/app/result/components/viewer/viewer.py
```diff
@@ -92,7 +92,4 @@ def _fetch_plugin_results(self, viewer_model: WorkChainResultsViewerModel):
             panel = entry["panel"]
             model = entry["model"]()
             viewer_model.add_model(identifier, model)
-            self.panels[identifier] = panel(
-                identifier=identifier,
```
edan-bainglass, line 96: `identifier` hasn't been used in some time now

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -81,16 +83,16 @@ def _render(self):
         super()._render()
 
         self.process_label = ipw.Text(
-            description="Label:",
-            layout=ipw.Layout(width="auto", indent="0px"),
```
edan-bainglass, line 85: `indent` is not a `Layout` argument and anyhow was not required.

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -205,7 +207,6 @@ def _on_qe_installed(self, _):
         self._toggle_qe_installation_widget()
         if self._model.qe_installed:
             self._model.update()
-            self._refresh_resources()
```
edan-bainglass, line 208: This was only necessary because the update was not done correctly

src/aiidalab_qe/app/submission/__init__.py
```diff
@@ -292,4 +295,4 @@ def _fetch_plugin_resource_settings(self):
 
             codes[identifier] = dict(model.get_models())
 
-        self.global_resources.set_up_codes(codes)
+        self.global_resources.build_global_codes(codes)
```
edan-bainglass, line 298: Hopefully a more clear name

src/aiidalab_qe/common/code/model.py
edan-bainglass, line 1: The changes in this file are the essence of the warning system we now provide to the user if the default code is invalid for any reason. Setting selected to `None` triggers the step blocker.

src/aiidalab_qe/common/widgets.py
```diff
@@ -693,18 +693,29 @@ def __init__(self, **kwargs):
...
+            description=kwargs.pop("description", None),
+            default_calc_job_plugin=kwargs.pop("default_calc_job_plugin", None),
             include_setup_widget=False,
             fetch_codes=True,  # TODO resolve testing issues when set to `False`
             **kwargs,
         )
         self.code_selection.layout.width = "80%"
 
         self.num_nodes = ipw.BoundedIntText(
-            value=1, step=1, min=1, max=1000, description="Nodes", width="10%"
```
edan-bainglass, line 703: `width` is not a thing for this widget

tests/conftest.py
```diff
@@ -422,20 +396,6 @@ def app(pw_code, dos_code, projwfc_code, projwfc_bands_code):
...
-    global_model.get_model("quantumespresso__projwfc").activate()
-
-    global_model.set_selected_codes(
-        {
-            "pw": {"code": pw_code.label},
-            "dos": {"code": dos_code.label},
-            "projwfc": {"code": projwfc_code.label},
-        }
-    )
-
```
edan-bainglass, line 438: This was only necessary due to incorrect implementation. No longer needed 🙂

tests/conftest.py
```diff
@@ -549,26 +504,22 @@ def _generate_workchain(process_class, inputs):
 @pytest.fixture
 def generate_pdos_workchain(
     fixture_localhost,
-    fixture_code,
+    pw_code,
+    dos_code,
+    projwfc_code,
```
edan-bainglass, line 509: We have test codes. Let's use them 🙂

tests/conftest.py
```diff
@@ -812,22 +754,23 @@ def _generate_qeapp_workchain(
...
+            )
 
         if run_bands:
             # Setting up inputs for bands_projwfc
-            inputs["bands"]["bands_projwfc"]["scf"]["pw"] = deepcopy(
+            inputs["bands"]["bands_projwfc"]["scf"]["pw"] = shallow_copy_nested_dict(
                 inputs["bands"]["bands"]["scf"]["pw"]
             )
-            inputs["bands"]["bands_projwfc"]["bands"]["pw"] = deepcopy(
+            inputs["bands"]["bands_projwfc"]["bands"]["pw"] = shallow_copy_nested_dict(
```
edan-bainglass, line 773: @AndresOrtegaGuerrero (pinging you as this is your code, but also to all other developers), whenever you feel like `deepcopy`ing something that includes an AiiDA node, pause and rethink, as it will create a new node (see #1341 for details). `shallow_copy_nested_dict` will achieve the deep copy without recreating the nodes. This was a royal pain to detect. Thanks @superstar54 for assisting with this one 🙏

tests/conftest.py
```diff
@@ -460,8 +420,9 @@ def _submit_app_generator(
         initial_magnetic_moments=0.0,
         electron_maxstep=80,
     ):
-        # Settings
-        app.configure_model.input_structure = generate_structure_data()
+        app.structure_model.input_structure = generate_structure_data()
+        app.structure_model.confirm()
+
```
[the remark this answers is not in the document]
superstar54: I think confirming step 1 is not needed for this test. But, yes, add it to make it more consistent.

src/aiidalab_qe/app/submission/global_settings/setting.py
```diff
@@ -73,13 +74,26 @@ def set_up_codes(self, codes: PluginCodes):
                         self._on_code_selection_change,
                         "selected",
                     )
+                    self.register_code(base_code_model)
+                    if base_code_model.default_calc_job_plugin == "quantumespresso.pw":
```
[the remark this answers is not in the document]
edan-bainglass: I'm not sure. This is something you added during the refactor of step 3. It appears to be only implemented for the global resources, whereas `register_code` is general for all resource panels. In principle, however, one would argue that a change in pw codes in any panel should prompt a resource check. However, this is not yet implemented I believe. Perhaps open an issue?

src/aiidalab_qe/common/panel.py
```diff
@@ -264,6 +275,27 @@ def __init__(self, model, **kwargs):
         super().__init__(model, **kwargs)
         self.code_widgets = {}
 
+    def register_code(self, code_model: CodeModel):
```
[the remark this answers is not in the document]
edan-bainglass: Potential names:
1. `register_code_events`
2. `register_code_trait_callbacks`
3. `observe_code`

Not sure which is better, if any.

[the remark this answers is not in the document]
edan-bainglass: Proceeding with option 2 following internal discussion

src/aiidalab_qe/app/submission/global_settings/setting.py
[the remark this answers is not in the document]
superstar54: Yes, it only checks the global resources and only `pw` for the moment. We should support check any codes also in the plugins in the future, good to open a PR.

PR #1370, opened 2025-08-13
Manually handle step advancement

edan-bainglass:
By relying on the AWB wizard step auto advancement feature, we were hitting inconsistent states, where a given step's `_post_render` method didn't actually have the required state variable (e.g., step 3 rendered prior to having access to the input parameters of step 2). This PR corrects this behavior by disabling auto advance and handling step advancement in `_on_<step>_confirmation` handlers.

### DIFF
--- a/src/aiidalab_qe/app/wizard_app.py
+++ b/src/aiidalab_qe/app/wizard_app.py
@@ -35,16 +35,13 @@ class WizardApp(ipw.VBox):
         # Create the application steps
         self.structure_step = StructureSelectionStep(
             model=self.structure_model,
-            auto_advance=True,
             auto_setup=auto_setup,
         )
         self.configure_step = ConfigureQeAppWorkChainStep(
             model=self.configure_model,
-            auto_advance=True,
         )
         self.submit_step = SubmitQeAppWorkChainStep(
             model=self.submit_model,
-            auto_advance=True,
             auto_setup=auto_setup,
         )
         self.results_step = ViewQeAppWorkChainStatusAndResultsStep(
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/app/wizard_app.py
```diff
@@ -131,12 +128,17 @@ def _on_step_change(self, change):
 
     def _on_structure_confirmation_change(self, _):
         self._update_configuration_step()
+        if not self.process:
+            self._wizard_app_widget.selected_index = 1
```
superstar54, line 132: Instead of hard-coding the index, better to increase the index by 1. Same for others.
```suggestion
            self._wizard_app_widget.selected_index = self._wizard_app_widget.selected_index + 1
```

src/aiidalab_qe/app/wizard_app.py
```diff
@@ -35,16 +35,13 @@ def __init__(self, auto_setup=True, **kwargs):
         # Create the application steps
         self.structure_step = StructureSelectionStep(
             model=self.structure_model,
-            auto_advance=True,
```
superstar54, line 38: What happens when loading the finished job, where we do not confirm manually? Or do we confirm automatically?

superstar54:
> It's just a timing issue. The event that moves you to the next step fires prior to the event that synchronizes the dependencies from the previous step.

Does this mean the `auto-advanced` is not reliable? Could you open an issue on AWB?

edan-bainglass:
> > It's just a timing issue. The event that moves you to the next step fires prior to the event that synchronizes the dependencies from the previous step.
> 
> Does this mean the `auto-advanced` is not reliable? Could you open an issue on AWB?

Not sure if I can conclude that yet. Will have a closer look after lunch.

edan-bainglass:
There is a simpler way to address the underlying behavior. Proceeding with it and closing this for now.

PR #1373, opened 2025-08-15
Add missing functional setting on process load

edan-bainglass:
Missed setting the functional in one of the two if-blocks. Also moved out setting the functional list (used to ensure functional consistency) out of the "pseudos" dictionary if-block (should always run).


AndresOrtegaGuerrero:
It looks good to me, any change that the current test uses PBE (non default) instead of PBESol to test the model?

PR #1375, opened 2025-08-20
Override traitlets behavior to improve static type checking

edan-bainglass:
`traitlets`, and by extension, `ipywidgets`, eventually fall back on `traitlets.HasDescriptors`, which implements
```python
def __new__(*args: t.Any, **kwargs: t.Any) -> t.Any: ...
```
causing any subclass to fail static type checking (VS Code at least sees all as `t.Any`).

This PR overrides this behavior in two ways:

1. Introducing a `HasTraits` utility overriding the `__new__` method returning `Self`

  ```python
  class HasTraits(tl.HasTraits):
      """Wrapper on `tl.HasTraits` for improved static type checking."""
  
      # For IDE type checking
      # Type checking struggles with `traitlets.HasTraits`, which inherits
      # `traitlets.HasDescriptors`, which in turn defines `__new__(...) -> t.Any`
      def __new__(cls, *args: t.Any, **kwargs: t.Any) -> Self:
          return super().__new__(cls, *args, **kwargs)
  ```

2. Injecting a similar `__new__` wrapper method in the root `Panel` class.

This covers all panels and models. This PR does not address ipywidgets widgets and our own custom widgets.


superstar54 (approved):
LGTM!

PR #1376, opened 2025-08-20
Lift model dependency register to root `Model` class

edan-bainglass:
Based on #1375

In principle, any `Model` may have dependencies, not just `SettingsModel`.

### DIFF
--- a/src/aiidalab_qe/common/mixins.py
+++ b/src/aiidalab_qe/common/mixins.py
@@ -10,8 +10,6 @@ from aiida_quantumespresso.data.hubbard_structure import HubbardStructureData
 from aiidalab_qe.common.mvc import Model
 from aiidalab_qe.utils import HasTraits
 
-T = t.TypeVar("T")
-
 
 class HasInputStructure(HasTraits):
     input_structure = tl.Union(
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/common/mixins.py
```diff
@@ -71,21 +73,22 @@ def _link_model(self, model: T):
...
-                sibling, trait = dependency_parts
-                target_model = self.get_model(sibling)
-            tl.dlink(
-                (target_model, trait),
-                (model, trait),
-            )
-
-
-class HasProcess(tl.HasTraits):
+        if isinstance(model, Model):
```
superstar54, line 76: Is this check necessary? What if it is not a model, should any error be raised?

edan-bainglass: Yeah, I should handle it differently. There is no need for the check because the model should be typed correctly as a Model. It is typed as a generic T. Maybe I can bound it to Model somehow. But yes, in principle, since this is Python, one can pass whatever they want. We should raise an error if they pass something that is not a model.

src/aiidalab_qe/common/mixins.py
```diff
@@ -37,36 +36,43 @@ def has_tags(self):
         )
 
 
-class HasModels(t.Generic[T]):
+M = t.TypeVar("M", bound=Model)
```
edan-bainglass, line 39: Now bounded to `Model`

src/aiidalab_qe/common/mixins.py
```diff
@@ -37,36 +36,43 @@ def has_tags(self):
...
 
-    def get_models(self) -> t.Iterable[tuple[str, T]]:
+    def get_models(self) -> t.Iterable[tuple[str, M]]:
         return self._models.items()
 
-    def _link_model(self, model: T):
-        if not hasattr(model, "dependencies"):
-            return
+    def _link_model(self, model: M):
+        assert isinstance(model, Model), "HasModels only works with Model instances"
```
edan-bainglass, line 75: If the user sees this, they should report - bad developer!

superstar54 (approved):
LGTM!

PR #1379, opened 2025-08-22
Update pseudopotentials panel

edan-bainglass:
Per team discussion, this PR does the following:

- [x] Allow full relativistic pseudopotentials for non-SOC calculations
- [x] Mark PseudoDojo family selectors with SR/FR
- ~~Add checkbox to include FR families when in non-SOC state~~
- [x] Allow uploading unique-md5 pseudos of existing filenames
- [x] Notify user post-pseudo-upload that switching the protocol will also reset all pseudos (undo upload)

### Comment regarding FR checkbox

~~Unfortunately, `ipywidgets.ToggleButtons` is designed to auto-select the first option on change of options. Hence, updating the family options on checkbox state change results in a reset of the current selected family to the first available option. There is no apparent public API (or hacking path) for blocking this. One can save the previous library and reset it after, but the UX is not great, as the user sees the bounce between current to first and back to current (if available). Not great, but doable if team requests it. Pinging @giovannipizzi.~~

#### Update

The checkbox has been removed in favor of always showing the FR families. The text has been adjusted. @AndresOrtegaGuerrero if you find issues with this, we can consider restoring the full relativistic checkbox. But great if you can provide a reason for not including them by default.

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/pseudos/pseudos.py
+++ b/src/aiidalab_qe/app/configuration/advanced/pseudos/pseudos.py
@@ -73,15 +73,6 @@ class PseudosConfigurationSettingsPanel(
             (self.family_help, "value"),
         )
 
-        self.functional_help = ipw.HTML("""
-            <div class="pseudo-text">
-                The exchange-correlation energy is calculated using this functional.
-                <br>
-                We currently provide support for two well-established generalized
-                gradient approximation (GGA) functionals: PBE and PBEsol.
-            </div>
-        """)
-
         self.functional = ipw.ToggleButtons()
         ipw.dlink(
             (self._model, "functional_options"),
@@ -92,7 +83,7 @@ class PseudosConfigurationSettingsPanel(
             (self.functional, "value"),
         )
 
-        self.library = ipw.ToggleButtons()
+        self.library = ipw.ToggleButtons(style={"button_width": "fit-content"})
         ipw.dlink(
             (self._model, "library_options"),
             (self.library, "options"),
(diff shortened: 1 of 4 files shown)


superstar54:
> Allow uploading unique-md5 pseudos of existing filenames

Could you point out which changes did this? I can not find the relevant code changes.

edan-bainglass:
> > Allow uploading unique-md5 pseudos of existing filenames
> 
> Could you point out which changes did this? I can not find the relevant code changes.

[9e3d415](https://github.com/aiidalab/aiidalab-qe/pull/1379/commits/9e3d4159593e96bcaae2c9412d76fd18df83a0e6)

src/aiidalab_qe/app/configuration/advanced/pseudos/pseudos.py
```diff
@@ -136,6 +141,9 @@ def render(self):
                     <br>
                     To reset to protocol-derived pseudopotentials, select both an
                     exchange-correlation functional and a pseudopotential family above
+                    <br>
```
AndresOrtegaGuerrero, line 144: <br> To reset to protocol defaults, select both an exchange–correlation functional and a pseudopotential family above. <br> Switching the protocol in <b>Basic settings</b> will also reset pseudopotentials to that protocol’s defaults.

superstar54 (approved):
LGTM!

AndresOrtegaGuerrero:
does it look excesive ? 
image
I was thinking about the current horizontal layout and wondered if a dropdown menu might be a better approach. I remember we discussed not expanding things horizontally, at it seems we are again on that path, and with the potential addition of SG15 pseudos in the future ( have both SR and FR), this layout could become too long.

AndresOrtegaGuerrero:
Another point is that Pseudo dojo has different versions for SR (0.5 0.41) while FR (0.4), I think this information is missing , and displaying both SR and FR without the version could be confusing (since this part is basically and advanced feature) so the users that check this part care about these things

edan-bainglass:
Thanks for your comments @AndresOrtegaGuerrero. I agree that the horizontal layout it is not ideal. As you mentioned, we have already discussed switching to dropdown selectors, which can support additional options without burdening the UI. However, such a change may warrant some discussion with the team. Maybe we can revisit the layout when SG15 is introduced.

For now, regarding the SR/FR versions, could we perhaps add it to the accompanying help text? If so, can you please suggest the exact wording? 🙏

AndresOrtegaGuerrero:
The AiiDA pseudo-family contains the version of pseudo installed. However, since the user can download the pseudopotentials, this is acceptable. I don't have a strong opinion on where this text should be located, as it also applies to SSSP. I believe the current version is 1.3. Perhaps you could add the version number, like "v1.3" or "v0.4", next to the scalar text. You can proceed with the merge, as the desired behavior, which was discussed in our meeting with Giovanni, has been implemented. Thank you!

PR #1384, opened 2025-08-26
Fix pseudo dojo library handling on process upload

edan-bainglass:
Some logic was missed in handling the relativistic information when loading the app from a process.


edan-bainglass:
Thanks for reporting this @AndresOrtegaGuerrero. Let's get this merged ASAP 🙏 I will PR additional tests in a follow-up.

AndresOrtegaGuerrero:
Since is a model logic, should it be a logic to test upload and so on , so for future changes the logic is not changed ?

edan-bainglass:
> Since is a model logic, should it be a logic to test upload and so on , so for future changes the logic is not changed ?

I'm not sure I understand what you mean. We test these things already.

AndresOrtegaGuerrero:
I tested and now the bug is not there anymore

PR #1387, opened 2025-08-27
Simplify model network hierarchy

edan-bainglass:
Based on #1388 

Some features of the `SettingsModel` and its `ConfigurationSettingsModel`
are arguably more general and thus are lifted here to the `PanelModel`, and
in some cases, higher up to the root `Model`. Consequently, `SettingsModel`
(and panel) and `ConfigurationSettingsModel` are dropped.

Moreover, the entire advanced sub-settings module is discarded following
similar reasoning.


edan-bainglass:
@superstar54 @AndresOrtegaGuerrero @mikibonacci since plugins extend `ConfigurationSettingsModel`, I left a dummy class behind with a deprecation warning. Please update your plugins 🙏

src/aiidalab_qe/app/configuration/__init__.py
```diff
@@ -45,10 +42,6 @@ def __init__(self, model: ConfigurationStepModel, **kwargs):
             },
             **kwargs,
         )
-        self._model.observe(
-            self._on_input_structure_change,
-            "input_structure",
-        )
```
edan-bainglass, line 51: `model` observations should in general come AFTER model registration (`add_model` calls) to ensure proper order of events.

src/aiidalab_qe/app/configuration/advanced/model.py
```diff
@@ -211,26 +209,22 @@ def set_model_state(self, parameters):
...
                 hubbard.set_active_eigenvalues(starting_ns_eigenvalue)
 
-    def _link_model(self, model: AdvancedCalculationSubSettingsModel):
-        ipw.dlink(
-            (self, "loaded_from_process"),
-            (model, "loaded_from_process"),
-        )
-        model.observe(
-            self._on_any_change,
-            tl.All,
```
edan-bainglass, line 233: It was wrong to attach the `_on_any_change` handler this way. The advanced model only needs to link the `confirmed` trait to/from its sub-models, as is done in the configuration step, linking the `confirmed` state of its panel models.

src/aiidalab_qe/app/configuration/advanced/subsettings.py
edan-bainglass, line 1: Redundant!

src/aiidalab_qe/common/mixins.py
```diff
@@ -73,10 +73,14 @@ def get_models(self) -> t.Iterable[tuple[str, M]]:
 
     def _link_model(self, model: M):
         assert isinstance(model, Model), "HasModels only works with Model instances"
+        tl.dlink(
+            (self, "locked"),
+            (model, "locked"),
+        )
```
edan-bainglass, line 79: This trait (formerly `loaded_from_process`) is now at the `Model` level, so we can already link it here. This is actually appropriate, allowing us (in the future) to lock the entire app (minus the Results step) on process load.

superstar54 (changes requested):
Thanks for the work!

Let's move some key comments from the GitHub page directly into the code. This will benefit anyone reading the code in the future. See my suggestions below.

PR #1388, opened 2025-08-30
Discard `traitlets` typing hack

edan-bainglass:
This PR manually reverts #1375.

#1375 introduced a hack to recover static analysis for ipywidgets/traitlets classes. Further research suggests that this was an issue due to a "recent" release of the `pyright` package (used by `pylance`). This is addressed by https://github.com/ipython/traitlets/pull/918. For the time being, downgrading `pylance` to `2024.12.1` restores static typing. Once `traitlets` releases a new version, we should consider upgrading, for the fix, but also to stay up-to-date if possible. If and when we upgrade, `pylance` may be updated as well.


superstar54 (approved):
LGTM!

PR #1389, opened 2025-09-03
Avoid storing nodes in frontend - use UUIDs

edan-bainglass:
We already use the process UUID in the results step to avoid DB session conflicts across threads. This PR extends this principle throughout the app, allowing a broader use of threads in the future.

In particular, the `HasInputStructure` and `HasProcess` mixins now operate with uuids. A property is introduced to each referencing the node. These properties are now cached using a new `cache_per_thread` decorator ensuring that each thread only loads the node once. Subsequent calls do not bother the DB. The cache is invalidated explicitly by the respective UUID.

### Breaking change

Plugins need to switch referencing the old "input_structure" trait to "structure_uuid"

### Update

Removing node caching - node loading is efficient


src/aiidalab_qe/app/configuration/advanced/hubbard/model.py
```diff
@@ -95,7 +95,7 @@ def set_active_eigenvalues(self, eigenvalues: list):
         self.has_eigenvalues = True
 
     def get_parameters_from_hubbard_structure(self):
-        hubbard_parameters = self.input_structure.hubbard.dict()["parameters"]
+        hubbard_parameters = self.input_structure.hubbard.model_dump()["parameters"]
```
edan-bainglass, line 98: `dict` is deprecated in favor of `model_dump` in pydantic V2 🙂

src/aiidalab_qe/common/decorators.py
edan-bainglass, line 1: Thread safe caching decorator. @superstar54 give this one in particular a closer look 🙏

PR #1397, opened 2025-09-17
Addign editor  to fix atomic positions in x,y,z and to add constraints

cpignedoli:
connected to https://github.com/aiidalab/aiidalab-widgets-base/pull/706
Constraints on x,y,z for each atom are stored in the ASE array 'fixed_atoms' .
image
image
The additional column is not present in case no constraints were specified

I need help (I would prefer in a separate PR), e.g. @edan-bainglass @superstar54, to add some warning logics:
If I load a structure that has constraints and I try to do phonons-related properties, I should get a warning. The warning can disappear only if 1)I remove the constraints, and 2) I do an additional geo-opt.

We should think about how to make this editor more general, including other [types of constraints](https://www.quantum-espresso.org/Doc/INPUT_PW.html#idm1633)

### DIFF
--- a/src/aiidalab_qe/app/result/components/summary/model.py
+++ b/src/aiidalab_qe/app/result/components/summary/model.py
@@ -193,13 +193,17 @@ class WorkChainSummaryModel(ResultsComponentModel):
             "cell_lengths": "{:.3f} {:.3f} {:.3f}".format(*structure.cell_lengths),
             "cell_angles": "{:.0f} {:.0f} {:.0f}".format(*structure.cell_angles),
         }
+        
 
         relax_value_mapping = {
             "none": "off",
             "positions": "atomic positions",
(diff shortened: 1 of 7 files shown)


superstar54 (changes requested):
Thanks for adding this feature!
I have a quick run on the app.
- Change the tab name "Edit Constrains" to "Constrains" directly, and same for the other tabs, because they already in the "Edit structure" section.
- If we also support other type of contrains, better to move current code in a subsection (or tab), e.g., "Fix positions"
- Would be nice to have a table, with atom index, and contraints (x, y, z), which the user can edit/save/reset directly.
- If the `angle`, and `distance` contrains are not supported by `aiida-quantumespresso`, we remove them in the text.
- in my tests, the contrains are not shown in the table of the results panel.
- add tests

PR #1400, opened 2025-10-01
Implement app duplication feature (and more)

edan-bainglass:
This PR introduces the app duplication feature. Users can now click on the top level Duplicate button to start a new calculation from the state of their current one. Much was required first prior to this being possible, so this is a large PR. However, it might be too large. I will review it first, adding comments where needed. Then happy to discuss if parts should be removed, or introduced in smaller PRs.

### Update

This PR is too big and does too much 😅 Converting it to a draft and keeping it open for reference as I break it down to smaller, focused PRs.


edan-bainglass:
Superseded by #1429

PR #1417, opened 2025-11-20
Fix step state updating in structure step

### LINKED ISSUE #1416: Loading a structure doesnt alllow to confirm the structure
In the main branch since https://github.com/aiidalab/aiidalab-qe/pull/1389[)](https://github.com/aiidalab/aiidalab-qe/commit/fae0013bf2130f593a9dfc3950215ec91549516f)
[edan-bainglass](https://github.com/aiidalab/aiidalab-qe/commits?author=edan-bainglass)

The app doesnt allow to confirm a structure if this one doesnt have structure_uuid ,like uploading a file or selecting an example. 

That change will produce a break since all the external plugins should be updated , and sync with the version. 
It would be better to revert this , or to allow the input_structure with the new structure_uuid, and display a warning deprecation. 

It would also good to include test for the confirming status button for selecting a aiida StructureData, or a ase structure data. 

@edan-bainglass

edan-bainglass:
Fixes #1416

### DIFF
--- a/src/aiidalab_qe/app/structure/__init__.py
+++ b/src/aiidalab_qe/app/structure/__init__.py
@@ -224,7 +224,7 @@ class StructureSelectionStep(QeConfirmableWizardStep[StructureStepModel]):
     def _update_state(self):
         if self._model.confirmed:
             self.state = self.State.SUCCESS
-        elif not self._model.has_structure:
-            self.state = self.State.READY
-        else:
+        elif self._model.structure_uuid:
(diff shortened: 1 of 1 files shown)


edan-bainglass:
@AndresOrtegaGuerrero please check if this resolves also for the CIF uploader 🙏

AndresOrtegaGuerrero:
We should a test for the logic of the confirm button , so in a future change this is not missed.

PR #1420, opened 2025-11-20
Update archive downloader w.r.t changes in aiida-core

### LINKED ISSUE #1419: Can not download raw data
**Describe the bug**
When I try to save the results of the calculation, the button to save the raw data is inactive, and the message says:
`The raw data download is not available because the AiiDA version is too old.` (see screenshot)

I have a very recent verion of Aiidalab (downloaded ~1 Nov 2025), AiiDA version 2.7.1

**To Reproduce**
- go to calculations history
- download link

**Expected behavior**
I expect to be able to download the raw data of the calculation, but the button is inactive and the error message is not reasonable.

**Screenshots**
Image

Image

Image

**Version (if known)**
I run aiidalab on windows in docker through WSL
 - AiiDA version 2.7.1
 - docker image ID sha256:ea1aba4327c35ace32b97d2c5b9e9d743a95ef03511ebbc0834cd9cc0f4e9d52

**Additional context**
This happens with all calculations, both successful and failed, my own or examples.

edan-bainglass:
Fixes #1419

https://github.com/aiidateam/aiida-core/pull/6723 moved `ProcessDumper` behind a facade import and deprecated it in favor of a new `.dump()` API on `ProcessNode`. This PR uses the facade import to check for availability and the `.dump()` method for the actual dump operation.

### DIFF
--- a/src/aiidalab_qe/app/result/components/summary/download_data.py
+++ b/src/aiidalab_qe/app/result/components/summary/download_data.py
@@ -2,6 +2,7 @@ import base64
 import pathlib
 import tempfile
 from threading import Thread
+from typing import cast
 
 import ipywidgets as ipw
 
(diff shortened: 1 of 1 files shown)


giovannipizzi (approved):
As a note for another PR, the function is called `produce_bitestream`, but I guess it's either `bytestream` or `bitstream`

PR #1421, opened 2025-11-20
Image cleanup

edan-bainglass:
This PR attempts to add some cleanup prior to image build and reduce cache explosion to reduce image size, which is currently failing all tests due to size limit exceeding.

### DIFF
--- a/.github/workflows/docker-build-test.yml
+++ b/.github/workflows/docker-build-test.yml
@@ -70,10 +70,6 @@ jobs:
                   push: ${{ ! (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork) }}
                   context: .
                   platforms: linux/amd64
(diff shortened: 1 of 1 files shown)


danielhollas (commented):
I haven't run into this problem before. Just one quick thought, it's perhaps worth trying to bump `docker/build-push-action@v5` to `docker/build-push-action@v6`. 
Otherwise the most straightforward way is to simply disable caching (to at least see if that fixes things?).

I also think somebody should invest time in addressing the root problem and slim down the image. Last time I looked there was possibly a lot of low-hanging fruit, I started looking into it in #1337, please have a look at that PR (I've revived it today and made a bit of progress but will not have time to pursue it tomorrow).

danielhollas:
@edan-bainglass I think this one can be closed?

PR #1422, opened 2025-11-21
Decouple build and integration tests

edan-bainglass:
This PR aims to separate out Chrome Selenium tests to its own job, making the build process depend on it.


danielhollas:
I mean it's the same way that works in AWB. The selenium tests run by launching the app via the Docker image (in the same way as you would launch the app locally).
(agree that should probably be documented better somewhere)

PR #1424, opened 2025-11-23
Step enhancement (prelim)

edan-bainglass:
Based on #1337 

This PR serves as preliminary work to an overall wizard-step improvement work. It does the following:
1. Refactor wizard steps to dedicated "step.py" files
2. Simplify class naming

### DON'T SQUASH-MERGE 🙏

The refactor to step.py files is done in two commits - the first renames the __init__.py files to step.py, and the second reintroduces the \_\_init__.py files. We MUST merge both (**don't squash merge**) to ensure history tracking.


mikibonacci (approved):
LGTM!

edan-bainglass:
Hadn't noticed that this was being PR'd into another branch, not `main`. Will fix this

PR #1426, opened 2025-11-26
Simplify step class naming and imports

edan-bainglass:
Some class names are overly complex. Simplifying a bit here.


mikibonacci (approved):
ok, LGTM!

Issue #1372, opened 2025-08-14
Possibly not handling threads correctly in results step


edan-bainglass:
Possibly all tests of the results step end with the following:

```python
  threadexception.py:73: PytestUnhandledThreadExceptionWarning: Exception in thread Thread-1
  
  Traceback (most recent call last):
    File "/opt/conda/lib/python3.9/threading.py", line 980, in _bootstrap_inner
      self.run()
    File "/opt/conda/lib/python3.9/threading.py", line 917, in run
      self._target(*self._args, **self._kwargs)
    File "/home/jovyan/.local/lib/python3.9/site-packages/aiidalab_widgets_base/process.py", line 763, in _monitor_process
      process = orm.load_node(process_uuid)
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/utils/loaders.py", line 207, in load_node
      return load_entity(
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/utils/loaders.py", line 93, in load_entity
      return entity_loader.load_entity(
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/utils/loaders.py", line 399, in load_entity
      entity = builder.one()[0]
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/querybuilder.py", line 1140, in one
      res = self.all()
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/querybuilder.py", line 1120, in all
      matches = list(self.iterall(batch_size=batch_size))
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/querybuilder.py", line 1086, in iterall
      item[i] = self._get_aiida_entity_res(item_entry)
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/querybuilder.py", line 1031, in _get_aiida_entity_res
      return convert.get_orm_entity(value)
    File "/opt/conda/lib/python3.9/functools.py", line 888, in wrapper
      return dispatch(args[0].__class__)(*args, **kw)
    File "/opt/conda/lib/python3.9/site-packages/aiida/orm/convert.py", line 118, in _
      node_class = load_node_class(backend_entity.node_type)
    File "/opt/conda/lib/python3.9/site-packages/aiida/storage/psql_dos/orm/nodes.py", line 122, in node_type
      return self.model.node_type
    File "/opt/conda/lib/python3.9/site-packages/aiida/storage/psql_dos/orm/utils.py", line 81, in __getattr__
[...]

PR #1427, opened 2025-11-26
Apply MVC to `Wizard` widget

edan-bainglass:
Based on #1465 

This PR extends the MVC pattern to the top-level main, app, and wizard components. Along the way, it decouples the wizard from AWB. The AWB wizard widget is not MVC compatible. Here we cast the QE app wizard as a standalone, MVC-compatible redesign of the AWB widget.

### DON'T SQUASH-MERGE 🙏

Refactoring the wizard and wrapper requires all commits to maintain history.


edan-bainglass:
@mikibonacci this one's next. No rush. This one's larger. I'd appreciate a closer look when you have time 🙏

mikibonacci (approved):
LGTM!

edan-bainglass:
> You've hit GitHub actions timeout (30 minutes). So some test must be stuck.

Fixed. Looks like one of my tests left the process monitor thread alive, holding up the test suite. Implemented cleanup for the relevant parts.

However, I've seen these sort of issues before. See https://github.com/aiidalab/aiidalab-qe/issues/1372

edan-bainglass:
@danielhollas @mikibonacci mind given this one more look, particularly the last 4 commits (added post-approval)? 🙏🙏

tests/conftest.py
```diff
@@ -60,6 +60,36 @@
 }
 
 
+def _terminate_process_monitor(wizard: Wizard | None):
+    from contextlib import suppress
+
+    if wizard is None:
+        return
+
+    with suppress(Exception):
```
danielhollas: Maybe comment on why are we suppressing exceptions in function?

src/aiidalab_qe/common/widgets.py
```diff
@@ -403,12 +403,12 @@ def _start_animate(self):
         if self._animate_thread is not None:
             raise RuntimeError("Cannot start animation more than once!")
 
-        self._animate_thread = Thread(target=self._animate)
```
danielhollas, line 406: Probably not for this PR, but would highly recommend getting rid of the animation thread -- imo its silly to do animation in the backend, and more importantly, the spinning thread can lead to a memory leak when the frontend disconnects (e.g. when user's browser crashes, or goes to sleep). This was an issue in the original WizardAppWidget as well.

See https://github.com/aiidalab/aiidalab-widgets-base/pull/479 and https://github.com/aiidalab/issues/issues/13

danielhollas (commented):
@edan-bainglass sorry I will not have time to delve into this PR at the moment. I did leave some comments from briefly looking at the last commits (haven't seen anything else that would stand out to me).

edan-bainglass:
> @edan-bainglass sorry I will not have time to delve into this PR at the moment. I did leave some comments from briefly looking at the last commits (haven't seen anything else that would stand out to me).

Thanks @danielhollas. I meant specifically only the last commits. The rest @mikibonacci has already reviewed 🙂 Cheers

src/aiidalab_qe/common/widgets.py
edan-bainglass: Thanks for the references @danielhollas. These sort of memory leaks is precisely what I was concerned about at why I opened #1372. Note that it is even the home page that does not get culled once its browser window is closed. Need to investigate in general our use of threads.

edan-bainglass: My first observation is that opening a home or app notebook introduces to `ps -aux` a new instance of the following:

```
/opt/conda/bin/python -m ipykernel_launcher -f /home/jovyan/.local/share/jupyter/runtime/kernel-03c9b25e-9436-4d7a-82c5-e964e3c44e35.json
```

These remain unless you force a shutdown of the kernel from the task manager.

tests/conftest.py
edan-bainglass: The idea here is to attempt serveral ways to shut down the process monitoring thread without failing the test, which already passed at this point (this code is run post test assertions). I'll have another look at it to see if one of the various shut down attempts consistently works already and attempt to fine-grain the exception to supress (likely `StorageClosed`, which appears to be the root issue for the thread). But I won't go to far beyond this, as it is already on the agenda to resolve it properly. It is unrelated to this PR (these threads existed prior).

edan-bainglass: I removed the defensive exception suppression and reduced the thread shut down attempts to just setting the process uuid to `None` (signals monitor shut down via traitlets). The hang appears to still be consistently gone, and also there are no `StorageClosed` exceptions raised. I conclude for now that the exception was raised due to us not shutting down the monitor post testing, causing the next monitor callback check to hit a closed storage. I will anyhow continue to monitor this (pun intended!).

tests/conftest.py
```diff
@@ -401,12 +385,11 @@ def app(pw_code, dos_code, projwfc_code):
 
     yield app
 
+    app.results_model.process_uuid = None  # terminate monitor
```
edan-bainglass, line 388: This alone resolves the hanging thread issue

tests/test_app.py
```diff
@@ -1,28 +1,88 @@
...
+        assert wizard.configure_model.relax_type == "positions"
+        assert wizard.configure_model.get_model("workchain").spin_type == "collinear"
+        assert wizard.configure_model.get_model("bands").include is True
+        assert wizard.configure_model.get_model("pdos").include is False
+        assert wizard.configure_model.state == State.SUCCESS
+        advanced_model = wizard.configure_model.get_model("advanced")
+        pseudos_model = advanced_model.get_model("pseudos")
+        assert len(pseudos_model.dictionary) > 0
+        assert pseudos_model.functional == "PBE"
+        self.controller.wizard.results_model.process_uuid = None  # terminate monitor
```
edan-bainglass, line 34: Since in `TestApp` we explicitly test the app, we generate our own MVC. Here we must also shutdown this MVC's proess monitor thread.

edan-bainglass:
Okay @danielhollas, I've addressed your comments. May I merge? 🙏

tests/conftest.py
danielhollas: Nice investigation! Tbh I think I didn't even noticed that this was in the test code 😅 

Yeah, the process monitor thread is really tricky, and is also one of the remaining possibilities of memory leak in AWB, I think I eliminated all other background threads in AWB.

danielhollas (approved):
Thanks! 

(In general for qeapp don't feel blocked by my reviews, since I do not have a stake here :-) )

edan-bainglass:
> Thanks! 
> 
> (In general for qeapp don't feel blocked by my reviews, since I do not have a stake here :-) )

I value your input/feedback 🙏

PR #1428, opened 2025-11-26
Cleanup

edan-bainglass:
This PR compiles several driveby updates extracted from the app duplication work.

## Summary of changes

- Added new fields `scf_conv_thr_abs` and `etot_conv_thr_abs` to the `ConvergenceConfigurationSettingsModel`, and updated their calculation to be per-structure, improving clarity and accuracy in threshold display. The UI now shows these absolute values with units and only displays them when a structure is present.

- Refactored `set_model_state` in `AdvancedConfigurationSettingsModel` to directly set all relevant sub-models (general, convergence, smearing, pseudos, magnetization, hubbard) using context managers for trait notifications, and removed the now-redundant `_set_pw_parameters` method, simplifying and centralizing state initialization.

- Renamed and refactored widgets and methods in the magnetization panel for clarity (e.g., `kind_moment_widgets` to `moments_list`, `_build_kinds_widget` to `_build_moments_list`), and simplified the update logic to remove unnecessary checks and flags.

- Simplified the update logic in the Hubbard panel by removing the `updated` flag and associated logic, and added a `needs_eigenvalues_widget` property to the model for clearer widget display logic.

- Made all advanced panels and models importable from the advanced settings module directly.

- Renamed various methods/classes for clarity/brevity.

### DIFF
--- a/src/aiidalab_qe/app/result/components/viewer/structure/model.py
+++ b/src/aiidalab_qe/app/result/components/viewer/structure/model.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import contextlib
+
 import traitlets as tl
 from ase.formula import Formula
 
@@ -57,11 +59,10 @@ class StructureResultsModel(ResultsModel):
         self.selected_view = "relaxed" if self.selected_view == "initial" else "initial"
 
     def _get_structure(self) -> orm.StructureData | None:
-        try:
-            return self.source.structure if self.source else None
-        except AttributeError:
-            # If source is outputs but job failed, there may not be a structure
+        if not self.source:
             return None
+        with contextlib.suppress(AttributeError):
+            return self.source.structure if self.source else None
 
     def _get_structure_info(self):
         structure = self.structure
--- a/src/aiidalab_qe/app/result/components/viewer/structure/model.py
+++ b/src/aiidalab_qe/app/result/components/viewer/structure/model.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-import contextlib
-
 import traitlets as tl
 from ase.formula import Formula
 
@@ -59,10 +57,10 @@ class StructureResultsModel(ResultsModel):
         self.selected_view = "relaxed" if self.selected_view == "initial" else "initial"
 
     def _get_structure(self) -> orm.StructureData | None:
-        if not self.source:
-            return None
-        with contextlib.suppress(AttributeError):
+        try:
             return self.source.structure if self.source else None
+        except AttributeError:
+            return None
 
     def _get_structure_info(self):
         structure = self.structure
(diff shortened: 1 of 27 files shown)


edan-bainglass:
Ciao @mikibonacci. This one is next on the list 🙏 The PR description covers the scope of changes. Happy to answer any questions 🙂

mikibonacci:
Hey @edan-bainglass, I added two small comments, after them I think we can go

edan-bainglass:
> Hey @edan-bainglass, I added two small comments, after them I think we can go

You did? I don't see any comments 🤔

src/aiidalab_qe/app/result/components/viewer/structure/model.py
```diff
@@ -57,11 +59,10 @@ def toggle_selected_view(self):
         self.selected_view = "relaxed" if self.selected_view == "initial" else "initial"
 
     def _get_structure(self) -> orm.StructureData | None:
-        try:
-            return self.source.structure if self.source else None
-        except AttributeError:
-            # If source is outputs but job failed, there may not be a structure
+        if not self.source:
```
mikibonacci: do we need both this if and the contextlib below? after discussion, it seems that the reason was that self.source might be an `AttributeDict` and might be `{}`, so it's not `None` and this will trigger the `AttributeError`

src/aiidalab_qe/app/result/components/viewer/structure/model.py
```diff
@@ -57,11 +59,10 @@ def toggle_selected_view(self):
...
             return None
+        with contextlib.suppress(AttributeError):
+            return self.source.structure if self.source else None
```
edan-bainglass: Hi @mikibonacci. Responding to your vanishing comment. This check is just an extra early bail, but indeed, it may be redundant. For the sake of moving things along, let's keep it here for safety. I will revisit this in a later PR.

mikibonacci (approved):
LGTM!

src/aiidalab_qe/app/result/components/viewer/structure/model.py
edan-bainglass: Correction after closer inspection - if the structure is relaxed, we get the `structure` from the outputs (`self.source` is `process.outputs`), otherwise we get the initial structure from the inputs (`self.source` is `process.inputs`). The only way `self.source` is `None` is if there isn't a process. Otherwise, the only other exception is if the structure was relaxed, but the job failed, so `structure` is not found in the outputs. As such, the code block is simplified to just
```python
        try:
            return self.source.structure if self.source else None
        except AttributeError:
            return None
```
I reverted the use of `contexlib.suppress` (here and in a couple of mixin methods) to explicit `try/except` due to `mypy` complaining about no explicit `return None` statements.

PR #1429, opened 2025-11-26
Implement app duplication

edan-bainglass:
Based on #1428

Supersedes #1400 

This PR implement a button that on click:
1. Writes the current app state to file
2. Opens a new app instance, tagging the URL with `duplicating=True`
3. New instance reads the previous state from file, populates the wizard, and removes the URL marker

See some notes [here](https://github.com/aiidalab/aiidalab-qe/pull/1400#issuecomment-3497771077)


mikibonacci (approved):
LGTM!

Issue #1402, opened 2025-10-02
Error when loading a aiida node in another account


AndresOrtegaGuerrero:
```python
---------------------------------------------------------------------------
NotExistent                               Traceback (most recent call last)
/opt/conda/lib/python3.9/site-packages/aiida/orm/utils/loaders.py in load_entity(cls, identifier, identifier_type, sub_classes, query_with_dashes)
    398         try:
--> 399             entity = builder.one()[0]
    400         except MultipleObjectsError:

/opt/conda/lib/python3.9/site-packages/aiida/orm/querybuilder.py in one(self)
   1145         elif len(res) == 0:
-> 1146             raise NotExistent('No result was found')
   1147         return res[0]

NotExistent: No result was found

During handling of the above exception, another exception occurred:

NotExistent                               Traceback (most recent call last)
/tmp/ipykernel_2779444/1083177014.py in <cell line: 1>()
----> 1 app.load()

~/apps/quantum-espresso/src/aiidalab_qe/app/main.py in load(self)
     90         # load a previous calculation if it is provided
     91         if self.process:
---> 92             self.app.process = self.process

/opt/conda/lib/python3.9/site-packages/traitlets/traitlets.py in __set__(self, obj, value)
    730             raise TraitError('The "%s" trait is read-only.' % self.name)
    731         else:
--> 732             self.set(obj, value)
    733 
    734     def _validate(self, obj, value):

/opt/conda/lib/python3.9/site-packages/traitlets/traitlets.py in set(self, obj, value)
    719             # we explicitly compare silent to True just in case the equality
    720             # comparison above returns something other than True/False
--> 721             obj._notify_trait(self.name, old_value, new_value)
    722 
    723     def __set__(self, obj, value):

/opt/conda/lib/python3.9/site-packages/traitlets/traitlets.py in _notify_trait(self, name, old_value, new_value)
   1503 
   1504     def _notify_trait(self, name, old_value, new_value):
-> 1505         self.notify_change(
   1506             Bunch(
   1507                 name=name,

/opt/conda/lib/python3.9/site-packages/ipywidgets/widgets/widget.py in notify_change(self, change)
    692                 # Send new state to front-end
    693                 self.send_state(key=name)
--> 694         super(Widget, self).notify_change(change)
    695 
    696     def __repr__(self):

/opt/conda/lib/python3.9/site-packages/traitlets/traitlets.py in notify_change(self, change)
   1515     def notify_change(self, change):
[...]

PR #1438, opened 2025-12-01
Including a method for displaying the pp_uuid

AndresOrtegaGuerrero:
This solves #1402 

The issue occurs when loading a pp node from another account (Since the pp calc doesnt have any settings from pseudos)


edan-bainglass (approved):
Thanks for opening this @AndresOrtegaGuerrero. The opinion of this PR (or rather the last commit) is that a PP-only workflow should not include PW-related sections in the summary. However, we should consider instead to add a link in the PP plugin's results panel to a new app instance populated with the corresponding PW calculation.

If you approve of this PR, please confirm with a thumb up 🙏

PR #1451, opened 2026-01-19
Pin aiidalab-widgets-base to 2.5.0

### LINKED ISSUE #1450: Do not depend on the master branch of AWB
Currently, the app depends on the latest version of AWB, which is likely to break the app, because we are migrating aiidalab-widgets-base to ipywidgets 8.x.

```
# Due to a technical issue, we can not make a new release of aiidalab-widgets-base.
# This is a temporary solution to use the latest feature from AWB.
aiidalab-widgets-base[optimade] @ git+https://github.com/aiidalab/aiidalab-widgets-base@master
```
The AWB 2.5.0 was released; we should use it.

superstar54:
Pin aiidalab-widgets-base to 2.5.0

The app currently depends on the latest AWB, which may break due to the ongoing migration to ipywidgets 8. Version 2.5.0 includes the required feature and provides a stable dependency.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ install_requires =
     aiida-core~=2.5,<3
     Jinja2~=3.0
     aiida-quantumespresso~=4.12.1
-    aiidalab-widgets-base[optimade] @ git+https://github.com/aiidalab/aiidalab-widgets-base@master
+    aiidalab-widgets-base[optimade]~=2.5.0
     aiida-pseudo~=1.4
     filelock~=3.8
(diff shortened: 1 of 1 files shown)


edan-bainglass (approved):
Okay 👍 Can you open an issue as a reminder to update things w.r.t the AWB migration once stable? 🙏

edan-bainglass:
> Do you have any other changes you want to include in the new release? Otherwise, let's make a new, quick release.

The duplication feature, but Miki is still reviewing it. Let's do a quick release and add that one later.

PR #1456, opened 2026-02-02
Move non-app modules/packages out of app package

edan-bainglass:
This PR resolves circular dependencies caused by placing common things inside the app package, e.g., parameters, entry point utilities, etc.


mikibonacci (approved):
LGTM!

PR #1460, opened 2026-02-11
Add QE app article citation

edan-bainglass:
This PR adds a citation to the QE app article in the following locations:
- AiiDAlab Home app container
- QE app About section
- QE app documentation

---

image

---

image

---

image


giovannipizzi (approved):
Thanks! Approving for simplicity, but in the last screenshot, I would suggest to swap: first mention the paper, then say "In addition, please follow these links to".

The rest is fine!

PR #1466, opened 2026-03-02
Test python 3.12 and aiida-core=2.8.0

danielhollas:
- test with Python 3.12 and aiida-core 2.8
- Minor CI updates


agoscinski:
Okay it is only the populate_defaults issue, ran with a tmpfix and it runs through https://github.com/agoscinski/aiidalab-qe/actions/runs/22613886527/job/65524560911 Need to fix this in aiida-core

edan-bainglass (approved):
LGTM! Thanks to all for testing 🙏

PR #1467, opened 2026-03-11
Fix case

edan-bainglass:
In the home app container, "Calculation History" should be "Calculation history". Fixed here.


mikibonacci (approved):
LGTM!

PR #1469, opened 2026-03-25
Bump full-stack image in Dockerfile

danielhollas:
The most notable update in the full-stack image is a bump of aiida-core from 2.7.1 to 2.7.3 which includes a lot of bugfixes.


edan-bainglass (approved):
LGTM!

PR #1470, opened 2026-03-25
Test python 3.12 and aiida-core=2.8.0 (#1466)

danielhollas:
Same thing as in #1466 which I merged into a wrong branch, whoops.


edan-bainglass (approved):
👍👍

PR #1477, opened 2026-04-21
Update about page

edan-bainglass:
This PR aims to update the language of the About page. In the process it also removes the citation of the QE paper from the home app container (looked bad - should keep container minimal), as well as add a top-level button to the plugin store.

image


mikibonacci (approved):
LGTM! The button for plugins is very useful there (easily discoverable)

PR #1425, opened 2025-11-24
Avoid using `ipw.Output` widget

yakutovicha:
Partially fixes https://github.com/aiidalab/aiidalab-docker-stack/issues/512

The widget behaves quite unpredictably. For versions of ipywidgets, it fails to display the content and prevents other widgets from being loaded.

PR #1481, opened 2026-04-22
Fix bug report handler

edan-bainglass:
#1425 changed the widget in which we inject the bug handler from an `Output` to a `VBox`. However, AWB expects it to be an `Output` widget, so this broke the feature. More details for why this was done in the PR, but for now, we must revert. At least until we update AWB.


mikibonacci (approved):
LGTM!

PR #1488, opened 2026-04-25
Enable MPI in Wannier90 compilation

edan-bainglass:
Compile Wannier90 with MPI enabled

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -190,11 +190,13 @@ USER root
 # Build wannier90 for all arches, and build bader from source ONLY on arm64
 RUN set -ex; \
     apt-get update && apt-get install -y --no-install-recommends \
-    gfortran libblas-dev liblapack-dev liblapack3 openmpi-bin; \
+    gfortran libblas-dev liblapack-dev liblapack3 openmpi-bin libopenmpi-dev; \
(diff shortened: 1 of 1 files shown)


danielhollas (approved):
Two notes for the fiture:
- if wannier package is providing a core functionality, it would be good to have it on conda-forge, rather than compiling it here.
- seems like this compilation should be happening in one of the previous build stages, and then the binary should be copied to the final image.

edan-bainglass:
Thanks @danielhollas. See replies below.

> * if wannier package is providing a core functionality, it would be good to have it on conda-forge, rather than compiling it here.

We can have a look at this, sure. I'm not too familiar with the package. It seems @superstar54 is compiling it directly from a repo? Anyhow, will discuss with @giovannipizzi

> * seems like this compilation should be happening in one of the previous build stages, and then the binary should be copied to the final image.

Yeah, I had the same feeling. As discussed, there are various improvements one can take with the image. Will have a look in a future PR.

giovannipizzi:
@mbercx worked on a conda version for Wannier90, he can provide more context

mbercx:
Feedstock is at: https://github.com/conda-forge/wannier90-feedstock

IIRC, currently only serial version. Was planning to update together with @giovannipizzi when the new Wannier90 release is out.

PR #1493, opened 2026-05-05
Implement parameters-aware code selector toggling

### LINKED ISSUE #1371: Active plugin's code based on the setting parameters
Currently, all plugin codes are activated by default. However, in some cases, e.g., if fat bands are selected, then projwfc is not needed.

edan-bainglass:
This PR implements a parameters-aware code selector toggling system.

When registering conditional codes for a plugin, one can define a conditional

```python
def check_condition(parameters: dict) -> bool:
    needs_this_code = ...  # some check on parameters
    return needs_this_code
```

then assign it to the code resource at definition time

```python
CodeModel(
    name='...',
    description='...',
    default_calc_job_plugin='...',
    condition=check_condition,
)
```

Internally, the condition is evaluated when input parameters change, (de)activating the code, which then triggers its toggle (`display.layout` mutation).

Closes #1371

### DIFF
--- a/src/aiidalab_qe/app/submission/global_settings/model.py
+++ b/src/aiidalab_qe/app/submission/global_settings/model.py
@@ -79,13 +79,12 @@ class GlobalResourceSettingsModel(
         code_model: CodeModel,
     ) -> CodeModel | None:
         """Registers a code with this model."""
-        base_code_model = None
         default_calc_job_plugin = code_model.default_calc_job_plugin
         name = default_calc_job_plugin.split(".")[-1]
         # "." in the model key means nested models
         model_key = default_calc_job_plugin.replace(".", "__")
 
-        if not self.has_model(default_calc_job_plugin):
+        if not self.has_model(model_key):
             if default_calc_job_plugin == "quantumespresso.pw":
                 base_code_model = PwCodeModel(
                     name=name,
@@ -107,7 +106,7 @@ class GlobalResourceSettingsModel(
         else:
             self.plugin_mapping[identifier].append(model_key)
 
-        return base_code_model
+        return self.get_model(model_key)
 
     def check_resources(self):
         if not self.has_model("quantumespresso__pw"):
--- a/src/aiidalab_qe/app/submission/global_settings/model.py
+++ b/src/aiidalab_qe/app/submission/global_settings/model.py
@@ -19,13 +19,6 @@ class GlobalResourceSettingsModel(
     title = "Global resources"
     identifier = "global"
 
-    dependencies = [
-        "structure_uuid",
-        "input_parameters",
-    ]
-
-    input_parameters = tl.Dict()
-
     plugin_overrides = tl.List(tl.Unicode())
     plugin_overrides_notification = tl.Unicode("")
 
@@ -33,6 +26,7 @@ class GlobalResourceSettingsModel(
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.dependencies.append("structure_uuid")
 
         self._RUN_ON_LOCALHOST_NUM_SITES_WARN_THRESHOLD = 10
         self._RUN_ON_LOCALHOST_VOLUME_WARN_THRESHOLD = 1000  # \AA^3
@@ -59,7 +53,9 @@ class GlobalResourceSettingsModel(
         for identifier, code_names in self.plugin_mapping.items():
             if identifier in properties:
                 for code_name in code_names:
-                    self.get_model(code_name).activate()
+                    code_model = self.get_model(code_name)
+                    if code_model.check_condition(self.input_parameters):
+                        code_model.activate()
 
     def update_plugin_overrides_notification(self):
         if self.plugin_overrides:
@@ -102,6 +98,7 @@ class GlobalResourceSettingsModel(
                     name=name,
                     description=name,
                     default_calc_job_plugin=default_calc_job_plugin,
+                    condition=code_model.check_condition,
                 )
             self.add_model(model_key, base_code_model)
 
@@ -238,7 +235,7 @@ class GlobalResourceSettingsModel(
             if identifier in properties:
                 for name in code_names:
                     code_model = self.get_model(name)
-                    if not code_model.is_ready:
+                    if code_model.is_active and code_model.selected is None:
                         blocker = True
                         yield message.format(
                             property=identifier,
--- a/src/aiidalab_qe/app/submission/global_settings/model.py
+++ b/src/aiidalab_qe/app/submission/global_settings/model.py
@@ -27,6 +27,7 @@ class GlobalResourceSettingsModel(
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.dependencies.append("structure_uuid")
+        self.conditions = {}
 
         self._RUN_ON_LOCALHOST_NUM_SITES_WARN_THRESHOLD = 10
         self._RUN_ON_LOCALHOST_VOLUME_WARN_THRESHOLD = 1000  # \AA^3
@@ -50,11 +51,14 @@ class GlobalResourceSettingsModel(
             if identifier != "quantumespresso__pw":
                 code_model.deactivate()
         properties = self._get_properties()
-        for identifier, code_names in self.plugin_mapping.items():
+        for identifier, global_model_keys in self.plugin_mapping.items():
             if identifier in properties:
-                for code_name in code_names:
-                    code_model = self.get_model(code_name)
-                    if code_model.check_condition(self.input_parameters):
+                for model_key in global_model_keys:
+                    code_model = self.get_model(model_key)
+                    if any(
+                        condition(self.input_parameters)
+                        for condition in self._get_active_conditions(model_key)
+                    ):
                         code_model.activate()
 
     def update_plugin_overrides_notification(self):
@@ -83,6 +87,10 @@ class GlobalResourceSettingsModel(
         name = default_calc_job_plugin.split(".")[-1]
         # "." in the model key means nested models
         model_key = default_calc_job_plugin.replace(".", "__")
+
+        if model_key not in self.conditions:
+            self.conditions[model_key] = {}
+        self.conditions[model_key][identifier] = code_model.check_condition
 
         if not self.has_model(model_key):
             if default_calc_job_plugin == "quantumespresso.pw":
@@ -203,6 +211,15 @@ class GlobalResourceSettingsModel(
 
     def _get_properties(self) -> list[str]:
         return self.input_parameters.get("workchain", {}).get("properties", [])
+
+    def _get_active_conditions(self, model_key: str) -> list[callable]:
+        active_conditions = []
+        properties = self._get_properties()
+        if model_key in self.conditions:
+            for identifier, condition in self.conditions[model_key].items():
+                if identifier in properties:
+                    active_conditions.append(condition)
+        return active_conditions
 
     def _check_blockers(self):
         if not self.input_parameters:
--- a/src/aiidalab_qe/app/submission/global_settings/setting.py
+++ b/src/aiidalab_qe/app/submission/global_settings/setting.py
@@ -31,11 +31,8 @@ class GlobalResourceSettingsPanel(ResourceSettingsPanel[GlobalResourceSettingsMo
             "plugin_overrides",
         )
 
-    def render(self):
-        if self.rendered:
-            return
-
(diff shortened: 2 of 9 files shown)


src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -19,20 +19,14 @@ class GlobalResourceSettingsModel(
...
-    input_parameters = tl.Dict()
-
     plugin_overrides = tl.List(tl.Unicode())
     plugin_overrides_notification = tl.Unicode("")
 
     include = True
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.dependencies.append("structure_uuid")
```
edan-bainglass, line 29: I append any specific dependencies, as now the parent `ResourceSettingsModel` defines the `input_parameters` dependency.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -59,7 +53,9 @@ def update_active_codes(self):
         for identifier, code_names in self.plugin_mapping.items():
             if identifier in properties:
                 for code_name in code_names:
-                    self.get_model(code_name).activate()
+                    code_model = self.get_model(code_name)
+                    if code_model.check_condition(self.input_parameters):
+                        code_model.activate()
```
edan-bainglass: The new stuff - only activate if condition passes.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -102,6 +98,7 @@ def add_global_model(
                     name=name,
                     description=name,
                     default_calc_job_plugin=default_calc_job_plugin,
+                    condition=code_model.check_condition,
```
edan-bainglass, line 108: When creating global codes, we carry over the conditional function.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -238,7 +235,7 @@ def _check_blockers(self):
             if identifier in properties:
                 for name in code_names:
                     code_model = self.get_model(name)
-                    if not code_model.is_ready:
+                    if code_model.is_active and code_model.selected is None:
```
edan-bainglass, line 254: `is_ready` = active and selected. But here we really want to check if it ISN'T selected ONLY if it is active.

src/aiidalab_qe/app/submission/global_settings/setting.py
```diff
@@ -31,11 +31,8 @@ def __init__(self, model: GlobalResourceSettingsModel, **kwargs):
             "plugin_overrides",
         )
 
-    def render(self):
-        if self.rendered:
-            return
-
-        self.code_widgets_container = ipw.VBox()
+    def _render(self):
```
edan-bainglass, line 34: The parent `ResourceSettingsPanel` now provides the general pattern, wrapping around `_render`, which is left to subclasses to implement. This is the same pattern we take in the results step.

src/aiidalab_qe/app/submission/global_settings/setting.py
```diff
@@ -53,39 +50,11 @@ def render(self):
             self.plugin_overrides_notification,
         ]
 
-        self.rendered = True
-
-        # Render any active codes
-        for _, code_model in self._model.get_models():
-            if code_model.is_active:
-                self._toggle_code(code_model)
```
edan-bainglass, line 61: This is moved to `_post_render`

src/aiidalab_qe/app/submission/global_settings/setting.py
```diff
@@ -53,39 +50,11 @@ def render(self):
...
-                            [
-                                "num_cpus",
-                                "num_nodes",
-                                "ntasks_per_node",
-                                "cpus_per_task",
-                                "max_wallclock_seconds",
-                            ],
-                        )
+                global_code_model = self._model.add_global_model(identifier, code_model)
+                self._register_code_trait_callbacks(global_code_model)
```
edan-bainglass, line 57: All the special handling for global codes is moved to `_register_code_trait_callbacks`.

src/aiidalab_qe/app/submission/global_settings/setting.py
```diff
@@ -101,18 +70,27 @@ def _on_input_structure_change(self, _):
     def _on_plugin_overrides_change(self, _):
         self._model.update_plugin_overrides_notification()
 
-    def _on_code_activation_change(self, change):
-        self._toggle_code(change["owner"])
-
-    def _on_code_selection_change(self, _):
-        self._model.update_blockers()
```
edan-bainglass, line 108: Moved to `ResourceSettingsPanel`

src/aiidalab_qe/app/submission/global_settings/setting.py
```diff
@@ -101,18 +70,27 @@ def _on_input_structure_change(self, _):
...
+            code_model.observe(
+                self._on_pw_code_resource_change,
+                [
+                    "num_cpus",
+                    "num_nodes",
+                    "ntasks_per_node",
+                    "cpus_per_task",
+                    "max_wallclock_seconds",
+                ],
+            )
```
edan-bainglass, line 92: Global-settings-only logic

src/aiidalab_qe/common/code/model.py
```diff
@@ -36,12 +40,14 @@ def __init__(
...
         code_widget_class=QEAppComputationalResourcesWidget,
+        condition: t.Callable[[dict], bool] | None = None,
     ):
         self.name = name
         self.description = description
         self.default_calc_job_plugin = default_calc_job_plugin
         self.code_widget_class = code_widget_class
-        self.is_rendered = False
+        self.is_widget_rendered = False
+        self._condition = condition or (lambda _: True)
```
edan-bainglass: Note the backwards-compatible default

src/aiidalab_qe/common/code/model.py
```diff
@@ -36,12 +40,14 @@ def __init__(
...
         default_calc_job_plugin,
         code_widget_class=QEAppComputationalResourcesWidget,
+        condition: t.Callable[[dict], bool] | None = None,
     ):
         self.name = name
         self.description = description
         self.default_calc_job_plugin = default_calc_job_plugin
         self.code_widget_class = code_widget_class
-        self.is_rendered = False
+        self.is_widget_rendered = False
```
edan-bainglass, line 49: For clarity - it is the widget that is rendered of course, not the model.

src/aiidalab_qe/common/panel.py
```diff
@@ -213,7 +223,47 @@ def __init__(self, model, **kwargs):
...
+        self._model.update_blockers()
+
+    def _on_code_activation_change(self, change):
+        self._toggle_code(change["owner"])
+
+    def _on_code_selection_change(self, _):
+        self._model.update_blockers()
+
+    def _on_code_resource_change(self, _):
+        pass
```
edan-bainglass, line 260: Though at the moment, only the global panel uses this, I decided to keep it general, in case down the road, plugin panels choose to take action on resource changes.

src/aiidalab_qe/common/panel.py
```diff
@@ -256,12 +307,16 @@ def _toggle_code(self, code_model: CodeModel):
...
+        if not code_model.is_widget_rendered:
             code_widget.observe(
                 code_widget.update_resources,
                 "value",
             )
             self._render_code_widget(code_model, code_widget)
+        if code_model.is_active:
+            code_widget.layout.display = "flex"
+        else:
+            code_widget.layout.display = "none"
```
edan-bainglass, line 319: Hide/show

src/aiidalab_qe/common/panel.py
edan-bainglass, line 1: Several things already commented above are implemented here w.r.t generalization of resource panels.

mikibonacci (approved):
Nice feature! LGTM!

edan-bainglass:
@mikibonacci this is not quite right. I did not consider cases where the same code is used by multiple plugins. For example, the core bands plugin would define a condition for projwfc on the fat bands parameter. But then already the PDOS plugin defines an unconditional projwfc. So if both are selected, the check needs to be done over all (or rather `any`). Moreover, even if just bands is selected, the global panel still shows projwfc. This is because there was a bug in adding global codes. But even with this fixed, the logic is still not quite right. Need to rethink before merging this.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -19,20 +19,15 @@ class GlobalResourceSettingsModel(
...
-
     plugin_overrides = tl.List(tl.Unicode())
     plugin_overrides_notification = tl.Unicode("")
 
     include = True
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.dependencies.append("structure_uuid")
+        self.conditions = {}
```
edan-bainglass, line 30: To store parameter-aware code conditions across codes.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -56,10 +51,15 @@ def update_active_codes(self):
...
             if identifier in properties:
-                for code_name in code_names:
-                    self.get_model(code_name).activate()
+                for model_key in global_model_keys:
+                    code_model = self.get_model(model_key)
+                    if any(
+                        condition(self.input_parameters)
+                        for condition in self._get_active_conditions(model_key)
+                    ):
+                        code_model.activate()
```
edan-bainglass, line 62: This checks if any of the active plugins requires this code.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -83,13 +83,16 @@ def add_global_model(
...
-        base_code_model = None
         default_calc_job_plugin = code_model.default_calc_job_plugin
         name = default_calc_job_plugin.split(".")[-1]
         # "." in the model key means nested models
         model_key = default_calc_job_plugin.replace(".", "__")
 
-        if not self.has_model(default_calc_job_plugin):
+        if model_key not in self.conditions:
+            self.conditions[model_key] = {}
+        self.conditions[model_key][identifier] = code_model.check_condition
```
edan-bainglass, line 93: Add any parameter-aware code conditionals.

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -83,13 +83,16 @@ def add_global_model(
...
+
+        if not self.has_model(model_key):
```
edan-bainglass, line 95: Driveby fix - we were checking the wrong variable!

src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -110,7 +114,7 @@ def add_global_model(
         else:
             self.plugin_mapping[identifier].append(model_key)
 
-        return base_code_model
+        return self.get_model(model_key)
```
edan-bainglass, line 117: Avoid relying on `base_code_model`, which is only temporarily used in adding global models.

src/aiidalab_qe/plugins/bands/bands_workchain.py
```diff
@@ -291,6 +291,10 @@ def get_builder_from_protocol(
             builder.bands = builder_bands
 
         elif simulation_mode == "fat_bands":
+            if projwfc_code is None:
+                raise ValueError(
+                    "projwfc_code must be provided for fat bands simulation mode"
+                )
```
edan-bainglass, line 297: This PR allows `projwfc_code` to be `None` if we're not computing fat bands!

edan-bainglass:
> @mikibonacci this is not quite right. I did not consider cases where the same code is used by multiple plugins. For example, the core bands plugin would define a condition for projwfc on the fat bands parameter. But then already the PDOS plugin defines an unconditional projwfc. So if both are selected, the check needs to be done over all (or rather `any`). Moreover, even if just bands is selected, the global panel still shows projwfc. This is because there was a bug in adding global codes. But even with this fixed, the logic is still not quite right. Need to rethink before merging this.

Okay @mikibonacci. This is now resolved. Please have a look 🙏

edan-bainglass:
> Okay @mikibonacci. This is now resolved. Please have a look 🙏

Merging (need to move forward). But please comment if you have any pressing issues with the latest changes. We can patch later if any such issues arise. For now, this has been tested quite a bit 🙂

mikibonacci:
Great! I'll test it as soon as I implement the feature in the plugins! I have no comments for now, thanks!

PR #1496, opened 2026-05-07
Plugin compatibility

edan-bainglass:
This PR updates two things:
1. `plugins.yaml`: all plugins have moved to the AiiDAlab/aiidaplugins organizations
2. Some QE calculationis are developed externally (e.g., Xspec, Hubbard). In these cases, the entry point is not `quantumespresso`, but e.g., `xspec`. However, the QE app's auto-code-installation scheme assumes the `quantumespresso` prefix. This PR adds a simple mapping patch (thanks @mbercx 🙏)

### Update

This PR now does **more than two things** 😅

In the process of updating other plugins, I noticed that we have a circular dependency issue. The newer process tree loads plugins for their process metadata. Since it lives in the common package, when plugins reach out for common things, we get a circular dependency, as the process tree package loads and fetches plugins metadata at the same time.

The solution here is simple - the tree is not common! It is strictly used by the app. Here I move it into the app package.
I also removed the old workchain selector widgets, which we haven't been using in a year or so.

I also pin `hiphive` to `1.3.1` to avoid conflicts with `spglib`, which is pinned in at least the muonic plugin.


src/aiidalab_qe/setup/codes.py
```diff
@@ -43,6 +43,12 @@ def get_qe_env():
     "hp",
 )
 
+# Some quantumespresso features are developed in separate repositories, with different
+# entry point prefix (e.g. `xspec` in aiida-qe-xspec)
+ENTRY_POINT_MAP = {
+    "xspectra": "xspec",
+}
```
edan-bainglass, line 50: @PNOGillespie this resolves the xspectra code auto-installation issue.

plugins.yaml
```diff
@@ -3,33 +3,34 @@ aiidalab-qe-vibroscopy:
...
     category: calculation
-    requires_aiidalab_qe: '>24.04.0'
+    requires_aiidalab_qe: '>26.05.0'
 
 aiidalab-qe-muon:
     title: Muon spectroscopy (aiidalab-qe-muon)
     description: Compute properties to assist µSR experiments, such as muon stopping sites and related properties (local fields and polarization signals).
     author: Miki Bonacci, Ifeanyi J. Onuorah, Pietro Bonfa', Giovanni Pizzi and Roberto de Renzi
-    github: https://github.com/mikibonacci/aiidalab-qe-muon@v1.0.0
+    github: https://github.com/aiidalab/aiidalab-qe-muon@v1.1.1
```
mikibonacci: I just published the v1.1.2

mikibonacci (changes requested):
LGTM, except the version of the aiidalab-qe-muon plugin which now should be v1.1.2. When this is fixed, for me it can go in.

mikibonacci (approved):

PR #1500, opened 2026-05-20
Fix absolute convergence thresholds

### LINKED ISSUE #1498: Incorrect display of energy convergence thresholds after job submission
**Describe the bug**
Several theshold values (SCF, total energy, force) under the Convergence/Accuracy tab are displayed (calculated) wrong after job submission. The previousely entered values remain the same, but the calculated energies in untis of Ry (e.g., based on input threshold [Ry/atom] times num_atoms) are displayed wrong (i.e., what you see is not congruent with the real values).

**To Reproduce**
Open a finished job.
Step 2 → Step 2.2: Customize calculation parameters → Advanced settings → Convergence/Accuracy

**Expected behavior**
Please fix bug.

**Screenshots**
Image

**Version (if known)**
v26.04.0

**Additional context**

edan-bainglass:
The absolute convergence thresholds weren't properly linked to their corresponding widget. Also, there is no need for these to be model traits, as their are purely a UI feature in sync with the actual threshold traits.

Closes #1498

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/model.py
@@ -41,7 +41,7 @@ class AdvancedConfigurationSettingsModel(
     include = True
 
     def get_model_state(self) -> dict:
-        num_atoms = len(self.input_structure.sites) if self.has_structure else 1
+        num_atoms = self.get_num_atoms() or 1
 
         general = t.cast(
             GeneralConfigurationSettingsModel,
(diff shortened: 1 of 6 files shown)


AndresOrtegaGuerrero:
Isnt any test that checks this as well ?

edan-bainglass:
> Isnt any test that checks this as well ?

There is now 🙂

src/aiidalab_qe/app/configuration/advanced/model.py
```diff
@@ -41,7 +41,7 @@ class AdvancedConfigurationSettingsModel(
     include = True
 
     def get_model_state(self) -> dict:
-        num_atoms = len(self.input_structure.sites) if self.has_structure else 1
+        num_atoms = self.get_num_atoms() or 1
```
AndresOrtegaGuerrero, line 44: Why `or 1`, if there is no structure should it be None or 0? as in `get_num_atoms()`

edan-bainglass: Because it was historically 1. See the changed line. It also seemed odd to me, but I don't have time at the moment to sift through the history 🙂

AndresOrtegaGuerrero (approved):
LGTM!

PR #1502, opened 2026-05-20
Minor Dockerfile cleanup

danielhollas:
This is just a minor cleanup, organizing all common ENV into the base image so that they don't have to be defined multiple times in various build stages


edan-bainglass (approved):
Thanks @danielhollas. LGTM!

Can you comment on the removal of `--cache-dir=${UV_CACHE_DIR}`?

danielhollas:
> Can you comment on the removal of --cache-dir=${UV_CACHE_DIR}?

I've replaced it with the global `UV_CACHE_DIR` enviroment variable: https://docs.astral.sh/uv/reference/environment/#uv_cache_dir

danielhollas:
I pushed one more change, using a `mount=type=cache` for uv cache, according to uv's own guide for Docker: https://docs.astral.sh/uv/guides/integration/docker/#caching

danielhollas:
Failing pytest job looks like a test flake and does not depend on Dockerfile so going ahead and merge

PR #1504, opened 2026-05-20
Fix bugs in pseudopotentials panel

### LINKED ISSUE #1495: Functional is not set correctly on process load
Likely it is overwritten (reset) when the protocol is set. Rethink!

edan-bainglass:
The panel was incorrectly set when loaded from a process (or duplicated).

Closes #1495

Note that this doesn't work on its own due to a larger issue with the blockers system. This is fixed in another PR.

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/model.py
@@ -210,6 +210,8 @@ class AdvancedConfigurationSettingsModel(
         )
         with pseudos.hold_trait_notifications():
             if pseudo_family_string := state.get("pseudo_family"):
+                # If a pseudo family string exists, we can assume that no custom pseudos were uploaded.
+                # We therefore populate the UI from the pseudo family.
                 pseudo_family = PseudoFamily.from_string(pseudo_family_string)
                 library = f"{pseudo_family.library} {pseudo_family.accuracy}"
                 if relativistic := pseudo_family.relativistic:
@@ -217,23 +219,32 @@ class AdvancedConfigurationSettingsModel(
                 pseudos.functional = pseudo_family.functional
                 pseudos.library = library
                 pseudos.family = pseudo_family_string
+                pseudos.functionals = [pseudos.functional] * num_kinds
             else:
+                # Otherwise, we assume a custom pseudo upload (warning enabled).
+                # Everything is set to None except the functionals, which we extract from the pseudos.
+                pseudos.show_upload_warning = True
+                functionals = []
                 try:
-                    pp_uuid = next(iter(PW["pseudos"].values()))
-                    pseudo_info = get_pseudo_info(pp_uuid)
-                    pseudos.functional = pseudo_info["functional"]
-                except Exception:
+                    for pp_uuid in PW["pseudos"].values():
+                        pseudo_info = get_pseudo_info(pp_uuid)
+                        functionals.append(pseudo_info["functional"])
+                    if len(set(functionals)) == 1:
+                        pseudos.functional = functionals[0]
+                    else:
+                        pseudos.functional = None
+                except Exception as err:
+                    print(f"Error loading pseudos: {err}")
                     pseudos.functional = None
                 pseudos.library = None
                 pseudos.family = None
-                pseudos.show_upload_warning = True
+                pseudos.functionals = functionals
 
-            pseudos.functionals = [pseudos.functional] * num_kinds
-
-            if pseudos_dictionary := PW.get("pseudos"):
-                pseudos.dictionary = pseudos_dictionary
-                pseudos.ecutwfc = SYSTEM.get("ecutwfc", 0.0)
-                pseudos.ecutrho = SYSTEM.get("ecutrho", 0.0)
+        if pseudos_dictionary := PW.get("pseudos"):
+            # We set these outside of the context manager to avoid a reset by other trait notifications.
+            pseudos.dictionary = pseudos_dictionary
+            pseudos.ecutwfc = SYSTEM.get("ecutwfc", 0.0)
+            pseudos.ecutrho = SYSTEM.get("ecutrho", 0.0)
 
         magnetization = t.cast(
             MagnetizationConfigurationSettingsModel,
--- a/src/aiidalab_qe/app/configuration/advanced/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/model.py
@@ -202,8 +202,6 @@ class AdvancedConfigurationSettingsModel(
             if "smearing" in SYSTEM:
                 smearing.type = SYSTEM["smearing"]
 
-        num_kinds = len(self.input_structure.kinds) if self.has_structure else 1
-
         pseudos = t.cast(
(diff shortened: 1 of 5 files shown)


edan-bainglass:
@AndresOrtegaGuerrero I did not add any tests for this. I'll open an issue to add tests, but I don't have time at the moment.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -149,17 +148,8 @@ def update_family_parameters(self):
...
             self.library = self._defaults["library"]
 
-    def update_functionals(self):
-        if self.locked or not (self.functional and self.family):
-            return
-        self.functionals = (
-            [self.functional] * len(self.input_structure.kinds)
-            if self.has_structure
-            else []
-        )
```
edan-bainglass, line 159: There is no need for `functionals`. We have all the information we need for what it was used for. Discarding it to reduce complexity.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -338,6 +328,27 @@ def update_library_options(self):
 
         self.update_family_parameters()
 
+    def get_cutoffs_by_index(self, index: int) -> list[float]:
+        return (
+            [self.cutoffs[0][index], self.cutoffs[1][index]]
+            if len(self.cutoffs[0]) > index
+            else [0.0, 0.0]
+        )
```
edan-bainglass, line 336: Cleaner to have this helper here than in the settings panel.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -338,6 +328,27 @@ def update_library_options(self):
...
+            try:
+                assert uuid is not None
+                pseudo = orm.load_node(uuid)
+                pseudos.append(pseudo)
+            except (AssertionError, exceptions.NotExistent):
+                continue
+        functional_set = {pp.base.extras.get("functional", None) for pp in pseudos}
+        functional = functional_set.pop() if len(functional_set) == 1 else None
+        self._defaults["functional"] = functional
+        self.functional = self._get_default("functional")
```
edan-bainglass, line 359: This is called on change of pseudos in the dictionary. It collects each pseudo's functional (from extras), and sets the `functional` trait to the common functional if exists, otherwise `None` - functional is disabled in the UI.

src/aiidalab_qe/app/configuration/advanced/pseudos/pseudos.py
```diff
@@ -288,65 +307,23 @@ def on_default_pseudo(
...
+            uploader = PseudoPotentialUploader(model=uploader_model)
             uploader.render()
 
-            on_default_pseudo()
-
             self._links.extend(uploader.links)
 
             children.append(uploader)
 
+            synchronize_pseudo()
```
edan-bainglass, line 327: This is mostly renaming and reshuffling for clarity, plus adjusting the logic w.r.t dropping `functionals`.

src/aiidalab_qe/app/configuration/advanced/model.py
```diff
@@ -218,22 +218,29 @@ def set_model_state(self, state: dict):
...
 
-            if pseudos_dictionary := PW.get("pseudos"):
-                pseudos.dictionary = pseudos_dictionary
-                pseudos.ecutwfc = SYSTEM.get("ecutwfc", 0.0)
-                pseudos.ecutrho = SYSTEM.get("ecutrho", 0.0)
+        if pseudos_dictionary := PW.get("pseudos"):
+            # We set these outside of the context manager to avoid a reset by other trait notifications.
+            pseudos.dictionary = pseudos_dictionary
+            pseudos.ecutwfc = SYSTEM.get("ecutwfc", 0.0)
+            pseudos.ecutrho = SYSTEM.get("ecutrho", 0.0)
```
edan-bainglass, line 243: This part I suspect can be simplified, but immediate efforts failed. Keeping as is for now.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -413,3 +424,9 @@ def _check_blockers(self):
         if self.spin_orbit == "soc":
             if relativistic_set != {"full"}:
                 yield "For spin-orbit coupling (SOC) calculations, all pseudopotentials must be fully relativistic."
+
+        if self.ecutwfc == 0.0:
```
AndresOrtegaGuerrero, line 437: Should this condition block step 2 ?

edan-bainglass: It should and it does. Blockers in any configuration settings panel are linked to the blockers of the configuration step and are displayed there (and block the step). Currently only implemented in the pseudos panel, but in principal it is available to all. Even more so after #1513.

edan-bainglass:
@AndresOrtegaGuerrero did you have any other comments? I'd like to get this in before releasing a new QE app version 🙏 If there are any issues that I missed, we can address them with a patch release next month (including further tests of course).

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -338,6 +328,27 @@ def update_library_options(self):
...
+            [self.cutoffs[0][index], self.cutoffs[1][index]]
+            if len(self.cutoffs[0]) > index
+            else [0.0, 0.0]
+        )
+
+    def update_functional(self):
+        pseudos: list[UpfData] = []
+        for uuid in self.dictionary.values():
+            try:
+                assert uuid is not None
```
AndresOrtegaGuerrero, line 343: why not `if uuid is None: `?  , I guess is to put raise an Error ?

AndresOrtegaGuerrero (approved):
LGMT! I left a comment but it was more like a question

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
edan-bainglass: `uuid` is rarely `None`. In practice, it shouldn't ever be `None`. Hence, instead of checking for a condition that is rarely there (if ever), I rather try what SHOULD work, catching the rare cases where something is out of sync. Ideally, I would implement a better logging system for debugging and add debug messages in these rare events, to let us know what happened. Actually, though it would be annoying for the user, perhaps it is good to still print out something. I'll update this block, separating the two errors with dedicated print messages.

edan-bainglass: Looking at some older code (see the `_check_blockers` method in the same file, I indeed split the two error paths, with the `assert` error leading to a blocker w.r.t non-existent pseudos in the selected sssp library for the selected element. I recall this being an issue for Ce at some point. I should do something similar for the new code.

AndresOrtegaGuerrero: Yes, there are missing Psedos for SOC for lanthanides... yes

PR #1507, opened 2026-05-20
Install Bader from conda on arm64

danielhollas:
I added the arm64 builds (both linux and macos) conda-forge/bader-feedstock: https://github.com/conda-forge/bader-feedstock/pull/4


edan-bainglass (approved):
LGTM! Keep 'em coming 👍

PR #1508, opened 2026-05-20
Move wannier compilation earlier in Dockerfile

danielhollas:
Wannier compilation is the slow step in the docker build, so it shouldn't be at the end. Move it earlier so that it can run in parallel with other build stages, which should speed up the build and also should make cache hits much more likely.

Also moved all the python packages installation with `uv pip` together in a single step, no reason for them to be separate!


edan-bainglass (approved):
Clean 👍

PR #1509, opened 2026-05-20
Dockerfile: Make sure files not tracked by git don't end up in the image!

danielhollas:
This has probably been broken for a while. Doesn't matter for images built in CI, but sure does in local development!


edan-bainglass (approved):
LGTM!

PR #1506, opened 2026-05-20
WIP: Build QeApp image with Python 3.12 and nbclassic

danielhollas:
Test the future nbclassic-based python 3.12 full-stack image from aiidalab/aiidalab-docker-stack#479

Issue #1492, opened 2026-05-01
Support for mamba 2.x


danielhollas:
mamba v2.0 is a big release, in which mamba was completely decoupled from conda and its CLI is now based on micromamba (standalone C++ implementation). For more details see https://mamba.readthedocs.io/en/latest/developer_zone/changes-2.0.html

If we upgrade to this version in the docker stack, the most significant breaking change is the `--prefix/-p` option was renamed to `--root-prefix/-r` so we'll have to change the Dockerfiles accordingly as this is used in bunch of places. QeApp Dockerfile will need to be adapted as well. For now we should stick to 1.5.x versions (which I did in #455).

(issue duplicated from https://github.com/aiidalab/aiidalab-docker-stack/issues/504)

edan-bainglass:
> For now we should stick to 1.5.x versions

What is holding this up? I'm happy to PR this and do a patch release if it speeds things up 🙂

danielhollas:
The nbclassic image already has the 2.x version :-)

PR #1512, opened 2026-05-21
Add Python 3.12 / AiiDA 2.8 compatibility for the QE app

cpignedoli:
Hi @edan-bainglass and @danielhollas,

I started this branch as an early compatibility port of the QE app to the upcoming AiiDAlab stack based on Python 3.12, AiiDA core 2.8, ipywidgets 8, and Notebook 7 through nbclassic.

This PR is meant to make the work visible, avoid duplicated effort, and provide a concrete example showing that the QE app can run successfully in the new environment. **I am perfectly fine if we decide not to merge it directly and instead cherry-pick selected changes from it.**

Testing environment:

I am testing this branch inside a local development AiiDAlab Docker image:

* Docker image: `aiidalab/full-stack:dev`
* Built locally on top of the Docker stack work in `aiidalab/aiidalab-docker-stack#479`
* Python: `3.12.11`
* AiiDA core: `2.8.0`
* AiiDAlab: `26.5.2`
* Notebook: `7.x` base, running through `nbclassic`
* ipywidgets: `8.x`
* aiida-quantumespresso: `4.17.0`
* aiidalab-widgets-base: `3.0.0a0`

Summary of changes so far:

* updates dependency metadata and CI for Python 3.12 / AiiDA 2.8
* uses the `aiidalab/aiidalab-docker-stack#479` full-stack image for Docker builds
* fixes homepage image loading in the new environment by embedding the start-page images directly in the widget, instead of relying on static paths that seem fragile in this stack
* fixes ipywidgets 8 accordion/title handling
* fixes AiiDA 2.8 plugin execution logic
* patches the bands/projwfc validator signature mismatch
* fixes Docker setup for `aiida-bader`
* updates Docker integration tests for nbclassic and container readiness

Latest checks are green:

* `CI`
* `Build Docker image and test`
* `Build Docker image and upload`

Manual smoke test: Si bulk workflow completed successfully, including relax, bands, projwfc bands, and PDOS.

I will now try to prepare an equivalent PR for the Python 3.9 environment. As a next step, I also want to check and, if appropriate, enable the new base-restart options in the QE app.

image

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
 
         strategy:
             matrix:
-                python-version: ['3.9', '3.12']
+                python-version: ['3.12']
                 aiida-core-version: ['2.8']
             fail-fast: false
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 ###############################################################################
 # 1) Global ARGs
 ###############################################################################
-ARG FULL_STACK_VER=2026.1030
+ARG FULL_STACK_VER=pr-479
 ARG UV_VER=0.11.15
 ARG QE_VER=7.4
 ARG QE_DIR=/opt/conda/envs/quantum-espresso-${QE_VER}
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,7 +85,7 @@ COPY --chown=${NB_UID}:${NB_GID} setup.cfg pyproject.toml LICENSE README.md ${QE
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=type=cache,sharing=locked,target=${UV_CACHE_DIR},uid=${NB_UID},gid=${NB_GID} \
     uv pip install --strict --no-build-isolation euphonic==1.3.2 && \
-    uv pip install --strict . ${AIIDA_HQ_PKG} ${MUON_PKG} aiidalab-qe-vibroscopy aiida-bader
+    uv pip install --strict . ${AIIDA_HQ_PKG} ${MUON_PKG} aiidalab-qe-vibroscopy "aiida-bader>=0.1.10"
 
 ###############################################################################
 # 6) home_build stage
@@ -126,8 +126,8 @@ RUN --mount=from=qe_conda_env,source=${QE_DIR},target=${QE_DIR} \
     verdi code create core.code.installed --label python --computer=${COMPUTER_LABEL} --default-calc-job-plugin pythonjob.pythonjob --filepath-executable=/opt/conda/bin/python -n && \
     verdi code create core.code.installed --label bader --computer=${COMPUTER_LABEL} --default-calc-job-plugin bader.bader --filepath-executable=${QE_DIR}/bin/bader -n && \
     verdi code create core.code.installed --label wannier90 --computer=${COMPUTER_LABEL} --default-calc-job-plugin wannier90.wannier90 --filepath-executable=/opt/conda/bin/wannier90.x -n && \
-    # run post_install for plugin
-    python -m aiida_bader post-install && \
+    # Import Bader PAW pseudopotentials; the code is created explicitly above.
+    python -c 'from aiida import load_profile; load_profile(); from aiida_bader.utils import install_pseudos; install_pseudos()' && \
     python -m aiidalab_qe_vibroscopy setup-phonopy && \
     python -m aiidalab_qe_muon setup-python3 && \
     # wannier90 plugin need SSSP 1.1
@@ -198,7 +198,7 @@ RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=type=cache,sharing=locked,target=${UV_CACHE_DIR},uid=${NB_UID},gid=${NB_GID} \
     --mount=from=build_deps,source=${QE_APP_SRC},target=${QE_APP_SRC},rw \
     uv pip install --strict --compile-bytecode \
-      ${QE_APP_SRC} ${AIIDA_HQ_PKG} ${MUON_PKG} aiidalab-qe-vibroscopy aiida-bader
+      ${QE_APP_SRC} ${AIIDA_HQ_PKG} ${MUON_PKG} aiidalab-qe-vibroscopy "aiida-bader>=0.1.10"
 
 # copy hq binary
 COPY --from=home_build /opt/conda/hq /usr/local/bin/
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@
 ###############################################################################
 # 1) Global ARGs
 ###############################################################################
-ARG FULL_STACK_VER=pr-479
+ARG FULL_STACK_VER=edge
 ARG UV_VER=0.11.15
 ARG QE_VER=7.4
 ARG QE_DIR=/opt/conda/envs/quantum-espresso-${QE_VER}
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,7 +37,7 @@ install_requires =
     plotly>=5.24,<7
     beautifulsoup4~=4.11
 
-python_requires = >=3.9
+python_requires = >=3.12
 
 [options.packages.find]
 where = src
@@ -57,6 +57,7 @@ dev =
     pre-commit~=3.2
     pytest~=7.4
     pytest-regressions~=2.2
+    upf_tools~=0.1.9
     pgtest==1.3.1
     pytest-cov~=5.0
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,20 +24,17 @@ package_dir =
     = src
 packages = find:
 install_requires =
-    aiida-core~=2.5,<3
+    aiida-core>=2.8,<3
     Jinja2~=3.0
-    aiida-quantumespresso~=4.12.1
-    aiidalab-widgets-base[optimade]~=2.5.0
+    aiida-quantumespresso>=4.17,<5
+    aiidalab-widgets-base[optimade]>=3.0.0a0,<4
     aiida-pseudo~=1.4
     filelock~=3.8
-    importlib-resources~=5.2
+    importlib-resources>=5.2,<8
     aiida-wannier90-workflows==2.7.0
-    anywidget==0.9.13
+    anywidget>=0.9.13,<1
     table_widget~=0.0.2
-    shakenbreak~=3.3.1
-    plotly~=5.24
-    kaleido~=0.2.1
-    upf_tools~=0.1.9
+    plotly>=5.24,<7
     beautifulsoup4~=4.11
 
 python_requires = >=3.9
@@ -46,6 +43,15 @@ python_requires = >=3.9
 where = src
 
 [options.extras_require]
+export =
+    kaleido~=0.2.1
+
+pseudos =
+    upf_tools~=0.1.9
+
+defects =
+    shakenbreak~=3.3.1
+
 dev =
     bumpver~=2023.1124
     pre-commit~=3.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,9 +30,9 @@ install_requires =
     aiidalab-widgets-base[optimade]>=3.0.0a0,<4
     aiida-pseudo~=1.4
     filelock~=3.8
-    importlib-resources>=5.2,<8
+    importlib-resources~=5.2
     aiida-wannier90-workflows==2.7.0
-    anywidget>=0.9.13,<1
+    anywidget==0.9.13
     table_widget~=0.0.2
     plotly>=5.24,<7
     beautifulsoup4~=4.11
--- a/src/aiidalab_qe/app/configuration/advanced/pseudos/utils.py
+++ b/src/aiidalab_qe/app/configuration/advanced/pseudos/utils.py
@@ -1,8 +1,6 @@
 import typing as t
 
 from aiida_pseudo.groups.family import PseudoPotentialFamily
-from upf_tools import UPFDict
-
 from aiida import orm
 from aiida.plugins import DataFactory, GroupFactory
 
@@ -89,6 +87,14 @@ def get_pseudo_family_by_label(label) -> PseudoPotentialFamily:
 
 
 def get_upf_dict(pseudo):
+    try:
+        from upf_tools import UPFDict
+    except ModuleNotFoundError as err:
+        raise ModuleNotFoundError(
+            "Parsing UPF pseudo files requires the optional 'upf_tools' dependency. "
+            "Install the 'pseudos' extra to enable this feature."
+        ) from err
+
     with pseudo.as_path() as pseudo_path:
         upf_dict = UPFDict.from_upf(pseudo_path.as_posix())
     return upf_dict
--- a/src/aiidalab_qe/app/configuration/advanced/pseudos/utils.py
+++ b/src/aiidalab_qe/app/configuration/advanced/pseudos/utils.py
@@ -1,6 +1,7 @@
 import typing as t
 
 from aiida_pseudo.groups.family import PseudoPotentialFamily
+
 from aiida import orm
 from aiida.plugins import DataFactory, GroupFactory
 
--- a/src/aiidalab_qe/common/process/tree.py
+++ b/src/aiidalab_qe/common/process/tree.py
@@ -18,7 +18,7 @@ from aiidalab_widgets_base import LoadingWidget
 
 from .state import STATE_ICONS
 
-TITLE_MAPPING = {
+BASE_TITLE_MAPPING = {
     "QeAppWorkChain": "Quantum ESPRESSO app workflow",
     "PwRelaxWorkChain": "Structure relaxation workflow manager",
     "PwBaseWorkChain": {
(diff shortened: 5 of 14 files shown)


danielhollas:
@cpignedoli awesome, thanks for working on this! Yesterday I've also opened #1506 for testing the QeApp image build based on the new python3.12 based image. So maybe you could cherry-pick the first commit from #1506 and I can close it (I don't plan to work on it myself, was just curious)

cpignedoli:
Si Bulk went through
image

danielhollas:
@cpignedoli awesome!

btw: any fixes that are specific to aiida-core 2.8 should also go to a separate PR that can be merged faster, since our plan was to release aiida-core 2.8 image still with python 3.9.

cpignedoli:
Hi @danielhollas, sure, I will cherry-pick #1506
For Python 3.9 and aiida-core 2.8 OK, I will make a separate commit as soon as I am confident with this one.

src/aiidalab_qe/app/configuration/advanced/pseudos/utils.py
```diff
@@ -89,6 +88,14 @@ def get_pseudo_family_by_label(label) -> PseudoPotentialFamily:
 
 
 def get_upf_dict(pseudo):
+    try:
+        from upf_tools import UPFDict
+    except ModuleNotFoundError as err:
+        raise ModuleNotFoundError(
+            "Parsing UPF pseudo files requires the optional 'upf_tools' dependency. "
+            "Install the 'pseudos' extra to enable this feature."
+        ) from err
```
edan-bainglass, line 97: So the user experience would likely be:
1. Try to upload a pseudo
2. Pseudo not part of family, so we need to parse it
3. upf_tools not installed - error raised
4. ~User reads error, installs package, and tries again 💪~
5. User doesn't read error and reports bug 🤦🏻‍♂️

This is fine for now, but likely, we will need to implement a pre-check on the module, and if not installed, disable the upload feature and show the same message you show here in the error instead in a friendly warning. On package install, enable buttons and remove warning. If you agree, please open an issue to do this 🙏

src/aiidalab_qe/common/process/tree.py
edan-bainglass, line 1: I made a comment about this in [another PR](https://github.com/aiidalab/aiidalab-qe/pull/1496#issue-4399634069). This one here handles it quite nicely 👍 Note that the other PR moves the tree module under the status results panel, as it is only used there (i.e., not a common model). I suppose whoever merges first escapes the conflicts 😅

src/aiidalab_qe/plugins/bands/bands_workchain.py
edan-bainglass, line 1: Can you say more about these changes? 🙏

src/aiidalab_qe/workflows/__init__.py
```diff
@@ -248,7 +248,8 @@ def inspect_relax(self):
             self.out("structure", self.ctx.current_structure)
 
     def should_run_plugin(self, name):
-        return name in self.inputs
+        """Return whether a plugin was requested by the selected properties."""
+        return name in self.inputs.properties.get_list()
```
edan-bainglass, line 252: I suppose `name in self.inputs` was a check on the input namespaces, whereas here you check instead in the properties. Was there an issue?

Dockerfile
```diff
@@ -85,7 +85,7 @@ COPY --chown=${NB_UID}:${NB_GID} setup.cfg pyproject.toml LICENSE README.md ${QE
 RUN --mount=from=uv,source=/uv,target=/bin/uv \
     --mount=type=cache,sharing=locked,target=${UV_CACHE_DIR},uid=${NB_UID},gid=${NB_GID} \
     uv pip install --strict --no-build-isolation euphonic==1.3.2 && \
-    uv pip install --strict . ${AIIDA_HQ_PKG} ${MUON_PKG} aiidalab-qe-vibroscopy aiida-bader
+    uv pip install --strict . ${AIIDA_HQ_PKG} ${MUON_PKG} aiidalab-qe-vibroscopy "aiida-bader>=0.1.10"
```
edan-bainglass, line 88: What was the issue with aiida-bader?

Dockerfile
```diff
@@ -126,8 +126,8 @@ RUN --mount=from=qe_conda_env,source=${QE_DIR},target=${QE_DIR} \
     verdi code create core.code.installed --label python --computer=${COMPUTER_LABEL} --default-calc-job-plugin pythonjob.pythonjob --filepath-executable=/opt/conda/bin/python -n && \
     verdi code create core.code.installed --label bader --computer=${COMPUTER_LABEL} --default-calc-job-plugin bader.bader --filepath-executable=${QE_DIR}/bin/bader -n && \
     verdi code create core.code.installed --label wannier90 --computer=${COMPUTER_LABEL} --default-calc-job-plugin wannier90.wannier90 --filepath-executable=/opt/conda/bin/wannier90.x -n && \
-    # run post_install for plugin
-    python -m aiida_bader post-install && \
+    # Import Bader PAW pseudopotentials; the code is created explicitly above.
+    python -c 'from aiida import load_profile; load_profile(); from aiida_bader.utils import install_pseudos; install_pseudos()' && \
```
edan-bainglass, line 130: Why? Did the CLI command have an issue?

setup.cfg
```diff
@@ -24,33 +24,40 @@ package_dir =
     = src
 packages = find:
 install_requires =
-    aiida-core~=2.5,<3
+    aiida-core>=2.8,<3
     Jinja2~=3.0
-    aiida-quantumespresso~=4.12.1
-    aiidalab-widgets-base[optimade]~=2.5.0
+    aiida-quantumespresso>=4.17,<5
```
edan-bainglass, line 29: We may need to synchronize again with the plugins, as the QE plugin pinning is often all over the place 🥲

setup.cfg
```diff
@@ -24,33 +24,40 @@ package_dir =
...
+
+defects =
+    shakenbreak~=3.3.1
+
 dev =
     bumpver~=2023.1124
     pre-commit~=3.2
     pytest~=7.4
     pytest-regressions~=2.2
+    upf_tools~=0.1.9
```
edan-bainglass, line 60: Why also in dev?

start.py
edan-bainglass, line 1: It's not clear to me what the issue was. Can you say a bit more on this? 🙏

edan-bainglass (commented):
Thanks @cpignedoli. Left comments 🙏

Dockerfile
danielhollas: It was failing in the new image with a strange error:

> 76.87 /opt/conda/bin/python: No module named aiida_bader.__main__; 'aiida_bader' is a package and cannot be directly executed

Preferrably, this should be fixed in the `aiida-bader` plugin

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
         strategy:
             matrix:
-                python-version: ['3.9', '3.12']
+                python-version: ['3.12']
```
danielhollas, line 23: It would be great if the changes in this PR were compatible with both python 3.9 and 3.12 since it would hugely simplify the testing (and the PR could be merged before the python 3.12 was officially released.

```suggestion
                python-version: ['3.9', '3.12']
```

setup.cfg
```diff
@@ -24,33 +24,40 @@ package_dir =
     = src
 packages = find:
 install_requires =
-    aiida-core~=2.5,<3
+    aiida-core>=2.8,<3
```
danielhollas, line 27: Here again, let's not bump the minimum aiida-core version unless really needed. (at least we should still support 2.7, which is the latest supported version in the `latest` tag of full-stack image.

setup.cfg
```diff
@@ -24,33 +24,40 @@ package_dir =
...
     beautifulsoup4~=4.11
 
-python_requires = >=3.9
+python_requires = >=3.12
 
 [options.packages.find]
 where = src
 
 [options.extras_require]
+export =
```
danielhollas, line 46: Seems like there should be a bigger discussion around using optional extras here. Users that install QeApp from the AppStore will run into issues. These changes don't seem pertinent to the python 3.12 upgrade so I'd revert them in this PR.

start.py
danielhollas: The logos on the home page were not loading (I see the same issue in AtmoSpec). Not sure what the root cause is or whether the changes here are the best way to deal with it.

tests_integration/docker-compose.yml
```diff
@@ -5,9 +5,11 @@ services:
         image: ${REGISTRY:-}${QE_IMAGE:-aiidalab/qe}${TAG:-}
         environment:
             TZ: Europe/Zurich
-            DOCKER_STACKS_JUPYTER_CMD: notebook
+            DOCKER_STACKS_JUPYTER_CMD: nbclassic
```
danielhollas, line 8: Let's just remove this line, it's not needed

```suggestion
```

tests_integration/docker-compose.yml
```diff
@@ -5,9 +5,11 @@ services:
...
             SETUP_DEFAULT_AIIDA_PROFILE: 'true'
             AIIDALAB_DEFAULT_APPS: ''
             JUPYTER_TOKEN: ${JUPYTER_TOKEN}
         ports:
             - 0.0.0.0:${AIIDALAB_PORT:-8998}:8888
+        healthcheck:
+            disable: true
```
danielhollas, line 15: Just a note that on the aiidalab-docker-stack PR Sasha had a different solution
https://github.com/aiidalab/aiidalab-docker-stack/pull/479/files#diff-feb072bbafd205c351a518a31f9271b9498df144bbf7fe7dd3a30dc8eefd2833

.github/workflows/ci.yml
edan-bainglass: @danielhollas no longer applicable, right? We're jumping to 3.12?

Dockerfile
edan-bainglass: @danielhollas https://github.com/aiidaplugins/aiida-bader/pull/20

Dockerfile
```diff
@@ -7,7 +7,7 @@
 ###############################################################################
 # 1) Global ARGs
 ###############################################################################
-ARG FULL_STACK_VER=2026.1030
+ARG FULL_STACK_VER=pr-479
```
danielhollas: The PR has been merged! :tada: 

```suggestion
ARG FULL_STACK_VER=edge
```

danielhollas:
Just a note that the new python 3.12 image has mamba 2.0, so #1492 is of concern. Installation of QeApp from AppStore should be tested.

.github/workflows/ci.yml
danielhollas: It's up to you, it might be wortwhile to keep testing with 3.9 for a while, in case you want to ever backport bugfixes for older QeApp versions (if you want to keep supporting the old image). Don't know what the plan is there...

Dockerfile
edan-bainglass: To conclude, it was failing because though the CLI command existed in the `aiida-bader` as a module, it did not exist as an actual registered command. The now-merged PR above registers the command. This bit of code can safely be reverted 🙂

src/aiidalab_qe/plugins/bands/bands_workchain.py
```diff
@@ -14,6 +17,23 @@
 KpointsData = DataFactory("core.array.kpoints")
 
 
+def _validate_projwfc_bands_inputs(inputs, port):
+    """Validate projected bands inputs with the AiiDA 2.8 validator signature."""
+    return validate_pw_bands_inputs(inputs, port)
+
+
+def _patch_projwfc_bands_validator(workchain_cls=None):
+    """Patch aiida-wannier90-workflows 2.7 validator signature for AiiDA 2.8."""
```
danielhollas, line 26: Would be better to fix this in `aiida-wannier90-workflows` package

src/aiidalab_qe/workflows/__init__.py
edan-bainglass: This change appears irrelevant to the upgrade (driveby change). That said, it is likely a better implementation. At the moment, a plugin's `name` will always be in `self.inputs` if installed, regardless of selection. If it is not selected, we pop it later from the builder. So the end result is correct but takes an indirect path to get there. It is better to test if we `should_run_plugin` by comparing against the requested properties. I will open a separate PR for this.

tests_integration/docker-compose.yml
edan-bainglass: Since that solution is accepted, perhaps it is best that we apply the same here? @yakutovicha

Dockerfile
edan-bainglass: This is no longer required. The latest aiida-bader is compatible.

Dockerfile
edan-bainglass: No longer needed, as aiida-bader now have the missing CLI command registered (see PR)

setup.cfg
edan-bainglass: @danielhollas is this still correct?

setup.cfg
```diff
@@ -24,33 +24,40 @@ package_dir =
...
-    aiidalab-widgets-base[optimade]~=2.5.0
+    aiida-quantumespresso>=4.17,<5
+    aiidalab-widgets-base[optimade]>=3.0.0a0,<4
     aiida-pseudo~=1.4
     filelock~=3.8
-    importlib-resources~=5.2
+    importlib-resources>=5.2,<8
     aiida-wannier90-workflows==2.7.0
-    anywidget==0.9.13
+    anywidget>=0.9.13,<1
```
edan-bainglass: Why?

setup.cfg
```diff
@@ -24,33 +24,40 @@ package_dir =
...
+    aiida-core>=2.8,<3
     Jinja2~=3.0
-    aiida-quantumespresso~=4.12.1
-    aiidalab-widgets-base[optimade]~=2.5.0
+    aiida-quantumespresso>=4.17,<5
+    aiidalab-widgets-base[optimade]>=3.0.0a0,<4
     aiida-pseudo~=1.4
     filelock~=3.8
-    importlib-resources~=5.2
+    importlib-resources>=5.2,<8
```
edan-bainglass: Why?

setup.cfg
edan-bainglass: Agreed!

setup.cfg
danielhollas: Up to tou how you want to do it.

start.py
edan-bainglass: Found the issue. Appmode was redirecting non-notebook URLs (e.g., static files) to /files/. The method to do this (imported from jupyter server) was made async in Jupyter Server 1.1.1. Appmode still calls them synchronously. This was tolerated as far as Jupyter Server 1.19 (demo server), but is broken in Jupyter Server 2.16 (our recent upgrades). I've [opened a PR in appmode](https://github.com/oschuett/appmode/pull/84) to fix it.

setup.cfg
cpignedoli: Good question. There was no demonstrated incompatibility here; this range was broadened incidentally while bringing up the new environment. I restored `importlib-resources~=5.2` in b74142a. On Python 3.12.11 / AiiDA 2.8.0 it resolves to 5.13.0, and the full non-slow test suite passes (76 passed, 3 skipped).

setup.cfg
cpignedoli: Same here: I found no evidence that `anywidget==0.9.13` was incompatible, while the broader range would admit untested 0.10 and 0.11 releases. I restored the exact pin in b74142a. It was tested on Python 3.12.11 / AiiDA 2.8.0 with the full non-slow suite (76 passed, 3 skipped).

src/aiidalab_qe/app/configuration/advanced/pseudos/utils.py
edan-bainglass: Okay, I realize this was only done because `upf_tools` was made an optional dependency in another commit of this PR. Though in principle likely the way to go, to not hold up #1540, this part (and the optional dependency restructure) will not be migrated there. Further discussion on the topic is deferred to https://github.com/aiidalab/aiidalab-qe/issues/1555.

src/aiidalab_qe/common/process/tree.py
edan-bainglass: Okay, this part is nice, but is unrequired for the (liekly) reasons it was implemented - to avoid circular dependencies with the plugins. Plugins import things from the common package, which is where the tree resides as of this PR, hence the circular dependencies. This was resolved in #1496 by moving the status tree into the status panel of the results step (inside the app package, from which no plugin imports). However, the changes in this PR go a step forward and defer the extension of `TITLE_MAPPING` from plugins to just when they are used. This is interesting and should be explored further (currently not cached and rebuilt per process - super inefficient), as it may reduce app load time, in the spirit of lazy loading.

edan-bainglass: See issue https://github.com/aiidalab/aiidalab-qe/issues/1556

edan-bainglass: Keeping off #1540

src/aiidalab_qe/plugins/bands/bands_workchain.py
edan-bainglass: This is only needed because of the bumping of aiida-quantumespresso to >=4.17, which is not actually required to support the new stack, as far as my testing shows (I've been testing on 4.12.1 this whole time). Will not include in migration to #1540

edan-bainglass:
Superseded by #1540

PR #1513, opened 2026-05-22
Fix blockers system

edan-bainglass:
This PR fixes various issues with the blockers system. It also extends the mechanism to all models, for furture support.
In passing, the PR fixes a few inconsistencies with state handling and component updating.

For context, a blocking system was introduced a while back to the app to allow blocking wizard steps for any given reason. For example, the submission step is blocked if QE or its codes are still being installed. Since, the system was expanded in the configuration step to block for unreasonable configuration. This is implemented in a way that allows each panel model to define its own blocker(s). However, the system was incomplete, evident by various inconsistencies and bugs discovered along the way. This PR aims to resolve these issues, as well as introduce `BLOCKED` as a new actionable wizard step state, so that we can better track blocking events.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ output-format = "full"
 target-version = "py39"
 
 [tool.ruff.lint]
-ignore = ["E501", "E402", "TRY003", "RUF012", "N806"]
+ignore = ["ARG002", "E501", "E402", "TRY003", "RUF012", "N806"]
 select = [
   "A",    # flake8-builtins
   "ARG",  # flake8-unused-arguments
--- a/src/aiidalab_qe/app/configuration/advanced/convergence/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/convergence/model.py
@@ -50,7 +50,7 @@ class ConvergenceConfigurationSettingsModel(
 
     include = True  # build-in panel
 
-    def update(self, specific=""):
+    def _update(self, specific=""):
         if specific == "structure":
             self._update_help_message()
         if specific in {"structure", "protocol"}:
--- a/src/aiidalab_qe/app/configuration/advanced/hubbard/hubbard.py
+++ b/src/aiidalab_qe/app/configuration/advanced/hubbard/hubbard.py
@@ -86,7 +86,7 @@ class HubbardConfigurationSettingsPanel(
     def _on_eigenvalues_definition(self, _):
         self._toggle_eigenvalues_widget()
 
-    def _update(self):
+    def _update_ui(self):
         self._show_loading()
         self._build_hubbard_widget()
         self._toggle_hubbard_widget()
@@ -171,18 +171,20 @@ class HubbardConfigurationSettingsPanel(
                     [
                         lambda eigenvalues,
                         kind_index=kind_index,
-                        state_index=state_index: str(
-                            int(eigenvalues[kind_index][0][state_index][-1])
+                        state_index=state_index: (
+                            str(int(eigenvalues[kind_index][0][state_index][-1]))
                         ),
                         lambda value,
                         kind_index=kind_index,
                         state_index=state_index,
-                        kind_name=kind_name: update(
-                            kind_index,
-                            1,
-                            state_index,
-                            kind_name,
-                            float(value),
+                        kind_name=kind_name: (
+                            update(
+                                kind_index,
+                                1,
+                                state_index,
+                                kind_name,
+                                float(value),
+                            )
                         ),
                     ],
                 )
@@ -204,18 +206,20 @@ class HubbardConfigurationSettingsPanel(
                     [
                         lambda eigenvalues,
                         kind_index=kind_index,
-                        state_index=state_index: str(
-                            int(eigenvalues[kind_index][1][state_index][-1])
+                        state_index=state_index: (
+                            str(int(eigenvalues[kind_index][1][state_index][-1]))
                         ),
                         lambda value,
                         kind_index=kind_index,
                         state_index=state_index,
-                        kind_name=kind_name: update(
-                            kind_index,
-                            2,
-                            state_index,
-                            kind_name,
-                            float(value),
+                        kind_name=kind_name: (
+                            update(
+                                kind_index,
+                                2,
+                                state_index,
+                                kind_name,
+                                float(value),
+                            )
                         ),
                     ],
                 )
--- a/src/aiidalab_qe/app/configuration/model.py
+++ b/src/aiidalab_qe/app/configuration/model.py
@@ -60,9 +60,8 @@ class ConfigurationStepModel(
             </div>
         """
 
-    def update(self):
+    def _update(self, specific=""):
         self.update_relaxation_options()
-        self.update_blockers()
 
     def update_relaxation_options(self):
         if self.has_pbc:
@@ -127,7 +126,9 @@ class ConfigurationStepModel(
                 model.set_model_state(state[identifier])
 
     def update_state(self):
-        if self.confirmed:
+        if self.is_blocked:
+            self.state = State.BLOCKED
+        elif self.confirmed:
             self.state = State.SUCCESS
         elif self.is_previous_step_successful and self.has_all_dependencies:
             self.state = State.CONFIGURED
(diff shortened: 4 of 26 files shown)


src/aiidalab_qe/app/configuration/advanced/convergence/model.py
```diff
@@ -50,7 +50,7 @@ class ConvergenceConfigurationSettingsModel(
 
     include = True  # build-in panel
 
-    def update(self, specific=""):
+    def _update(self, specific=""):
```
edan-bainglass, line 53: This is a breaking change that has already been addressed in the only plugin that used it (https://github.com/aiidaplugins/aiida-qe-xspec/pull/37)

src/aiidalab_qe/app/configuration/advanced/hubbard/hubbard.py
```diff
@@ -86,7 +86,7 @@ def _on_hubbard_activation(self, _):
     def _on_eigenvalues_definition(self, _):
         self._toggle_eigenvalues_widget()
 
-    def _update(self):
+    def _update_ui(self):
```
edan-bainglass, line 89: Same comment as above

src/aiidalab_qe/app/configuration/model.py
```diff
@@ -127,7 +126,9 @@ def set_model_state(self, state: dict):
                 model.set_model_state(state[identifier])
 
     def update_state(self):
-        if self.confirmed:
+        if self.is_blocked:
+            self.state = State.BLOCKED
```
edan-bainglass, line 130: Blocking a step is now an official state we can react to. We use this to also change the color of a blocked step to red (same as failed) to better alert the user of a blocking issue.

src/aiidalab_qe/app/configuration/step.py
```diff
@@ -256,14 +255,16 @@ def _fetch_plugin_calculation_settings(self):
...
+
+            def toggle_plugin(change, identifier=identifier, panel=panel, info=info):
                 if change["new"]:
                     info.value = (
                         f"Customize {identifier} settings in <b>Step 2.2</b> if needed"
                     )
                 else:
                     info.value = ""
-                model.update()
+                panel.refresh()
```
edan-bainglass, line 267: Instead of just triggering a model update, we trigger a full panel refresh, which unlinks any widgets before updating the model, so that in case a model update triggers a UI update, we don't end up with orphan links.

src/aiidalab_qe/common/mixins.py
```diff
@@ -80,16 +82,14 @@ def get_models(self) -> t.Iterable[tuple[str, M]]:
...
         )
-        if isinstance(self, HasBlockers) and isinstance(model, HasBlockers):
-            tl.dlink(
-                (model, "blockers"),
-                (self, "blockers"),
-            )
+        tl.dlink(
+            (model, "blockers"),
+            (self, "blockers"),
+        )
```
edan-bainglass, line 92: `HasBlockers` is now placed on the root `Model`, so any `Model` can be blocked. Hence, there is no longer a need to check this on a given model.

src/aiidalab_qe/common/mixins.py
```diff
@@ -173,33 +173,25 @@ def _unconfirm(self):
...
     blockers = tl.List(tl.Unicode())
-    blocker_messages = tl.Unicode("")
 
     @property
     def is_blocked(self):
         return any(self.blockers)
 
     def update_blockers(self):
-        blockers = list(self._check_blockers())
+        blockers = list(self._check_blockers() or [])
```
edan-bainglass, line 182: `_check_blockers` yields `str` blockers, or `None` if no blockers are yielded. The latter scenario cannot be used in `list`, so we guard here with `or []`. `list([])` is just `[]` 👍

src/aiidalab_qe/common/mixins.py
```diff
@@ -173,33 +173,25 @@ def _unconfirm(self):
...
-            self.blocker_messages = ""
-
-    def _check_blockers(self):
-        raise NotImplementedError
+    def reset_blockers(self):
+        self.blockers = []
+        if isinstance(self, HasModels):
+            for _, model in self.get_models():
+                if isinstance(model, HasBlockers):
+                    model.reset_blockers()
```
edan-bainglass, line 194: In general good to have a reset for the blockers register, but more practically, this is used when refreshing panels.

src/aiidalab_qe/common/mvc.py
```diff
@@ -18,7 +20,11 @@ def __new__(cls, name, bases, classdict):
         return super().__new__(cls, name, bases, classdict)
 
 
-class Model(tl.HasTraits, metaclass=MetaHasTraitsLast):
+class Model(
+    tl.HasTraits,
+    HasBlockers,
+    metaclass=MetaHasTraitsLast,
+):
```
edan-bainglass, line 27: Extending the blocking mecahnism to all models!

src/aiidalab_qe/common/mvc.py
```diff
@@ -32,15 +38,12 @@ class Model(tl.HasTraits, metaclass=MetaHasTraitsLast):
         The identifier for this model.
     `dependencies` : `list[str]`
         A list of dependencies for this model.
-    `updated` : `bool`
-        Whether the model has been updated.
```
edan-bainglass, line 36: Unused, but also overcomplicates things at minimal-to-no gain.

src/aiidalab_qe/common/mvc.py
```diff
@@ -62,6 +65,14 @@ def update(self, specific=""):
...
             If provided, specifies the level of update.
         """
+        if self.locked or specific == "widgets":
+            return
+        self._update(specific)
+        self.update_blockers()
+        self.update_state()
+
+    def update_state(self):
+        """Update the model state."""
```
edan-bainglass, line 75: Generalizing the concept of refreshing a model at the root `Model` level.

src/aiidalab_qe/common/mvc.py
```diff
@@ -76,5 +87,9 @@ def reset(self):
         """Reset the model to present defaults."""
         pass
 
+    def _update(self, specific=""):
+        """Internal method to update the model."""
+        pass
```
edan-bainglass, line 92: Within the generalization, this allows each specific `Model` subclass to define what it means by "updating", in addition to the general part.

src/aiidalab_qe/common/panel.py
```diff
@@ -103,9 +107,6 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
-CSM = t.TypeVar("CSM", bound=ConfigurationSettingsModel)
```
edan-bainglass, line 106: Missed this in a previous PR deprecating `ConfigurationSettingsModel`

src/aiidalab_qe/common/panel.py
```diff
@@ -44,6 +44,10 @@ class PanelModel(Model, Confirmable, HasBlockers):
 
     include = tl.Bool(False)
 
+    def update_blockers(self):
+        if self.include:
+            super().update_blockers()
```
edan-bainglass, line 49: For `PanelModel` instances, we want to update the blocking system ONLY if the panel is included.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ output-format = "full"
 target-version = "py39"
 
 [tool.ruff.lint]
-ignore = ["E501", "E402", "TRY003", "RUF012", "N806"]
+ignore = ["ARG002", "E501", "E402", "TRY003", "RUF012", "N806"]
```
edan-bainglass, line 30: ARG002 to avoid ruff complaining about the often unused `specific` argument of `update`

edan-bainglass:
@AndresOrtegaGuerrero just finished a self-review. Hope it helps you review this medium PR 🙂 Happy to answer any questions 🙏

src/aiidalab_qe/app/configuration/advanced/convergence/model.py
danielhollas: If this method is being used in the plugins, why make it private?

danielhollas (commented):
Just had a quick look, seems reasonable, but I am not really familiar with the current QeApp codebase.

yakutovicha (approved):
We discussed the PR in a call, the changes make sense to me.

PR #1514, opened 2026-05-23
Aiida core 2.8 py39

cpignedoli:
Hi @danielhollas and @edan-bainglass,

Following #1512, this PR provides an example of the QE app working with Python 3.9 and AiiDA core 2.8.

I tested it locally only for bulk silicon. **As for #1512, please feel free not to merge this branch directly, but rather to cherry-pick the relevant fixes if that is more convenient.**

My only strategic concern is whether this intermediate Python 3.9 / AiiDA 2.8 compatibility step is still worth pursuing, given the pressure to move to Python 3.12 while Python 3.9 is becoming obsolete and Python 3.14 is already on the horizon.

## Summary

This PR contains the Python 3.9-compatible AiiDA core 2.8 fixes for the QE app.

It is intentionally separate from the larger Python 3.12 / Notebook 7 / nbclassic work in #1512. The goal here is only to make the current Python 3.9 app stack work with AiiDA core 2.8 and the compatible QE plugin stack.

## Main Changes

- Allow AiiDA core 2.8 while keeping Python 3.9 support:
  - `aiida-core>=2.8,<3`
  - Python 3.9 uses `aiida-quantumespresso>=4.15,<4.16`
  - Python >=3.10 uses `aiida-quantumespresso>=4.17,<5`
  - Python 3.9 keeps `aiidalab-widgets-base~=2.5`
  - Python >=3.10 can resolve `aiidalab-widgets-base>=3.0.0a0,<4`

- Fix QE workflow execution on AiiDA 2.8.

- Fix first-start code setup checks by explicitly loading the AiiDA profile before checking or creating QE codes.

- Fix live CalcJob output display on AiiDA 2.8 / Python 3.9 by ensuring worker threads have an asyncio event loop.

- Improve pseudo installation robustness:
  - Retry individual `aiida-pseudo install ...` commands.
  - Preserve useful stdout/stderr when pseudo installation ultimately fails.
  - Add Docker-build retry around pseudo downloads.

- Align Docker CI with an AiiDA 2.8-compatible base image, while avoiding unrelated Python 3.12 / nbclassic app changes.

- Fix integration-test compatibility:
  - Let pytest handle notebook readiness instead of Docker Compose healthcheck gating.
  - Use a stable visible selector for the “Step 1: Select structure” accordion instead of the old `p-Accordion-child` class.
  - Do not fail CI if no screenshots are produced.

## Tested Locally

Test environment:

- Python: `3.9.13`
- AiiDA core: `2.8.0`
- aiida-quantumespresso: `4.15.0`
- AiiDAlab: `26.2.0`
- aiidalab-widgets-base: `2.5.0`
image

Commands run locally:

```bash
python -m pip install -e .[dev]
python -m pip check
python -m pytest -q tests --skip-slow

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -110,8 +110,13 @@ RUN set -ex; \
 
 ENV PSEUDO_FOLDER=/tmp/pseudo
 
-RUN mkdir -p ${PSEUDO_FOLDER} && \
-    python -m aiidalab_qe download-pseudos --dest ${PSEUDO_FOLDER}
+RUN set -ex; \
+    mkdir -p ${PSEUDO_FOLDER}; \
+    for attempt in 1 2 3; do \
```
danielhollas, line 115: I understand what pain this is trying to solve, but at the same time I am not sure we should be complicating our Dockerfile to paper over GitHub's poor reliability like this. (don't feel strongly, just wanted to flag this)

src/aiidalab_qe/common/widgets.py
```diff
@@ -338,6 +339,11 @@ def _fetch_output(self, calcjob):
 
     def _push_output(self, calcjob_uuid, delay=0.2):
         """Push new log lines onto the queue."""
+        try:
```
danielhollas, line 342: Thank you for this fix, but this feels like a hack. I think we should really try to understand why this is needed. We don't use asyncio explicitly in this file so it is confusing, and it might be just papering over a bug elsewhere. If it is needed here, are there other widgets / places where it is needed?

danielhollas (commented):
There are still a quite a lot of changes here that don't seem pertinent to aiida-core 2.8 support so a bit hard to review. But my main concern is about the asyncio issue.

src/aiidalab_qe/common/widgets.py
cpignedoli: Hi @danielhollas, I fully agree. Let's just consider these modifications as an indication of some tricky points that require some care. This, however, goes beyond my capability, that's why I consider this PR just as a possible starting point that does not necessarily need to be merged

Dockerfile
cpignedoli: this could be a memory problem, see e.g. #1516

cpignedoli:
This was a test, and for the moment looks like we will skip this step and switch directly to Python 3.12

PR #1519, opened 2026-06-05
Fix code loading bug

edan-bainglass:
To prevent users from submitting calculations with unconfigured codes, we filter them out:
https://github.com/aiidalab/aiidalab-qe/blob/9ba5c427838b0a171495a9376981834006767d31/src/aiidalab_qe/common/code/model.py#L136-L142

However, this SHOULD NOT happen when loading a process, since the user cannot submit in that state. But it was, so step 3 was emitting blockers due to unselected codes. Prior to #1513, the blockers didn't do their job of blocking the step, but now they do. So if someone loads the app from a process that was run by someone else (e.g., a downloaded example), step 3 is blocked and step 4 never loads.

This PR adds a `filter_codes_for_user` flag injected through the resources refresh mechanism that is generally `True` but set to `False` if we're loading a process. In addition, the PR relaxes default code handling, falling back on the first available code option if the default code does not exist. Finally, the PR removes the blocking of step 3 w.r.t the local QE installation status, since users should still be allowed to proceed with remote submission.

### DIFF
--- a/src/aiidalab_qe/app/submission/global_settings/model.py
+++ b/src/aiidalab_qe/app/submission/global_settings/model.py
@@ -47,6 +47,11 @@ class GlobalResourceSettingsModel(
         self.global_codes = self.get_model_state()["codes"]
 
     def update_active_codes(self):
+        """Toggle code selectors conditional on plugin activity and required parameters.
+
+        For a given code (e.g., pw), if at least one active plugin requires it (if any condition
+        registered for the code is met), we activate it.
+        """
         for identifier, code_model in self.get_models():
             if identifier != "quantumespresso__pw":
                 code_model.deactivate()
--- a/src/aiidalab_qe/app/submission/model.py
+++ b/src/aiidalab_qe/app/submission/model.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import ipywidgets as ipw
 import traitlets as tl
 from IPython.display import Javascript, display
 
@@ -64,6 +63,7 @@ class SubmissionStepModel(
         super().confirm()
         if not self.has_process:
             self._submit()
+        self.lock()
 
     def _update(self, specific=""):
         self.update_process_label()
@@ -150,7 +150,10 @@ class SubmissionStepModel(
             return
         self.process_label = self.process.label
         self.process_description = self.process.description
-        self.locked = True
+
+    def refresh_codes(self, filter_codes_for_user: bool = True):
+        for _, resource_model in self.get_models():
+            resource_model.refresh_codes(filter_codes_for_user=filter_codes_for_user)
 
     def get_model_state(self) -> dict:
         return {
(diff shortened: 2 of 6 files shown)


src/aiidalab_qe/app/submission/global_settings/model.py
```diff
@@ -47,6 +47,11 @@ def update_global_codes(self):
         self.global_codes = self.get_model_state()["codes"]
 
     def update_active_codes(self):
+        """Toggle code selectors conditional on plugin activity and required parameters.
+
+        For a given code (e.g., pw), if at least one active plugin requires it (if any condition
+        registered for the code is met), we activate it.
+        """
```
edan-bainglass, line 54: Driveby change - just adding some documentation to clarify what this method is doing. Unrelated to this PR.

src/aiidalab_qe/app/submission/model.py
```diff
@@ -64,6 +63,7 @@ def confirm(self):
         super().confirm()
         if not self.has_process:
             self._submit()
+        self.lock()
```
edan-bainglass, line 66: This is the best logical point for the locking mechanism. If step 3 is confirmed, it should be locked to prevent further submissions.

src/aiidalab_qe/app/submission/model.py
```diff
@@ -150,7 +150,10 @@ def update_process_metadata(self):
             return
         self.process_label = self.process.label
         self.process_description = self.process.description
-        self.locked = True
+
+    def refresh_codes(self, filter_codes_for_user: bool = True):
+        for _, resource_model in self.get_models():
+            resource_model.refresh_codes(filter_codes_for_user=filter_codes_for_user)
```
edan-bainglass, line 156: Helper method used when loading a process.

src/aiidalab_qe/app/submission/model.py
```diff
@@ -207,20 +210,6 @@ def _submit(self):
...
-            if len(dependency_parts) == 1:  # from parent, e.g. input_structure
-                target_model = self
-                trait = dependency
-            else:  # from sibling, e.g. workchain.protocol
-                sibling, trait = dependency_parts
-                target_model = self.get_model(sibling)
-            ipw.dlink(
-                (target_model, trait),
-                (model, trait),
-            )
```
edan-bainglass, line 222: This is a remnant that avoided removal in previous PRs. Effectively, it was leading to a broken state, as the parent `_link_model`, in additional to already covering this logic, has additional logic that is missed by having this override here (e.g., wiring the `locked` and `blockers` traits).

src/aiidalab_qe/app/wizard/model.py
```diff
@@ -45,6 +45,8 @@ def load_from_state(self, state: dict):
                         SubmissionStepModel,
                         self.get_model("submit"),
                     )
+                    if process_uuid is not None:
+                        submission_model.refresh_codes(filter_codes_for_user=False)
```
edan-bainglass, line 49: Th key change - if there's a process, don't filter out codes!

src/aiidalab_qe/app/submission/model.py
```diff
@@ -256,12 +245,6 @@ def _check_blockers(self):
         if not self.input_parameters:
             yield "No selected input parameters"
 
-        if self.installing_qe:
-            yield "Installing Quantum ESPRESSO codes..."
-
-        if not self.qe_installed:
-            yield "Quantum ESPRESSO is not yet installed"
-
```
edan-bainglass, line 264: Users should be allowed to proceed with remote submission on a machine that hosts some QE installation.

src/aiidalab_qe/common/code/model.py
```diff
@@ -68,17 +68,20 @@ def activate(self):
...
+            self.options = self._get_codes(user_email, filter_codes_for_user)
             if default_code:
                 try:
                     selected = orm.load_code(default_code).uuid
                 except NotExistent:
-                    selected = None
-                    self.warning = self._WARNING_TEMPLATE.format(
-                        warning=f"Code '{default_code}' not found"
-                    )
+                    selected = self.first_option
```
edan-bainglass, line 84: `None` would block the step, but there may be valid code options, e.g., remote ones.

yakutovicha (approved):
The code looks good. I've also tested it for some discovered failures - it works flawlessly now.

PR #1523, opened 2026-06-06
Add retry mecanism around `run_cmd`

edan-bainglass:
Our CI keeps failing due to likely transient network with fetching pseudos, particularly from the PseudoDojo family. This commit adds a retry mechanism around the `run_cmd` function to mitigate this issue.


edan-bainglass:
Doesn't solve the issue. It is still unclear as to why the `python -m aiidalab_qe download-pseudos` command works seamlessly on local but fails nearly every time on remote (particularly for the PseudoDojo family).

PR #1526, opened 2026-06-10
Fix pseudo state and blockers propagation

edan-bainglass:
This PR addresses the following two issues:
1. The new functional update method was not guarded for an empty pseudos dictionary, which was incorrectly leading to `None` functionals.
2. The blockers were overwriting one another rather then accumulating, which is now correctly implemented for wizard steps and the advanced settings (which have sub-models).

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/advanced.py
+++ b/src/aiidalab_qe/app/configuration/advanced/advanced.py
@@ -85,6 +85,10 @@ class AdvancedConfigurationSettingsPanel(
             self._on_spin_type_change,
             "spin_type",
         )
+        self._model.observe(
+            self._on_blockers_change,
+            "blockers",
+        )
 
         self.sub_settings: dict[str, ConfigurationSettingsPanel] = {
             "general": self.general,
@@ -129,6 +133,16 @@ class AdvancedConfigurationSettingsPanel(
     def _on_spin_type_change(self, _):
         self._update_tabs()
 
+    def _on_advanced_tab_change(self, change):
+        tab: ConfigurationSettingsPanel = self.advanced_tabs.children[change["new"]]  # type: ignore
+        tab.render()
+
+    def _on_reset_to_defaults_button_click(self, _):
+        self._reset()
+
+    def _on_blockers_change(self, _):
+        self._model.update_blockers()
+
     def _update_tabs(self):
         if not self.rendered:
             return
@@ -148,13 +162,6 @@ class AdvancedConfigurationSettingsPanel(
         # Reset tab selection to the first tab (rendered by default) to avoid
         # possible redirection to an un-rendered tab
         self.advanced_tabs.selected_index = 0
-
-    def _on_advanced_tab_change(self, change):
-        tab: ConfigurationSettingsPanel = self.advanced_tabs.children[change["new"]]  # type: ignore
-        tab.render()
-
-    def _on_reset_to_defaults_button_click(self, _):
-        self._reset()
 
     def _reset(self):
         self._model.reset()
--- a/src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
@@ -328,14 +328,9 @@ class PseudosConfigurationSettingsModel(
 
         self.update_family_parameters()
 
-    def get_cutoffs_by_index(self, index: int) -> list[float]:
-        return (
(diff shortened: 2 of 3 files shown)


src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -328,14 +328,9 @@ def update_library_options(self):
...
-    def get_cutoffs_by_index(self, index: int) -> list[float]:
-        return (
-            [self.cutoffs[0][index], self.cutoffs[1][index]]
-            if len(self.cutoffs[0]) > index
-            else [0.0, 0.0]
-        )
-
     def update_functional(self):
+        if not self.dictionary or self.locked:
+            return
```
edan-bainglass, line 333: If we don't bail early here, we end up with a null functional, which incorrectly blocks the step. An empty dictionary occurs transiently here:
https://github.com/aiidalab/aiidalab-qe/blob/49ec62dcf00de762f378a9fbfa47296002dbcb27/src/aiidalab_qe/app/configuration/advanced/pseudos/model.py#L230-L236

src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -129,6 +133,16 @@ def _on_input_structure_change(self, _):
...
 
+    def _on_advanced_tab_change(self, change):
+        tab: ConfigurationSettingsPanel = self.advanced_tabs.children[change["new"]]  # type: ignore
+        tab.render()
+
+    def _on_reset_to_defaults_button_click(self, _):
+        self._reset()
+
+    def _on_blockers_change(self, _):
+        self._model.update_blockers()
```
edan-bainglass, line 144: Triggering an update here sets the advanced model's `blockers` to the sum of all sub-advanced model blockers. This is later collected by a similar mechanism in the configuration step.

src/aiidalab_qe/common/wizard.py
```diff
@@ -195,6 +195,7 @@ def _on_confirmation_change(self, _):
 
     def _on_blockers_change(self, _):
         if self._model.state is not State.INIT:
+            self._model.update_blockers()
```
edan-bainglass, line 198: The update triggers the accumulation of all configuration settings panel blockers.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -358,6 +353,13 @@ def update_functional(self):
         self._defaults["functional"] = functional
         self.functional = self._get_default("functional")
 
+    def get_cutoffs_by_index(self, index: int) -> list[float]:
```
yakutovicha, line 356: why?

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
yakutovicha: So this is to prevent the step from briefly appearing as blocked while the whole traitlets machinery is working?

src/aiidalab_qe/app/configuration/advanced/advanced.py
```diff
@@ -129,6 +133,16 @@ def _on_input_structure_change(self, _):
     def _on_spin_type_change(self, _):
         self._update_tabs()
 
+    def _on_advanced_tab_change(self, change):
```
yakutovicha, line 136: Not sure why this move is needed either.

yakutovicha (commented):
a few minor questions

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
edan-bainglass: Yes. I was experiencing flashing blockers.

src/aiidalab_qe/app/configuration/advanced/advanced.py
edan-bainglass: The moves are to keep like-methods together 🙂

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
edan-bainglass: Keeping update methods together 🙂

edan-bainglass:
> a few minor questions

Thanks @yakutovicha. Replied. Regarding the moves, as stated, I'd like to keep like-things together for ease of maintainability 🙂

yakutovicha (approved):
Let's go!

PR #1528, opened 2026-06-16
Follow calcjob output on the event loop instead of threads

### LINKED ISSUE #1491: CalcJobOutputFollower widget error with aiiida-core 2.8
So, the calculation seems to run fine, but the `aiida.out` preview just shows the following:

Screenshot 2026-04-27 at 13 03 48

_Originally posted by @mbercx in https://github.com/aiidalab/aiidalab-docker-stack/issues/482#issuecomment-4326455102_
            
The widget in question above is in `src/aiidalab_qe/common/widgets.py:CalcJobOutputFollower`. I have a brief look and there's no asyncio code (the code uses threads) so the errors are baffling.

This was reported in https://github.com/aiidalab/aiidalab-docker-stack/pull/482 but is also seen in the current `full-stack:edge` image which has aiida-core 2.8.0. 
The big change in aiida-core was to get rid of `nest_asyncio` so perhaps that could be a trigger for this bug? But also could be something entirely different.

I think this is a blocker for releasing a stable (`latest`) image with aiida-core 2.8.0
(CC @cpigneoli who was interested in it)

The first step here should be to confirm that the problem is not present with `latest` image with aiida-core 2.7.3

khsrali:
Fixes #1491
Read the commit message for more explanation

### DIFF
--- a/src/aiidalab_qe/common/widgets.py
+++ b/src/aiidalab_qe/common/widgets.py
@@ -3,13 +3,14 @@
 Authors: AiiDAlab team
 """
 
+import asyncio
 import base64
 import hashlib
+import os
 import warnings
 from copy import deepcopy
-from queue import Queue
 from tempfile import NamedTemporaryFile
-from threading import Event, Lock, Thread
+from threading import Event, Thread
 from time import time
 
 import anywidget
@@ -269,13 +270,7 @@ class CalcJobOutputFollower(traitlets.HasTraits):
     lineno = traitlets.Int()
 
     def __init__(self, **kwargs):
-        self._output_queue = Queue()
-
-        self._lock = Lock()
-        self._push_thread = None
-        self._pull_thread = None
-        self._stop_follow_output = Event()
-        self._follow_output_thread = None
+        self._follow_task = None
 
         super().__init__(**kwargs)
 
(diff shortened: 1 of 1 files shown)


src/aiidalab_qe/common/widgets.py
```diff
@@ -285,34 +280,56 @@ def _observe_calcjob(self, change):
...
+        """Schedule a coroutine on the running event loop.
+
+        Falls back to ``get_event_loop`` if no loop is currently running so the
+        widget never crashes at construction time.
+        """
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = asyncio.get_event_loop()
+        return loop.create_task(coro)
```
edan-bainglass, line 307: Is it safe to equate `RuntimeError` to "no loop is currently running"?

src/aiidalab_qe/common/widgets.py
```diff
@@ -285,34 +280,56 @@ def _observe_calcjob(self, change):
...
+
+    async def _follow_output(self, calcjob_uuid, delay=0.2):
+        """Poll the calcjob output and update traitlets until it is sealed."""
```
edan-bainglass, line 310:
```suggestion
        """Poll the calcjob output and update the traits until it is sealed."""
```

edan-bainglass (approved):
Thanks @khsrali. One question, and one docstring change for clarity. Other than that, LGTM! Tested with a relaxation + bands + pdos on Si - no event loop error messages 👍

A more general question regarding threads - with the Greenlet changes in AiiDA now in (via v2.8), do you recommend we avoid all threads in general, or is it an issue only when threads interact with AiiDA?

edan-bainglass:
Image build failure unrelated to this PR. See #1543 for details.

edan-bainglass:
@danielhollas @khsrali as discussed on the issue, this matter is a non-issue in Python 3.10+. As the intent is to move AiiDAlab apps towards a React + FastAPI architecture (where all backend ops would be async), with the QE app being the priority test case, I would consider perhaps closing this PR. Thoughts?

khsrali:
@edan-bainglass 
As we discussed in [the issue](https://github.com/aiidalab/aiidalab-qe/issues/1491) the fact that it works for 3.10+ is only by a lucky chance --not a thoughtful design--, therefore I cannot guarantee, having that said, it's perhaps ok to move on with closing this PR, and if the issue showed its face again, you can come back and merge this.

Nevertheless, it would be a great help for future to scan your ecosystem and find out how many of such use cases exists. That would help us to save time and act faster if ever this comes back.

edan-bainglass:
Thanks @khsrali. I agree with our former discussions but also am prioritizing. As stated, this matter may not longer be relevant in the new tech stack. But I will keep this in mind 🙂

khsrali:
> Thanks @khsrali. I agree with our former discussions but also am prioritizing. As stated, this matter may not longer be relevant in the new tech stack. But I will keep this in mind 🙂

Thanks @edan-bainglass 😊 
I understand your planning to move to a new tech stack. But that by no means will free you from the logic of calling on a transport in an asynchronous manner.

edan-bainglass:
My only point is that this code will not be used as is, as the QE app will be upgrading to the new tech (where we will be more careful).

As for AWB, that remains to be checked, as it will be used by legacy apps 👍

danielhollas:
Agreed with both of you: The current code kinda works by accident, but I feel much less worried about it now that I've checked the CPython code and done the analysis in https://github.com/aiidalab/aiidalab-qe/issues/1491#issuecomment-5003162607

Agreed with @edan-bainglass re prioritization, in light of planned migration. I might eventually "borrow" this PR for my app, where I use the same widget.

khsrali:
Closed as concluded from the discussion above.

PR #1531, opened 2026-06-23
Fix pseudopotential metadata handling

### LINKED ISSUE #1530: Inconsistencies in pseudo handling
#1348 and related PRs aimed to improve the pseudopotential upload feature (e.g., detecting functional/relativistic mismatch). However, as it turns out, the design is perhaps not sufficiently robust.

As it stands, on upload, we check the loaded UPF md5 against the DB to reject and load instead existing nodes. If we don't find one, and the loaded pseudo is valid (matches symbol), we store the new node and also populate its extras with the functional and relativistic values parsed from the UPF file.

The same extras population also acts on existing pseudos (the ones we install using aiida-pseudo). We do this, so that in other places in the app, we can check for functional/relativistic consistency from a single source - the extras.

However, for those pseudos from libraries, in the current version (v26.06.2), the check for functional consistency happens prior to the extras setting, leading to a global null functional. Manually selecting the functional triggers the extras setting, so going forward, that pseudopotential does not raise issues.

- **Immediate solution:** revert to an older version
- **Hacky solution:** run a script pre-app-load to populate all library pseudos with functional/relativistic info in extras (write a configuration line somewhere to not do it again?)
- **Best solution:** Unclear at the moment

Pinging @giovannipizzi

edan-bainglass:
We were mishandling the functional/relativistic information from
pseudo extras. This PR aims to resolve these issues by first guarding
against redundant functional updates, then removing a blocker from
setting the extras info, and finally by defaulting to the family if
the extras info is missing (likely not yet set).

Peripherally, the PR adds a cache for observations set at each 
rerender of the pseudopotential uploader widgets. The cache is used to
unsubscribe from the observations at each refresh (structure change)
to avoid observer accumulation.

Closes #1530

### DIFF
--- a/src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
+++ b/src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
@@ -16,9 +16,8 @@ from aiidalab_qe.parameters import DEFAULT_PARAMETERS
 from aiidalab_qe.setup.pseudos import PSEUDODOJO_VERSION, SSSP_VERSION, PseudoFamily
 
 from .utils import (
-    get_info_from_family,
+    get_info_from_family_by_md5,
     get_pseudo_family_by_label,
-    get_pseudo_family_by_md5,
     get_upf_dict,
 )
 
@@ -356,11 +355,7 @@ class PseudosConfigurationSettingsModel(
             else:
                 functional = pseudo.base.extras.get("functional", None)
                 if functional is None:
-                    # Family-based pseudos may not yet have the extra information.
-                    # We use the pseudo's md5 hash to retrieve the family and extract the info.
-                    family = get_pseudo_family_by_md5(pseudo.md5)
-                    if family is not None:
-                        functional, _ = get_info_from_family(family)
+                    functional, _ = get_info_from_family_by_md5(pseudo.md5)
                 functional_set.add(functional)
 
         functional = functional_set.pop() if len(functional_set) == 1 else None
@@ -442,11 +437,7 @@ class PseudosConfigurationSettingsModel(
                 functional = pseudo.base.extras.get("functional", None)
                 relativistic = pseudo.base.extras.get("relativistic", None)
                 if functional is None or relativistic is None:
-                    # Family-based pseudos may not yet have the extra information.
-                    # We use the pseudo's md5 hash to retrieve the family and extract the info.
-                    family = get_pseudo_family_by_md5(pseudo.md5)
-                    if family is not None:
-                        functional, relativistic = get_info_from_family(family)
+                    functional, relativistic = get_info_from_family_by_md5(pseudo.md5)
                 functional_set.add(functional)
                 relativistic_set.add(relativistic)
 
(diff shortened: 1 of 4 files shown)


src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -329,15 +334,15 @@ def update_library_options(self):
         self.update_family_parameters()
 
     def update_functional(self):
-        if not self.dictionary or self.locked:
+        if not self.dictionary or self.family or self.locked:
```
AndresOrtegaGuerrero, line 337:
```python
skip = not self.dictionary or self.family or self.locked
```

? so is easy to read ?

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
```diff
@@ -411,27 +425,36 @@ def _check_blockers(self):
...
             except AssertionError:
                 yield f"The selected pseudopotential family does not contain a pseudopotential for {kind.symbol}. Consider changing the family or uploading a custom pseudopotential."
                 return
             except exceptions.NotExistent:
                 yield f"Pseudopotential with UUID {uuid} does not exist for {kind.symbol}."
                 return
+            else:
+                functional = pseudo.base.extras.get("functional", None)
+                relativistic = pseudo.base.extras.get("relativistic", None)
+                if functional is None or relativistic is None:
```
AndresOrtegaGuerrero, line 444: This logic seems repeated , from the one above, should you write a small function to do the get_pseudo_family, and you set the condition?

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
edan-bainglass: If you're suggesting
```python
def update_functional(self):
    skip = not self.dictionary or self.family or self.locked 
    if skip:
        return
    ...
```
then no, I don't think that really makes it any easier to read.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
edan-bainglass: Yeah, I tried to do it but ran into a small issue. It tool some time to resolve the remaining issues, so I hadn't returned to this part. Let me have another look.

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
AndresOrtegaGuerrero: to me it is but ok ,

src/aiidalab_qe/app/configuration/advanced/pseudos/model.py
edan-bainglass: Okay, check it now. Marginally better? Note that at the moment, we return `None, None` for the functional and relativistic treatment if we fail to retrieve them from extras, or from the family if extras aren't yet set. One could argue that returning `None` might be masking an issue and instead we should raise an error. Thoughts?

edan-bainglass: Ugh, now tests are failing

edan-bainglass: Reverted the change. Let's not worry about repeated code right now. I'd like to get these bugs fixes asap and move on to reimplementing this whole app in React. The app's state is far too complex to be managing cleanly with traitlets.

AndresOrtegaGuerrero (approved):
LGTM!

PR #1536, opened 2026-06-26
Remove uv from Dockerfile

danielhollas:
Since we're now installing most deps into `~/.local`, and we cannot use uv for that, let's just remove it and simplify the Dockerfile.

This also joins the `build_deps` and `home_build` stages since there's not much reason to have them separate.

> [!NOTE]
> We could use the `--target ~/.local/lib/python3.x/site-packages` to instruct uv to install to user packages, but uv would install everything there, including the packages already installed in conda env (such as aiida-core)


edan-bainglass (approved):
Thanks @danielhollas, also for excluding some junk from the tar 🙂

PR #1534, opened 2026-06-25
Install app and plugins in .local

edan-bainglass:
#1337 was great for ephemeral deployments (i.e., demo server), but
does not work for persistent ones (e.g., institutional). #1337
installs the app and its plugins into conda, instead of .local.
However, existing instances prior to #1337 have the app and its
plugins in .local. Though Python defers to .local for imports, this is
 not true for entry points. The app's plugin system is based on
 loading entry points, which triggers an entry point search across a
 plugin's dependencies, which are now duplicated between .local and
 conda. This prevents the use of the given plugin.

This PR keeps most of the recent changes to the image but shifts the
installation of the app and plugins to .local. This means they are
tarred and untarred on home when the container is first deployed
(fresh volume). For existing volumes, conflicts are only in version.
No more entry point conflicts!

PR #1537, opened 2026-06-26
Deduplicate packages in QE image

danielhollas:
Packages from `aiida-core.atomic_tools` were being installed both in `.local` and `/opt/conda`.

Staged on top of #1536 

This PR saves around 400Mb, roughly the same amount gained in #1534 

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,12 @@ RUN set -ex; \
     fi && \
     tar xf hq.tar.gz -C /opt/conda/
 
+# Install common dependencies such as pymatgen and pandas via conda,
+# to avoid building them when installing via pip
+# TODO: Remove this once it is part of the full-stack image.
+RUN mamba install aiida-core.atomic_tools -y && \
+    mamba clean --all -f -y
+
 # Install the app and its plugins into the user's local Python environment
 RUN python -m pip install --user --no-cache-dir . ${MUON_PKG} aiidalab-qe-vibroscopy aiida-bader
 


Dockerfile
```diff
@@ -80,6 +80,12 @@ RUN set -ex; \
...
     tar xf hq.tar.gz -C /opt/conda/
 
+# Install common dependencies such as pymatgen and pandas via conda,
+# to avoid building them when installing via pip
+# TODO: Remove this once it is part of the full-stack image.
+RUN mamba install aiida-core.atomic_tools -y && \
+    mamba clean --all -f -y
+
 # Install the app and its plugins into the user's local Python environment
 RUN python -m pip install --user --no-cache-dir . ${MUON_PKG} aiidalab-qe-vibroscopy aiida-bader
```
danielhollas, line 90: This command was installing things like `pymatgen` and `scipy` to `~/.local`

Dockerfile
```diff
@@ -80,6 +80,12 @@ RUN set -ex; \
     fi && \
     tar xf hq.tar.gz -C /opt/conda/
 
+# Install common dependencies such as pymatgen and pandas via conda,
+# to avoid building them when installing via pip
+# TODO: Remove this once it is part of the full-stack image.
+RUN mamba install aiida-core.atomic_tools -y && \
+    mamba clean --all -f -y
```
edan-bainglass, line 87: This is a copy of the same command in the final stage. Here we do it only to avoid the following command from installing some shared packages in .local, which would get tarred, whereas in the final image, we install them into the final image for actual use. Do I have this correct?

edan-bainglass (approved):
Thanks @danielhollas. I think I'm starting to get it 💪🙂

Dockerfile
danielhollas: Yep! Note that we will be able to remove both of them once we upgrade to the python3.12 image.

PR #1538, opened 2026-06-26
Move wannier90 build out of the final stage

danielhollas:
In general, builds shouldn't be done in the final stage since they can leave junk behind them (even though we tried hard to cleanup after ourselves).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,10 +31,20 @@ ARG QE_DIR
 ARG TARGETARCH
 
 USER ${NB_USER}
-RUN set -ex; \
-    echo "Installing QE and Bader..." && \
+RUN echo "Installing QE and Bader..." && \
     mamba create -p ${QE_DIR} --yes qe=${QE_VER} bader && \
     mamba clean --all -f -y
+
+# Build wannier90 and copy the binary to $QE_DIR conda env
+# TODO: Make a conda-forge package for wannier90!
+RUN apt-get -q update && \
+    apt-get -q install -y --no-install-recommends gfortran libblas-dev liblapack-dev libopenmpi-dev && \
+    git clone --depth=1 https://github.com/wannier-developers/wannier90.git /tmp/wannier90 && \
+    cd /tmp/wannier90 && \
+    cp config/make.inc.gfort make.inc && \
+    echo -e "COMMS=mpi\nMPIF90=mpif90" >> make.inc && \
+    make -j wannier && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -36,6 +36,19 @@ RUN set -ex; \
...
+RUN set -ex; \
+    apt-get -q update && apt-get -q install -y --no-install-recommends \
+    gfortran libblas-dev liblapack-dev libopenmpi-dev; \
+    git clone --depth=1 https://github.com/wannier-developers/wannier90.git /tmp/wannier90; \
+    cd /tmp/wannier90; \
+    cp config/make.inc.gfort make.inc; \
+    echo "COMMS=mpi" >> make.inc; \
+    echo "MPIF90=mpif90" >> make.inc; \
+    make -j"$(nproc)" wannier; \
+    cp wannier90.x ${QE_DIR/bin/wannier90.x
```
edan-bainglass:
```suggestion
    cp wannier90.x ${QE_DIR}/bin/wannier90.x
```

edan-bainglass (approved):
Thanks @danielhollas. LGTM!

Note that I'd like to soon add also the compilation of SKEAF into the image, which is needed for the calculation of Fermi surfaces by the Wannier90 plugin. If you recall, the workflow required Julia. That is no longer the case - only Python! We just need the SKEAF code installed + aiida-skeaf.

For the SKEAF code, I started looking at placing it on conda-forge. Seems doable.

danielhollas:
Yes, ideally wannier90 should be in conda-forge as well, should be relatively easy. This should make the build faster and also the image smaller I suspect.

Dockerfile
```diff
@@ -180,6 +181,8 @@ RUN python -m pip install --no-user --no-cache-dir ${AIIDA_HQ_PKG}
 COPY --from=home_build /opt/conda/hq /usr/local/bin/
 # Copy the QE conda environment
 COPY --from=qe_conda_env ${QE_DIR} ${QE_DIR}
+# This is for backwards compatibility
+COPY --from=qe_conda_env ${QE_DIR}/bin/wannier90.x /opt/conda/bin/wannier90.x
```
edan-bainglass, line 185: @danielhollas can you explain how this is taking care of backwards compatibility? What exactly was the issue?

danielhollas: Existing deployments will have AiiDA Codes pointing to this path so we need to preserve it.

PR #1540, opened 2026-06-26
Python 3.12 + nbclassic + aiida-core 2.9

danielhollas:
Split from #1512, curious what the minimum set of changes is to get the build and tests passing.


danielhollas:
This PR might be a good springboard for full python3.12 support. @edan-bainglass feel free to push here, I don't plan to work on it further. But I'd suggest that any further fixes from #1512 should start with a failing test, and fixes related to aiida-core 2.8 could go via separate PRs.

edan-bainglass:
> This PR might be a good springboard for full python3.12 support. @edan-bainglass feel free to push here, I don't plan to work on it further. But I'd suggest that any further fixes from #1512 should start with a failing test, and fixes related to aiida-core 2.8 could go via separate PRs.

Looking into this now. Regarding aiida-core 2.8, at the moment, this PR WILL install aiida-core 2.8, as it is pinned in /opt/requirements.txt (by the base:edge image). Are you suggesting to pin it back to ~2.7.x as we continue to assess Python 3.12 and nbclassic, leaving aiida-core 2.8 support for another PR?

danielhollas:
I am suggesting that aiida-core 2.8 support should be a pre-requisite for this PR (i.e. reviewing and merging #1528).

edan-bainglass:
> I am suggesting that aiida-core 2.8 support should be a pre-requisite for this PR (i.e. reviewing and merging #1528).

Thanks @danielhollas. Clear! I've now reviewed #1528, which itself looks good to me. However, as I mentioned on that PR, I'd like to make sure we don't make similar "threading" misuse anywhere else in our ecosystem before we proceed. Let's see how Ali responds.

One thing to note is that the aiida-pythonjob 0.2.5, on which several QE app plugins depend, is incompatible with aiida-core 2.8. Alex fixed the issue in a much later version (~0.4.9). Just noting here that updating the plugins may be a bit of work considering the major jump in version. Maybe easier to add a support branch off 0.2.5. I'll ping the relevant parties.

edan-bainglass:
Recording here that there appears to be something wrong with the rendering of the pip install part in plugins:
```
Installing collected packages: smmap, gitdb, gitpython, aiida-qe-xspec
[?25l
[2K [91m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━[0m[90m╺[0m[90m━━━━━━━━━[0m [32m3/4[0m [aiida-qe-xspec]
[2K [90m━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━[0m [32m4/4[0m [aiida-qe-xspec]
[?25h
[1A[2KSuccessfully installed aiida-qe-xspec-0.3.1 gitdb-4.0.12 gitpython-3.1.57 smmap-5.0.3
```

danielhollas:
> Recording here that there appears to be something wrong with the rendering of the pip install part in plugins:

This might be fixed in the recent pip 26.2 version, we're still on 26.1
Does this occur in CI on GitHub? Or where exactly do you see it?

edan-bainglass:
> > Recording here that there appears to be something wrong with the rendering of the pip install part in plugins:
> 
> This might be fixed in the recent pip 26.2 version, we're still on 26.1 Does this occur in CI on GitHub? Or where exactly do you see it?

In the QE app's plugin store, from which we install plugins via pip.

edan-bainglass:
Another apparent issue is that some options-based widgets no longer auto select their first option by default (e.g., dropdowns, radio buttons). This is breaking code that assumed this.

danielhollas:
> Another apparent issue is that some options-based widgets no longer auto select their first option by default (e.g., dropdowns, radio buttons). 

Interesting, so this means we always need to set the default value explicitly? We might need to audit AWB widgets for this as well.

edan-bainglass:
> Recording here that there appears to be something wrong with the rendering of the pip install part in plugins:

Resolved with new pip version 👍

edan-bainglass:
@danielhollas @yakutovicha I discard Python 3.9 from the testing suite. It's not only that it is no longer supported, it is actually no longer compatible! More specifically, dependencies are no longer compatible with 3.9.

edan-bainglass:
Okay. Looks good. I'll check in the morning if there remains anything else from #1512 to port over.

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
         strategy:
             matrix:
-                python-version: ['3.9', '3.12']
+                python-version: ['3.12']
                 aiida-core-version: ['2.8']
```
danielhollas:
```suggestion
                 aiida-core-version: ['2.9']
```

danielhollas:
> Another apparent issue is that some options-based widgets no longer auto select their first option by default (e.g., dropdowns, radio buttons).

Just curious, has this been addressed? If so in which commit? (asking in case I need to port these fixes to my app as well)

edan-bainglass:
> > Another apparent issue is that some options-based widgets no longer auto select their first option by default (e.g., dropdowns, radio buttons).
> 
> Just curious, has this been addressed? If so in which commit? (asking in case I need to port these fixes to my app as well)

Not yet addressed

edan-bainglass:
> > > Another apparent issue is that some options-based widgets no longer auto select their first option by default (e.g., dropdowns, radio buttons).
> > 
> > 
> > Just curious, has this been addressed? If so in which commit? (asking in case I need to port these fixes to my app as well)
> 
> Not yet addressed

So `_Selection` widgets still select their first option on instantiation. What changed (in https://github.com/jupyter-widgets/ipywidgets/pull/3284) is the behavior of dynamic options setting on a widget whose value is explicitly `None`.

### v7.7.4
```python
    @observe('options')
    def _propagate_options(self, change):
        "Set the values and labels, and select the first option if we aren't initializing"
        options = self._options_full
        self.set_trait('_options_labels', tuple(i[0] for i in options))
        self._options_values = tuple(i[1] for i in options)
        if self._initializing_traits_ is not True:
            if len(options) > 0:
                if self.index == 0:
                    # Explicitly trigger the observers to pick up the new value and
                    # label. Just setting the value would not trigger the observers
                    # since traitlets thinks the value hasn't changed.
                    self._notify_trait('index', 0, 0)
                else:
                    self.index = 0
            else:
                self.index = None
```
                
### v8.x

```python
    @observe('options')
    def _propagate_options(self, change):
        "Set the values and labels, and select the first option if we aren't initializing"
        options = self._options_full
        self.set_trait('_options_labels', tuple(i[0] for i in options))
        self._options_values = tuple(i[1] for i in options)

        if self.index is None:
            # Do nothing, we don't want to force a selection if
            # the options list changed
            return

        if self._initializing_traits_ is not True:
            if len(options) > 0:
                if self.index == 0:
                    # Explicitly trigger the observers to pick up the new value and
                    # label. Just setting the value would not trigger the observers
                    # since traitlets thinks the value hasn't changed.
                    self._notify_trait('index', 0, 0)
                else:
                    self.index = 0
            else:
                self.index = None
```
So we now must explicitly select the first option when setting options dynamically!

See discussion here - https://github.com/jupyter-widgets/ipywidgets/issues/2230#issuecomment-425693572. Despite the burden, this is actually something that was an issue for me in the past, where dynamic options setting selected the first option, which undesirably triggered a subscriber callback.

IMO, this is a good change. It gives us more control on if and when we want to auto-select an option. The solution then is not general - we simply need to explicitly do `widget.value = options[0]` where we want to.

### Example where we don't want to auto-select an option

The QE app examples widget fetches examples from an examples repo to populate a select widget, from which the user can select an example to download and import into their instance. We DON'T want to select any for them 🙂

edan-bainglass:
> The solution then is not general - we simply need to explicitly do `widget.value = options[0]` where we want to.

@danielhollas @yakutovicha I've looked across AWB and the QE app ecosystem and found no instances where we wish to have a default option selected and don't already set it explictly.

edan-bainglass:
@yakutovicha this one's done and ready for review! @danielhollas once we have this merged, do we plan to do a new docker-stack release? I'd like to replace `edge` here with a stable tag 🙂🙏

danielhollas:
Great news! (though the build appears to be failing?)

> do we plan to do a new docker-stack release? I'd like to replace edge here with a stable tag 🙂🙏

We should release AWB 3.0 first, are you still planning on testing the widgets?
More importantly, we should figure out the migration story first before releasing the Python 3.12 image officially. That probably means getting the Configuration page in aiidalab-home done?

edan-bainglass:
> Great news! (though the build appears to be failing?)

Right, had to update a few pins w.r.t changes in the wannier90 packages. Should pass now (or will once I'm done fixing it 😅)

> We should release AWB 3.0 first...
> are you still planning on testing the widgets?

Tested. Didn't see any immediate issues. However, I recall now your issue with the output follower not following correctly. I'll check this again.

> More importantly, we should figure out the migration story first before releasing the Python 3.12 image officially. That probably means getting the Configuration page in aiidalab-home done?

I agree. I discussed this briefly with @giovannipizzi. We have a plan! Let's discuss next AiiDAlab meeting.

danielhollas (commented):
Changes here look reasonable, thanks! (Cannot approve since it is my PR)

tests/test_result.py
```diff
@@ -56,7 +56,7 @@ def test_workchainview(generate_qeapp_workchain):
     model.process_uuid = workchain.node.uuid
     viewer.render()
     assert len(viewer.tabs.children) == 2
-    assert viewer.tabs._titles["0"] == "Structure"  # type: ignore
+    assert viewer.tabs.titles[0] == "Structure"
```
yakutovicha, line 59: Nice that they made it a public attribute 👍

.gitignore
```diff
@@ -3,6 +3,7 @@
 *.tar.gz
 *.code-workspace
 .*.ipynb
+*-[0-9]*.ipynb
```
yakutovicha, line 6: Drive-by fix, but acceptable in this case since the new image comes with the new app mode version.

Dockerfile
```diff
@@ -92,11 +86,7 @@ RUN python -m pip install --user --no-cache-dir . \
     aiidalab-qe-hp \
     aiidalab-qe-pp \
     aiida-qe-xspec \
-    # the following git-installed plugins are due to PyPI quarantines - discard when resolved
-    aiida-wannier90@git+https://github.com/aiidateam/aiida-wannier90@v2.2.0 \
-    aiida-wannier90-workflows@git+https://github.com/aiidateam/aiida-wannier90-workflows@v2.7.1 \
-    aiida-skeaf@git+https://github.com/aiidaplugins/aiida-skeaf@v0.2.0 \
-    aiidalab-qe-wannier90
+    aiidalab-qe-wannier90@git+https://github.com/aiidalab/aiidalab-qe-wannier90
```
yakutovicha, line 89: I like how cleaner it became 👍

setup.cfg
```diff
@@ -25,20 +25,19 @@ package_dir =
...
     Jinja2~=3.0
     aiida-quantumespresso~=4.12.1
-    aiidalab-widgets-base[optimade]~=2.5.0
+    aiidalab-widgets-base[optimade]~=3.0.0a3
     aiida-pseudo~=1.4
     filelock~=3.8
     importlib-resources~=5.2
-    aiida-wannier90 @ git+https://github.com/aiidateam/aiida-wannier90@v2.2.0
-    aiida-wannier90-workflows @ git+https://github.com/aiidateam/aiida-wannier90-workflows@v2.7.1
+    aiida-wannier90-workflows @ git+https://github.com/aiidateam/aiida-wannier90-workflows@v2.7.2
```
yakutovicha: unfortunate mess :(

src/aiidalab_qe/common/widgets.py
```diff
@@ -41,24 +42,30 @@
...
+        layout.overflow = "auto"
+
         super().__init__(
-            children=[
-                self._output,
-            ],
-            layout=ipw.Layout(max_height=max_output_height),
+            children=[self._output],
+            layout=layout,
+            **kwargs,
```
yakutovicha, line 68: Good catch 👍, but can you add a test for this? I guess there is no `test_widgets` atm.

src/aiidalab_qe/common/widgets.py
```diff
@@ -41,24 +42,30 @@
...
     auto_scroll = traitlets.Bool()
 
     def __init__(self, num_min_lines=10, max_output_height="200px", **kwargs):
         self._num_min_lines = num_min_lines
+        self._scroll_class = f"rolling-output-{uuid4().hex}"
+
         self._output = ipw.HTML(layout=ipw.Layout(min_width="80em"))
         self._refresh_output()
+
+        layout = kwargs.pop("layout", ipw.Layout())
```
yakutovicha: so if someone passes the layout with max_height different from 200px, we still want to set it to 200px, unless they explicitly provide max_output_height as a separate input?

Maybe something like this instead?

```python
    def __init__(self, num_min_lines=10, **kwargs):
        self._num_min_lines = num_min_lines
        self._scroll_class = f"rolling-output-{uuid4().hex}"

        ...
        layout = kwargs.pop("layout", ipw.Layout())
        if not layout.max_height:
            layout.max_height = "200px"
        layout.overflow = "auto"
```

setup.cfg
```diff
@@ -25,20 +25,19 @@ package_dir =
 packages = find:
 # Dependency pinning:
 # - hiphive==1.3.1: to avoid an spglib inconsistency
-# - phonopy<2.26.0: to avoid scikit-build-core build issues
+# - phonopy>3,<4: to avoid scikit-build-core build issues in v2.26.0 and later
```
yakutovicha: nitpicking 
```suggestion
# - phonopy>=3,<4: to avoid scikit-build-core build issues in v2.26.0 and later
```

yakutovicha (commented):
A few  comments (see below). The most important one - I couldn't start the app with numpy 2.x.
```python
<lots of text>

~/.local/lib/python3.12/site-packages/numpy/__init__.py in __getattr__(attr)
    767             )
    768 
--> 769         raise AttributeError(f"module {__name__!r} has no attribute {attr!r}")
    770 
    771     def __dir__():

AttributeError: module 'numpy' has no attribute 'cumproduct'
```

 Is it migrated? Cause when I switched to numpy 1.x it didn't work either.

tests/test_result.py
edan-bainglass: Thanks! That explains why my tests were failing in #1562, where I add a "new" `_titles` attribute (or at least what I thought was new 😅

.gitignore
edan-bainglass: I wouldn't say this is drive-by, for the reason you mention 😅

Dockerfile
edan-bainglass: It would be much cleaner once the PyPI quarantine is lifted 😅

setup.cfg
edan-bainglass: No doubt. We're still waiting to hear what happened

src/aiidalab_qe/common/widgets.py
edan-bainglass: I fixed this before I even looked at your review, after having the same thought 😅

src/aiidalab_qe/common/widgets.py
edan-bainglass: I would test this with playwright, to make sure scrolling actually occurs. But Let's do this in a separate PR, one that moves all selenium tests to playwright.

edan-bainglass (approved):
LGTM, but I'm biased 🙂

src/aiidalab_qe/common/widgets.py
yakutovicha: But the way it is written right now, if the `layout` argument in the input parameters has `max_height`, it will always be overridden. Do we want that?

edan-bainglass: I'm finding other issues with this part also in other work. I'll modify it, considering also your comment.

edan-bainglass: @yakutovicha Have another look 🙏

danielhollas:
> @danielhollas @yakutovicha I've looked across AWB and the QE app ecosystem and found no instances where we wish to have a default option selected and don't already set it explictly.

I've found one :wink: 
aiidalab/aiidalab-widgets-base#793

edan-bainglass:
> I've found one 😉
> [aiidalab/aiidalab-widgets-base#793](https://github.com/aiidalab/aiidalab-widgets-base/issues/793)

You state in the issue that we set it explicitly though 😅

danielhollas:
I don't think I do? I am saying that WizardAppWidget is the only one where we don't set `selected_index` explicitly at the moment.

edan-bainglass:
> I don't think I do? I am saying that WizardAppWidget is the only one where we don't set `selected_index` explicitly at the moment.

Right. "in all other" 👍

yakutovicha (approved):
thanks!

src/aiidalab_qe/common/widgets.py
yakutovicha: All good 👍

edan-bainglass:
There appeasrs to be an issue with the citation reader in the app store 🤔 I can't tell yet if it is related to the new stack, because this feature is in a version of the home app newer than that deployed on the demo server.

See comment by @danielhollas [here](https://github.com/aiidalab/aiidalab/pull/510#pullrequestreview-4378130926)

---

image

edan-bainglass:
> There appears to be an issue with the citation reader in the app store 🤔

Until I added citations, it was correctly assumed that all entries are string based: (from aiidalab/app.py)
```python
    def _get_from_metadata(self, what: str) -> str:
        """Get information from metadata."""
        try:
            return f"{self._app.metadata[what]}" <-------------------- 🥲
        except KeyError:
            return f'Field "{what}" is not present in app metadata.'
```
That's turning my citations list of dictionaries into a string. So instead of iterating over the list, it is iterating over the chars, yielding a lot of garbage 🤦🏻‍♂️

Works without the f-string wrapper

image

Question is - why was the wrapper introduced? So having a quick look at the metadata dataclass
```python
    title: str
    description: str
    authors: None | str = None
    state: None | str = None
    documentation_url: None | str = None
    external_url: None | str = None
    logo: None | str = None
    categories: list[str] = field(default_factory=list)
    version: None | str = None
    citations: list[SimpleCitation | StandardCitation] = field(default_factory=list)
```
suggests that it is not needed? And looks like I'm not the only list - we also had categories. Anyhow, unclear why we were force-casting to a string.

danielhollas:
Congrats! :tada:

PR #1541, opened 2026-07-06
Check blocked state only if not confirmed

edan-bainglass:
We were checking for a blocked state first, which applies also to completed jobs.
This PR moves blocking state checks after completion checks, to avoid blocking
already finished jobs.

Blocking such jobs SHOULD NOT happen in principle, but it can happen due to
bugs. This PR ensures at least that users aren't blocked for such reasons. The
blockers still appear (so we can still detect bugs), but now they don't block the
wizard.

### DIFF
--- a/src/aiidalab_qe/app/configuration/model.py
+++ b/src/aiidalab_qe/app/configuration/model.py
@@ -126,10 +126,10 @@ class ConfigurationStepModel(
                 model.set_model_state(state[identifier])
 
     def update_state(self):
-        if self.is_blocked:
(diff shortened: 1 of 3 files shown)


AndresOrtegaGuerrero (approved):
LGTM! , I wonder is there is a possibility to include a small test to check the change of the states ,?

edan-bainglass:
> LGTM! , I wonder is there is a possibility to include a small test to check the change of the states ,?

Sure. There are several things that could benefit from extending our testing suite. I'll try to revisit this soon 🙂

PR #1551, opened 2026-08-05
Keep active guide as "No Guides if specific guide is null

edan-bainglass:
In the new tech stack, the first option of radio buttons is not selected by default. In this case, this causes the guide selector to remain null when a guides category is selected. This PR prevents an attempt to parse the active guide when `guide` is `None`.

### DIFF
--- a/src/aiidalab_qe/app/app.py
+++ b/src/aiidalab_qe/app/app.py
@@ -257,7 +257,7 @@ class AppModel(tl.HasTraits):
 
     def update_active_guide(self, category, guide):
         """Sets the current active guide."""
-        active_guide = f"{category}/{guide}" if category != "No guides" else category
+        active_guide = "No guides" if guide is None else f"{category}/{guide}"
         guide_manager.active_guide = active_guide
 
     def get_state_from_process(self) -> dict:


edan-bainglass:
@yakutovicha tests failing because dependencies have moved to 3.12 while we're still stuck in 3.9 🥲

yakutovicha:
> @yakutovicha tests failing because dependencies have moved to 3.12 while we're still stuck in 3.9 🥲

Don't you want to change the base image to "edge", so things are tested against Python 3.12 already?

yakutovicha (approved):
👍

edan-bainglass:
> > @yakutovicha tests failing because dependencies have moved to 3.12 while we're still stuck in 3.9 🥲
> 
> Don't you want to change the base image to "edge", so things are tested against Python 3.12 already?
>  

Permanently, or to pass tests, then revert before merging? Or are we YOLOing `main`?

yakutovicha:
> Permanently, or to pass tests, the revert before merging? Or are we YOLOing main?

Another word I learned today.

I guess the answer is, as always - it depends. Do you plan to make another stable release before YOLOing? If not, then I would open a support branch and YOLO away.

edan-bainglass:
> Another word I learned today.

You're welcome 🙂

> I guess the answer is, as always - it depends. Do you plan to make another stable release before YOLOing? If not, then I would open a support branch and YOLO away.

Just to clarify, the YOLOing is on `main`, right? 😅 The support branch would be for backporting bug fixes, as discussed?

yakutovicha:
> Just to clarify, the YOLOing is on `main`, right? 😅 The support branch would be for backporting bug fixes, as discussed?

Yes, exactly that

PR #1552, opened 2026-08-07
Fix dependency incompatibilities

edan-bainglass:
This PR aims to stabilize the current, pre-upgrades state of the app:
- ~Install Wannier90 ecosystem from GitHub due to recent PyPI issues~ (see note below)
- Wannier90 v4 is now on Conda - update image accordingly
- Install all app plugins in the image (~10MB increase in image size)
- Pin phonopy due to breaking changes to the build system introduced in v2.26.0
- Remove inline comments from dependencies, which broke AiiDAlab compatibility checks that scan dependencies
- Pin aiidalab-qe-muon to a support branch that pins undi to a commit prior to dropping Python 3.9 support

I'd like to do one final release supporting Python 3.9, before we proceed with upgrading the app to the new infrastructure, to support existing deployments.

### Note regarding Wannier90 ecosystem

~I've pinned the problematic dependencies in the aiidalab-qe-wannier90 plugin directly, in version 0.1.10. It will take a day or so for @superstar54 to add me as maintainer on the PyPI project, as it appears that the API token needs refreshing. Once released, the image test will again pass here.~
Can't release to PyPI with Git-based dependencies 🥲 Restoring Git-references here

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ARG COMPUTER_LABEL="localhost"
 
 ARG HQ_URL_AMD64="https://github.com/It4innovations/hyperqueue/releases/download/v${HQ_VER}/hq-v${HQ_VER}-linux-x64.tar.gz"
 ARG HQ_URL_ARM64="https://github.com/It4innovations/hyperqueue/releases/download/v${HQ_VER}/hq-v${HQ_VER}-linux-arm64-linux.tar.gz"
-ARG MUON_PKG="aiidalab-qe-muon@git+https://github.com/aiidalab/aiidalab-qe-muon@v1.1.3"
+ARG MUON_PKG="aiidalab-qe-muon@git+https://github.com/aiidalab/aiidalab-qe-muon@support/1.1.x"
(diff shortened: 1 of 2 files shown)


edan-bainglass:
@danielhollas I was a bit disappointed with the Conda Wannier90 installation not reducing the image size (tested prior to adding all plugins). Can you check me on this? 🙏 Maybe I'm missing something.

danielhollas:
I don't see any problem in this PR. Not sure why you are expecting a decrease? It could have been the case, but we were careful to cleanup after the build so it's not that surprising that we don't see a difference.

edan-bainglass:
> I don't see any problem in this PR. Not sure why you are expecting a decrease? It could have been the case, but we were careful to cleanup after the build so it's not that surprising that we don't see a difference.

Paraphrasing you - "we try our best to clean up, but we aren't perfect" 😅 Anyhow, if no issues, can you kindly approved? 🙏Test pass now - see comment in PR description 🙂

yakutovicha (approved):
Looks good 👍

danielhollas (approved):

PR #1554, opened 2026-08-10
Bump version v26.06.5 -> v26.06.6

edan-bainglass:


yakutovicha (approved):
It would be very tempting to have only one source of truth for the version number.

edan-bainglass:
> It would be very tempting to have only one source of truth for the version number.

What do you mean?

yakutovicha:
> > It would be very tempting to have only one source of truth for the version number.
> 
> What do you mean?

There are 4 different places where the version number is specified.

PR #1562, opened 2026-08-13
Wizard refactor

edan-bainglass:
https://github.com/aiidalab/aiidalab-qe/pull/1427 detached from the AWB wizard and developed a generalized MVC-based  wizard, but only its steps. Here I add also classes for the entire Wizard (and model), with a dedicated "wizard" CSS class, which I then leverage to color-coordinate the wizard steps instead of relying on Jupyter CSS class names.

### DIFF
--- a/src/aiidalab_qe/app/wizard/model.py
+++ b/src/aiidalab_qe/app/wizard/model.py
@@ -7,16 +7,17 @@ from aiidalab_qe.app.result import ResultsStepModel
 from aiidalab_qe.app.structure import StructureStepModel
 from aiidalab_qe.app.submission import SubmissionStepModel
 from aiidalab_qe.common.mixins import HasModels
-from aiidalab_qe.common.mvc import Model
-from aiidalab_qe.common.wizard import WizardStepModel
+from aiidalab_qe.common.wizard import WizardModel, WizardStepModel
 
 
-class WizardModel(Model, HasModels[WizardStepModel]):
+class QeWizardModel(WizardModel, HasModels[WizardStepModel]):
     state = tl.Dict(None, allow_none=True)
     selected_index = tl.Int(None, allow_none=True)
     loading = tl.Bool(False)
 
     def load_from_state(self, state: dict):
+        super().load_from_state(state)
+
         step_index = state.get("step", 0) - 1
         structure_state = state.get("structure_state")
         configuration_state = state.get("configuration_state")
@@ -35,7 +36,7 @@ class WizardModel(Model, HasModels[WizardStepModel]):
                 structure_model.confirm()
                 configuration_model = t.cast(
                     ConfigurationStepModel,
-                    self.get_model("configure"),
+                    self.get_model("configuration"),
                 )
                 configuration_model.set_model_state(configuration_state)
 
@@ -43,7 +44,7 @@ class WizardModel(Model, HasModels[WizardStepModel]):
                     configuration_model.confirm()
                     submission_model = t.cast(
                         SubmissionStepModel,
-                        self.get_model("submit"),
+                        self.get_model("submission"),
                     )
                     if process_uuid is not None:
                         submission_model.refresh_codes(filter_codes_for_user=False)
@@ -67,7 +68,7 @@ class WizardModel(Model, HasModels[WizardStepModel]):
         )
         configuration_model = t.cast(
             ConfigurationStepModel,
-            self.get_model("configure"),
+            self.get_model("configuration"),
         )
         if structure_model.confirmed:
             configuration_model.structure_uuid = structure_model.structure_uuid
@@ -81,11 +82,11 @@ class WizardModel(Model, HasModels[WizardStepModel]):
         )
         configuration_model = t.cast(
             ConfigurationStepModel,
-            self.get_model("configure"),
+            self.get_model("configuration"),
         )
         submission_model = t.cast(
             SubmissionStepModel,
-            self.get_model("submit"),
+            self.get_model("submission"),
         )
         if configuration_model.confirmed:
             submission_model.structure_uuid = structure_model.structure_uuid
@@ -97,7 +98,7 @@ class WizardModel(Model, HasModels[WizardStepModel]):
     def update_results_model(self):
         submission_model = t.cast(
             SubmissionStepModel,
-            self.get_model("submit"),
+            self.get_model("submission"),
         )
         results_model = t.cast(
             ResultsStepModel,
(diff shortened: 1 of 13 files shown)


src/aiidalab_qe/app/utils/examples_importer.py
```diff
@@ -164,8 +164,14 @@ def _get_options(self) -> list[tuple[str, str]]:
 
 
 class ExamplesImporter(ipw.Tab):
-    def __init__(self, repo: str, tag: str):
-        super().__init__()
+    def __init__(self, repo: str, tag: str, **kwargs):
+        super().__init__(
+            layout=ipw.Layout(
+                **kwargs.pop("layout", {}),
```
danielhollas: What happens if the Layout here already contains `min_width`? I would expect this to fail since you would provide the same argument twice?

danielhollas (commented):
I will not have time to dive deeply into this as I am not that familiar with the MVC model here, but looks good on a quick skim.

src/aiidalab_qe/app/utils/examples_importer.py
edan-bainglass: True, I was intending for the latter to take over, but I'm doing that wrong (this is not a dict). But in any case, I'm reverting this part, as the `Tab` widget is a wrapper of tabs, not the individual ones, for which I want to apply the min-width field. Need to rethink.

danielhollas:
Sorry, this is beyond my scope at the moment, feel free to go ahead.

src/aiidalab_qe/app/wizard/model.py
```diff
@@ -7,16 +7,17 @@
 from aiidalab_qe.app.structure import StructureStepModel
 from aiidalab_qe.app.submission import SubmissionStepModel
 from aiidalab_qe.common.mixins import HasModels
-from aiidalab_qe.common.mvc import Model
-from aiidalab_qe.common.wizard import WizardStepModel
+from aiidalab_qe.common.wizard import WizardModel, WizardStepModel
 
 
-class WizardModel(Model, HasModels[WizardStepModel]):
+class QeWizardModel(WizardModel, HasModels[WizardStepModel]):
```
yakutovicha: WizardModel already inherits from `HasModels[WizardStepModel]`
```suggestion
class QeWizardModel(WizardModel):
```

edan-bainglass:
> Sorry, this is beyond my scope at the moment, feel free to go ahead.

Accidently clicked the re-review for you 😅

src/aiidalab_qe/common/wizard.py
```diff
@@ -280,3 +281,110 @@ class ConfirmableDependentWizardStep(
...
+
+        if (
+            (selected_step := model_list[index]).auto_advance
+            and not (index + 1 == len(model_list))
+            and selected_step.is_successful
+        ):
+            self.selected_index += 1
+
+
+class Wizard(ipw.Accordion):
```
yakutovicha, line 309: It would be great if you could add a bunch of unit tests for the class methods.

src/aiidalab_qe/common/wizard.py
```diff
@@ -280,3 +281,110 @@ class ConfirmableDependentWizardStep(
     ConfirmableWizardStep[CDWSM],
 ):
     """A confirmable dependent wizard step."""
+
+
+class WizardModel(Model, HasModels[WizardStepModel]):
+    state = tl.UseEnum(State, default_value=State.INIT)
```
yakutovicha: Defined as an enum, but used as a dict everywhere else:


```suggestion
    state = tl.Dict(None, allow_none=True)
```

src/aiidalab_qe/common/wizard.py
```diff
@@ -280,3 +281,110 @@ class ConfirmableDependentWizardStep(
...
+
+    def _on_step_state_change(self, _=None):
+        self._update_titles()
+
+    def _on_step_change(self, change: dict):
+        if (step_index := change["new"]) is not None:
+            self._render_step(step_index)
+
+    def _update_titles(self):
+        for i, (identifier, title) in enumerate(self._titles.items()):
```
yakutovicha, line 387: The code here believes that the list of titles and the list of models (created using `add_step`) are in sync, however, this is not enforced. Maybe something like this would be better:

```python
def _update_titles(self):
    for i, model in enumerate(self._models):
        title = self._titles[model.identifier]
        icon = self._icons[model.state]
        self.set_title(i, f"{icon} Step {i + 1}: {title}")
```

src/aiidalab_qe/common/wizard.py
```diff
@@ -280,3 +281,110 @@ class ConfirmableDependentWizardStep(
...
+            step_model = self._model.get_model(identifier)
+            icon = self._icons.get(step_model.state)
```
yakutovicha: This returns `None` if the state is not in the dict. Should we keep it just as direct dict access to avoid weird-looking titles such as: `None Step 1: Select structure`?

src/aiidalab_qe/app/wizard/model.py
```diff
@@ -7,16 +7,17 @@
...
     state = tl.Dict(None, allow_none=True)
     selected_index = tl.Int(None, allow_none=True)
     loading = tl.Bool(False)
```
yakutovicha: Those class vars are already present in the `WizardModel` class
```suggestion
```

yakutovicha (changes requested):
Thanks, @edan-bainglass, I have a few requests.

src/aiidalab_qe/common/wizard.py
yakutovicha: Or at least make get return an empty string.

src/aiidalab_qe/app/wizard/model.py
edan-bainglass: Good catch 👍

src/aiidalab_qe/common/wizard.py
edan-bainglass: Used as an Enum 🙂

```python
    def update_state(self):
        if self.confirmed:
            self.state = State.SUCCESS
        elif self.is_blocked:
            self.state = State.BLOCKED
        elif self.is_previous_step_successful and self.has_all_dependencies:
            self.state = State.CONFIGURED
        else:
            self.state = State.INIT
```

src/aiidalab_qe/common/wizard.py
edan-bainglass: Better to inject titles in `add_step`, thus forcing order. Also a nicer API I'd say.

src/aiidalab_qe/common/wizard.py
edan-bainglass: Sorry, that's the state of a WizardStep, not of the Wizard itself. That is indeed a dict 👍

src/aiidalab_qe/common/wizard.py
edan-bainglass: Note that we never added tests for the common widgets module. We start with the wizard 💪 Added tests for the whole wizard.py module.

yakutovicha (approved):
bellissimo!

PR #1565, opened 2026-08-14
Set `structure_name` in structure step from process

edan-bainglass:
This PR sets the `structure_name` in `StructureStepModel.set_model_state` when loading a process (completed or duplicated workflow). Without it set, the link between the structure manager and the step that keeps these in sync may fire on first render (if the step is not locked), which will reset the step, effectively defeating the duplication feature. Setting the trait resolves this issue, and more generally, was simply missed - it SHOULD be set!


yakutovicha (approved):
Looks good

PR #1573, opened 2026-08-18
Keep only some plugins pre-installed

edan-bainglass:
In principle, we should consider only Muon and Vibroscopy plugins, as these are the "big" client communities. We presently keep also Wannier90 due to recent PyPI issues.

Ideally, we should move a good chunk of executable installation and code setup to the plugin's respective post-install scripts.


yakutovicha (approved):
Looks good 👍

PR #1577, opened 2026-08-31
Sync with awb's `ipw.Output` -> `ipw.VBox` migration

yakutovicha:
`ProcessMonitor.log_widget` and `install_create_github_issue_exception_handler` now expect an `ipw.VBox` instead of an `ipw.Output`.


src/aiidalab_qe/app/main.py
```diff
@@ -65,7 +65,7 @@ def __init__(
                 icon="trash",
                 layout=ipw.Layout(width="fit-content"),
             )
-            reset_button.on_click(lambda _: self.log_widget.clear_output())
+            reset_button.on_click(lambda _: setattr(self.log_widget, "children", ()))
```
danielhollas, line 68: Why is `setattr` needed here?

danielhollas (commented):
LG, but I'll let Edan to take a closer look. Thanks

src/aiidalab_qe/app/main.py
edan-bainglass: No assignment in `lambda`, right? Not sure if `children` has a `.clear()` method.

edan-bainglass (approved):
Thanks @yakutovicha. LGTM!

